PR #1, opened 2015-03-19
Add PosgreSQL 9.2 image

mnagy:
It is based on the MySQL image and works similarly. You must set
`POSTGRESQL_{USER,PASSWORD,DATABASE}` environment variables and can
optionally set the 'postgres' db user password using the
`POSTGRESQL_ADMIN_PASSWORD` variable.

There are few minor outstanding issues that are marked and commented
with a `TODO`. Ideas and suggestions would be appreciated.

You can test the image with `make test && make test TARGET=rhel7`.

Documentation in README is still missing, but wanted early feedback.

### DIFF
--- a/9.2/Dockerfile
+++ b/9.2/Dockerfile
@@ -1,0 +1,40 @@
+FROM centos:centos7
+MAINTAINER  Martin Nagy <mnagy@redhat.com>
+
+# PostgreSQL image for OpenShift.
+# Volumes:
+#  * /var/lib/psql/data   - Database cluster for PostgreSQL
+# Environment:
+#  * $POSTGRESQL_USER     - Database user name
+#  * $POSTGRESQL_PASSWORD - User's password
+#  * $POSTGRESQL_DATABASE - Name of the database to create
+#  * $POSTGRESQL_ADMIN_PASSWORD (Optional) - Password for the 'postgres'
+#                           PostgreSQL administrative account
+
+# Image metadata
+ENV POSTGRESQL_VERSION      9.2
+ENV IMAGE_DESCRIPTION       PostgreSQL 9.2
+ENV IMAGE_TAGS              postgresql,postgresql92
+ENV IMAGE_EXPOSE_SERVICES   5432:postgresql
+
+EXPOSE 5432
+
+# This image must forever use UID 26 for postgres user so our volumes are
+# safe in the future. This should *never* change, the last test is there
+# to make sure of that.
+RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/postgresql92/epel-7-x86_64/download/rhscl-postgresql92-epel-7-x86_64.noarch.rpm && \
+    yum -y --setopt=tsflags=nodocs install postgresql92 && \
+    yum clean all && \
+    mkdir -p /var/lib/pgsql/data && chown postgres.postgres /var/lib/pgsql/data && \
+    test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)"
+
+COPY run-postgresql.sh /usr/local/bin/
+COPY contrib/.bashrc /var/lib/pgsql/
+
+VOLUME ["/var/lib/pgsql/data"]
+
+USER postgres
+
+ENTRYPOINT ["run-postgresql.sh"]
+CMD ["postgres"]
--- a/9.2/run-postgresql.sh
+++ b/9.2/run-postgresql.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+# For SCL enablement
+source $HOME/.bashrc
+
+set -eu
+
+# Data dir
+export PGDATA=/var/lib/pgsql/data
+
+# Be paranoid and stricter than we should be.
+psql_identifier_regex='^[a-zA-Z_][a-zA-Z0-9_]*$'
+psql_password_regex='^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$'
+
+function usage() {
+	if [ $# == 2 ]; then
+		echo "error: $1"
+	fi
+	echo "You must specify following environment variables:"
+	echo "  \$POSTGRESQL_USER (regex: '$psql_identifier_regex')"
+	echo "  \$POSTGRESQL_PASSWORD (regex: '$psql_password_regex')"
+	echo "  \$POSTGRESQL_DATABASE (regex: '$psql_identifier_regex')"
+	echo "Optional:"
+	echo "  \$POSTGRESQL_ADMIN_PASSWORD (regex: '$psql_password_regex')"
+	exit 1
+}
+
+function check_env_vars() {
+	[[ "$POSTGRESQL_USER"     =~ $psql_identifier_regex ]] || usage
+	[ ${#POSTGRESQL_USER} -le 63 ] || usage "PostgreSQL username too long (maximum 63 characters)"
+	[[ "$POSTGRESQL_PASSWORD" =~ $psql_password_regex   ]] || usage
+	[[ "$POSTGRESQL_DATABASE" =~ $psql_identifier_regex ]] || usage
+	[ ${#POSTGRESQL_DATABASE} -le 63 ] || usage "Database name too long (maximum 63 characters)"
+	if [ -v POSTGRESQL_ADMIN_PASSWORD ]; then
+		[[ "$POSTGRESQL_ADMIN_PASSWORD" =~ $psql_password_regex ]] || usage
+	fi
+}
+
+# Make sure env variables don't propagate to PostgreSQL process.
+function unset_env_vars() {
+	unset POSTGRESQL_USER
+	unset POSTGRESQL_PASSWORD
+	unset POSTGRESQL_DATABASE
+	unset POSTGRESQL_ADMIN_PASSWORD
+}
+
+if [ "$1" = "postgres" -a ! -f "$PGDATA/postgresql.conf" ]; then
+
+	check_env_vars
+
+	# Initialize the database cluster with utf8 support enabled by default.
+	# This might affect performance, see:
+	# http://www.postgresql.org/docs/9.2/static/locale.html
+	LANG=${LANG:-en_US.utf8} initdb
+
+	# PostgreSQL configuration.
+	cat >> "$PGDATA/postgresql.conf" <<-EOF
+
+		#
+		# Custom OpenShift configuration starting at this point.
+		#
+
+		# Listen on all interfaces.
+		listen_addresses = '*'
+	EOF
+
+	# Access control configuration.
+	cat >> "$PGDATA/pg_hba.conf" <<-EOF
+
+		#
+		# Custom OpenShift configuration starting at this point.
+		#
+
+		# Allow connections from all hosts.
+		host all all all md5
+	EOF
+
+	pg_ctl -w start
+	createuser "$POSTGRESQL_USER"
+	createdb --owner="$POSTGRESQL_USER" "$POSTGRESQL_DATABASE"
+	psql --command "ALTER USER \"${POSTGRESQL_USER}\" WITH ENCRYPTED PASSWORD '${POSTGRESQL_PASSWORD}';"
+
+	if [ -v POSTGRESQL_ADMIN_PASSWORD ]; then
+		psql --command "ALTER USER \"postgres\" WITH ENCRYPTED PASSWORD '${POSTGRESQL_ADMIN_PASSWORD}';"
+	fi
+
+	pg_ctl stop
+fi
+
+unset_env_vars
+exec "$@"
--- a/9.2/test/run
+++ b/9.2/test/run
@@ -1,0 +1,169 @@
+#!/bin/bash
+#
+# Test the PostgreSQL image.
+#
+# IMAGE_NAME specifies the name of the candidate image used for testing.
+# The image has to be available before this script is executed.
+#
+
+set -exo nounset
+shopt -s nullglob
+
+IMAGE_NAME=${IMAGE_NAME-openshift/postgresql-92-centos7-candidate}
+
+CIDFILE_DIR=$(mktemp --suffix=postgresql_test_cidfiles -d)
+
+function cleanup() {
+	for cidfile in $CIDFILE_DIR/* ; do
+		CONTAINER=$(cat $cidfile)
+
+		echo "Stopping and removing container $CONTAINER..."
+		docker stop $CONTAINER
+		exit_status=$(docker inspect -f '{{.State.ExitCode}}' $CONTAINER)
+		if [ "$exit_status" != "0" ]; then
+			echo "Dumping logs for $CONTAINER"
+			docker logs $CONTAINER
+		fi
+		docker rm $CONTAINER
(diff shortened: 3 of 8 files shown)


9.2/run-postgresql.sh
```diff
@@ -0,0 +1,83 @@
...
+# The '|| exit 1' is here so -e doesn't propagate into scl_source.
+source scl_source enable postgresql92 || exit 1
+
+function postgres_run {
+	runuser -u postgres -- "$@"
+}
+
+if [ ! -f "$PGDATA/postgresql.conf" ]; then
+
+	chown postgres.postgres "$PGDATA"
```
mnagy: A note about USER. We run this as root, but only thing that really requires root privileges is this chown. I would love to run the whole entrypoint as user postgres. However, we would have to make sure the volume is owned by the postgres user (uid 26).

Any ideas if this is possible?

9.2/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+ENV POSTGRESQL_VERSION      9.2
+ENV IMAGE_DESCRIPTION       PostgreSQL 9.2
+ENV IMAGE_TAGS              postgresql,postgresql92
+ENV IMAGE_EXPOSE_SERVICES   5432:postgresql
+
+# These are relevant to PostgreSQL
+ENV PGDATA /var/lib/pgsql/data
+# TODO: We should consider if this should be changeable. Can affect performance.
+#       http://www.postgresql.org/docs/9.2/static/locale.html
+ENV LANG en_US.utf8
```
mfojtik: Since we are not doing this for any other image, I would vote for nuking this line. You can set it in run-postgresql.sh (and users can change it by using -e)

9.2/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+MAINTAINER  Martin Nagy <mnagy@redhat.com>
```
mfojtik: Move this after FROM

9.2/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# to make sure of that.
+RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/postgresql92/epel-7-x86_64/download/rhscl-postgresql92-epel-7-x86_64.noarch.rpm && \
+    yum -y --setopt=tsflags=nodocs install postgresql92 && \
+    yum clean all && \
+    test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)"
+
+COPY run-postgresql.sh /usr/local/bin/
+
+VOLUME ["/var/lib/pgsql/data"]
```
mfojtik: why not have /var/lib/pgsql as VOLUME? Then you can persistently store your config files

9.2/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+# TODO: Can we run as non-root?
```
mfojtik: that is requirement :-)

9.2/Dockerfile
soltysh: +1

9.2/test/run
```diff
@@ -0,0 +1,105 @@
...
+# IMAGE_NAME specifies the name of the candidate image used for testing.
+# The image has to be available before this script is executed.
+#
+
+set -exo nounset
+shopt -s nullglob
+
+IMAGE_NAME=${IMAGE_NAME-openshift/postgresql-92-centos7-candidate}
+
+CIDFILE_DIR=$(mktemp --suffix=mysql_test_cidfiles -d)
```
soltysh: mysql :wink: ?

9.2/run-postgresql.sh
mfojtik: Why not chown in Dockerfile then?

9.2/run-postgresql.sh
```diff
@@ -0,0 +1,83 @@
...
+function valid_psql_password {
+	local var="$1" ; shift
+	[[ "${var}" =~ $psql_password_regex ]]
+}
+
+valid_psql_identifier "$POSTGRESQL_USER"     || usage
+valid_psql_password   "$POSTGRESQL_PASSWORD" || usage
+valid_psql_identifier "$POSTGRESQL_DATABASE" || usage
+
+# Make sure env variables don't propagate to mysqld process.
```
mfojtik: copy&paste, right?

9.2/run-postgresql.sh
soltysh: Nope if you have mounted volume `$PGDATA`.

9.2/run-postgresql.sh
```diff
@@ -0,0 +1,83 @@
...
+
+# Be paranoid and stricter than we should be.
+psql_identifier_regex='^[a-zA-Z_][a-zA-Z0-9_]*$'
+psql_password_regex='^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$'
+
+function usage {
+	echo "You must specify following environment variables:"
+	echo "  \$POSTGRESQL_USER (regex: '$psql_identifier_regex')"
+	echo "  \$POSTGRESQL_PASSWORD (regex: '$psql_password_regex')"
+	echo "  \$POSTGRESQL_DATABASE (regex: '$psql_identifier_regex')"
```
jhadvig: In MongoDB I am creating the `production` database if none is defined. I think some similar approach we had in v2. Wouldn't it be better to have it in all DB images (also MySQL) ?

9.2/test/run
```diff
@@ -0,0 +1,105 @@
...
+			echo "Dumping logs for $CONTAINER"
+			docker logs $CONTAINER
+		fi
+		docker rm $CONTAINER
+		rm $cidfile
+		echo "Done."
+	done
+	rmdir $CIDFILE_DIR
+}
+trap cleanup EXIT
```
jhadvig: + `SIGINT`

9.2/test/run
```diff
@@ -0,0 +1,105 @@
...
+	create_container $name $envs
+	CONTAINER_IP=$(get_container_ip $name)
+	test_connection $name
+	test_postgresql $name
+}
+
+# Tests.
+
+USER=user PASS=pass run_tests no_admin
+USER=user1 PASS=pass1 ADMIN_PASS=r00t run_tests admin
```
jhadvig: You should be also testing the admin user, not only standalone user.

9.2/run-postgresql.sh
soltysh: I'd rather see a failure when that name isn't specified. It's required the
same way username and password is. This is the behavior I'd vote for all
images.

9.2/Dockerfile
mnagy: /var/lib/pgsql/data is a commonly used path for a database cluster, which also contains within the configuration that is necessary, so config files are persistently stored. It didn't make much sense to me to use /var/lib/pgsql. Besides, I think that while experimenting I did see some psql history files there, or something like that. Probably not relevant to us, but stil..

9.2/test/run
mnagy: Haha, good catch, I thought I killed all mysql references :)

9.2/run-postgresql.sh
mnagy: Yep, will fix.

9.2/run-postgresql.sh
mnagy: I think we discussed somewhere for MySQL image that we don't want optional vars and that we recommend that in the image writing guide, that's why it's not optional.

Personally, I'd be in favour of that, it would make using the image outside Openshift a LOT easier.

9.2/test/run
mnagy: I'll add in access tests that we do in mysql.

9.2/test/run
mnagy: Good catch.

mnagy:
Addressed all issues plus something extra :) PTAL

9.2/run-postgresql.sh
jhadvig: I dont recall any discussion about optional vars. Also the MYSQL_ROOT_PASSWORD var is optional, is it not ?
I would go with similar behavior as the MongoDB, less typing => happy users :)

mnagy: Hm. We can talk about that on scrum I suppose?

mfojtik: database name is required, always... we need to fix mongo

9.2/test/run
mfojtik: also test failures ... no user set, etc...

9.2/run-postgresql.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash -e
+
+# Data dir
+export PGDATA=/var/lib/pgsql/data
+
+# Enable utf8 support. This might affect performance:
+# http://www.postgresql.org/docs/9.2/static/locale.html
+export LANG=en_US.utf8
```
mfojtik: this needs to be tuneable via env var... why is default not enough? can't you set this in postgres config file instead?

9.2/run-postgresql.sh
mnagy: @jhadvig also, MYSQL_ROOT_PASSWORD is a special case since it doesn't have a default but not specifying it alters the behavior (remote login to root account is disabled).

9.2/test/run
mnagy: You mean no env vars set when creating the container? Please elaborate.

mfojtik:
LGTM. Nice job @mnagy, kudos!

9.2/test/run
mfojtik: invalid values, empty values.... exercise the validate functions in image

9.2/run-postgresql.sh
mfojtik: if you docker exec to running container and run psql, does it connect to postres? it should without having to specify any arguments... please test that

9.2/run-postgresql.sh
mnagy: Right, will fix that tomorrow so LANG can be overridden. I might, but setting the LANG var is the simplest way to do it, I didn't see any reason to modify config file because of it. Btw, it mostly affects initdb. And yeah, the default is probably faster a bit, but without utf support, at least that's how I understand it.

9.2/run-postgresql.sh
mnagy: You'll get command not found error because of SCL :) But yeah, if you run it through scl, psql will work as is.

9.2/test/run
mnagy: I take that back; it now calls cleanup two times. EXIT trap is invoked after an interrupt as well.

mnagy:
Some enhancements:
No further need to use scl when doing docker exec, added contrib directory. Added tests for env variable validations, simplified that code a bit as well. PTAL.

mnagy:
Very small change added. We're now also validating maximum user/db name length. Also added tests for that.

mnagy:
@jhadvig @soltysh can I get more reviews in addition to @mfojtik's, please?

9.2/run-postgresql.sh
```diff
@@ -0,0 +1,91 @@
...
+set -eu
+
+# Data dir
+export PGDATA=/var/lib/pgsql/data
+
+# Be paranoid and stricter than we should be.
+psql_identifier_regex='^[a-zA-Z_][a-zA-Z0-9_]*$'
+psql_password_regex='^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$'
+
+function usage {
```
jhadvig: could you maybe do the function declaration in this file more consistent with the test/run script?

mnagy: Done. PTAL

9.2/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+# to make sure of that.
+RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/postgresql92/epel-7-x86_64/download/rhscl-postgresql92-epel-7-x86_64.noarch.rpm && \
+    yum -y --setopt=tsflags=nodocs install postgresql92 && \
+    yum clean all && \
+    mkdir -p /var/lib/pgsql/data && chown postgres.postgres /var/lib/pgsql/data && \
+    test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)"
+
+COPY run-postgresql.sh /usr/local/bin/
+COPY contrib/.bashrc /var/lib/pgsql/
```
soltysh: Can you do that also for https://github.com/openshift/mongodb/ and https://github.com/openshift/mysql/, please :smile_cat:

soltysh: Even more, can you add that to all our images :sweat_smile:

mnagy: @soltysh For mysql, see https://github.com/openshift/mysql/pull/24
For MongoDB, @jhadvig will do that. :)

soltysh: many kudos @mnagy

mnagy: @soltysh thanks :) I'd also appreciate a certain four-letter acronym :P

9.2/test/run
```diff
@@ -0,0 +1,169 @@
...
+		postgresql_cmd <<< "SELECT 1;" &&
+			echo "    $USER($PASS) access granted as expected"
+	else
+		postgresql_cmd <<< "SELECT 1;" ||
+			echo "    $USER($PASS) access denied as expected"
+	fi
+}
+
+# Make sure the invocation of docker run fails.
+function assert_image_creation_fails {
```
jhadvig: It should be `assert_container_creation_fails`

mnagy:
Added SCL info into README.md, few minor fixes. PTAL

soltysh:
Here's mine LGTM :)

jhadvig:
@mnagy LGTM, good job :)
Merging

9.2/test/run
```diff
@@ -0,0 +1,169 @@
...
+	echo $(cat "$CIDFILE_DIR/$id")
+}
+
+function get_container_ip() {
+	local id="$1" ; shift
+	docker inspect --format='{{.NetworkSettings.IPAddress}}' $(get_cid "$id")
+}
+
+function postgresql_cmd() {
+	docker run --rm -e PGPASSWORD="${PASS}" $IMAGE_NAME psql postgresql://$USER@$CONTAINER_IP:5432/db "$@"
```
rhcarvalho: This uses `PASS` and `USER` as global variables, so we're never actually testing the access of the admin user...

soltysh: Can you open a bug report against this repo, so it doesn't get forgotten?

rhcarvalho: I was wrong. In fact, what happens is that `assert_login_access` shadows `USER` and `PASS` with local variables of the same name, and because in bash [local variables are visible in the scope of a function *and its children*](http://tldp.org/LDP/abs/html/localvar.html#FTN.AEN18568), `postgresql_cmd` eventually uses those local variables when called from within `assert_login_access`.

Needless to say, that's very very error prone and hard to reason about...

PR #8, opened 2015-04-03
Updated and unified README.md with other images

soltysh:


mnagy:
I dislike having README lines wrapped. It's harder to review diffs when something changes.

soltysh:
@mnagy this is what we have everywhere else. Personally I don't like having long lines, it's harder to review changes if just one word changes, with shorter lines it's more readable

mnagy:
@soltysh I disagree. It does not make a difference, you can see one word difference on github and git has that feature as well, see --word-diff. On the other hand, if you change more lines, you need to reformat and then it's very hard to see what changed.

soltysh:
@mnagy still not convinced.

README.md
```diff
@@ -55,24 +79,65 @@ You can also set following mount points by passing `-v /host:/container` flag to
...
 
-We will assume that you are using the `openshift/postgresql-92-centos7` image. Supposing that you want to set only mandatory environment variables and not store the database in a host directory, you need to execute the following command:
+Usage
+----------------------
 
-```console
-docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 openshift/postgresql-92-centos7
+We will assume that you are using the `openshift/postgresql-92-centos7` image.
+Suppose that you want to set only mandatory environment variables and not store
+the database in a host directory, you need to execute the following command:
```
mnagy: I think this should stay "Supposing", @bparees can you please review this sentence?

And again, this change is very hard to notice when the paragraph is not on a single line.

mnagy:
Besides the suppose word, it looks ok.

README.md
bparees: I'm actually not sure! :frowning: 

From a little googling it seems they are actually pretty interchangeable:
http://forum.wordreference.com/showthread.php?t=520484

You could replace suppose/supposing with "If" and simplify it :)  ("If you want to......")

soltysh: I found that too, I've chosen suppose because we have that form in other images, so it was easier to fix in one than in a couple of them :wink:

README.md
```diff
@@ -1,39 +1,63 @@
...
+
+
+Installation
+----------------------
 Choose between CentOS7 or RHEL7 based image:
 
 *  **RHEL7 based image**
 
-To build a RHEL7-based image, you need to run Docker build on a properly subscribed RHEL machine.
+    To build a RHEL7-based image, you need to run Docker build on properly
```
bparees: "on a properly"

bparees:
one minor nit, otherwise lgtm.

mfojtik:
LGTM

mnagy:
LGTM

PR #9, opened 2015-04-09
Getting rid of unpractical dollar signs

jhadvig:
@mfojtik I've got rid of the dollar signs as you asked int the mongodb replicaSet PR.
PTAL


mnagy:
LGTM

PR #10, opened 2015-04-22
Support random UIDs

csrwng:
This PR allows the postgresql image to start with a random uid as in:
```
docker run -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass \
   -e POSTGRESQL_DATABASE=db -p 5432:5432 \
   -u 12345 openshift/postgresql-92-centos7
```
It requires the use of a nss provider because postgresql tries to look up the current username with the uid. By passing in a dynamically generated passwd file to nss_wrapper.so, it's possible to make postgresql think that the current user's name is 'postgres'. Currently nss_wrapper.so is included in the PR but ideally it'd be downloaded from the SCL server. There's a fedora package for it here: https://apps.fedoraproject.org/packages/nss_wrapper

### DIFF
--- a/9.2/Dockerfile
+++ b/9.2/Dockerfile
@@ -24,13 +24,16 @@ EXPOSE 5432
 # to make sure of that.
 RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/postgresql92/epel-7-x86_64/download/rhscl-postgresql92-epel-7-x86_64.noarch.rpm && \
-    yum -y --setopt=tsflags=nodocs install gettext postgresql92 && \
+    yum -y --setopt=tsflags=nodocs --enablerepo=centosplus install gettext postgresql92 epel-release && \
+    yum -y --setopt=tsflags=nodocs install nss_wrapper && \
     yum clean all && \
     mkdir -p /var/lib/pgsql/data && chown postgres.postgres /var/lib/pgsql/data && \
     test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)"
 
 COPY run-postgresql.sh /usr/local/bin/
 COPY contrib /var/lib/pgsql/
+
+RUN chmod -R go+rwx /var/lib/pgsql && chmod go+w /var/run/postgresql
 
 VOLUME ["/var/lib/pgsql/data"]
 
--- a/9.2/Dockerfile.rhel7
+++ b/9.2/Dockerfile.rhel7
@@ -26,12 +26,17 @@ RUN yum install -y yum-utils gettext && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
     yum install -y --setopt=tsflags=nodocs postgresql92 && \
+    curl -O https://dl.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-5.noarch.rpm && \
+    rpm -Uvh epel-release-7-5.noarch.rpm && rm epel-release-7-5.noarch.rpm && \
+    yum install -y --enablerepo=epel --setopt=tsflags=nodocs nss_wrapper && \
     yum clean all && \
     mkdir -p /var/lib/pgsql/data && chown postgres.postgres /var/lib/pgsql/data && \
     test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)"
 
 COPY run-postgresql.sh /usr/local/bin/
 COPY contrib /var/lib/pgsql/
+
+RUN chmod -R go+rwx /var/lib/pgsql && chmod go+w /var/run/postgresql
 
 VOLUME ["/var/lib/pgsql/data"]
 
--- a/9.2/run-postgresql.sh
+++ b/9.2/run-postgresql.sh
@@ -1,12 +1,22 @@
 #!/bin/bash
 
+# Set HOME variable
+export HOME=/var/lib/pgsql
+
 # For SCL enablement
 source $HOME/.bashrc
 
 set -eu
 
 # Data dir
-export PGDATA=$HOME/data
+if [ -O $HOME/data ]; then
+  export PGDATA=$HOME/data
+else
+  # If current user does not own data directory
+  # create a subdirectory that the user does own
+  mkdir $HOME/data/userdata
+  export PGDATA=$HOME/data/userdata
+fi
 POSTGRESQL_CONFIG_FILE=$HOME/openshift-custom-postgresql.conf
 
 # Configuration settings.
@@ -68,7 +78,7 @@ function initialize_database() {
 	cat >> "$PGDATA/postgresql.conf" <<-EOF
 
 		# Custom OpenShift configuration:
-		include '../openshift-custom-postgresql.conf'
+		include '${POSTGRESQL_CONFIG_FILE}'
 	EOF
 
 	# Access control configuration.
@@ -97,6 +107,13 @@ function initialize_database() {
 # New config is generated every time a container is created. It only contains
 # additional custom settings and is included from $PGDATA/postgresql.conf.
 envsubst < ${POSTGRESQL_CONFIG_FILE}.template > ${POSTGRESQL_CONFIG_FILE}
+
+# Generate passwd file based on current uid
+export USER_ID=$(id -u)
+envsubst < ${HOME}/passwd.template > ${HOME}/passwd
+export LD_PRELOAD=libnss_wrapper.so
+export NSS_WRAPPER_PASSWD=/var/lib/pgsql/passwd
+export NSS_WRAPPER_GROUP=/etc/group
 
 if [ "$1" = "postgres" -a ! -f "$PGDATA/postgresql.conf" ]; then
 	initialize_database
--- a/9.2/test/run
+++ b/9.2/test/run
@@ -79,9 +79,10 @@ function test_postgresql() {
 
 function create_container() {
 	local name=$1 ; shift
+	local cargs=${CONTAINER_ARGS:-}
 	cidfile="$CIDFILE_DIR/$name"
 	# create container with a cidfile in a directory for cleanup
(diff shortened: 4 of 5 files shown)


9.2/run-postgresql.sh
```diff
@@ -1,12 +1,23 @@
 #!/bin/bash
 
+# Set HOME variable
+export HOME=/var/lib/pgsql
```
mnagy: Why isn't the HOME var set?

9.2/run-postgresql.sh
```diff
@@ -1,12 +1,23 @@
...
 # Data dir
-export PGDATA=$HOME/data
+if [ -O $HOME/data ]
+then
+  export PGDATA=$HOME/data
+else
+  # If current user does not own data directory
+  # create a subdirectory that the user does own
+  mkdir $HOME/data/userdata
+  export PGDATA=$HOME/data/userdata
```
mnagy: How will the dir be created if the user does not own $HOME/data? Also, I think that in this case, we should fail.

mnagy:
Do we really need to use the nss wrapper? Can't postgres run with user name being the same as the uid?

9.2/run-postgresql.sh
csrwng: In this case the user has write access to $HOME/data because it is now set as world-writeable by the Dockerfile. However, the user is not the owner. When postgres tries to start, it attempts to fix the permissions on its data directory and if the user is not the owner, it complains. The solution I found was to create a subdirectory under data that is owned by the current user.

csrwng:
@mnagy - with a random user name (like 'u1000'), I get a message like:
```
psql: FATAL:  database "u1000" does not exist
```
and the process dies. It expects the name of the current user to be 'postgres'

csrwng:
Switched to locally building nss_wrapper. Added tests

csrwng:
@mnagy - I switched to using EPEL for centos. However, I haven't switched it for RHEL since I wasn't sure if there's an "official" way of doing it.

csrwng:
@mnagy @bparees - Updated the rhel7 Dockerfile; should be ready for final review.

9.2/run-postgresql.sh
bparees: $HOME could be (normailly will be) a bindmount right?  in that case are we just assuming the permissions are correct?

9.2/run-postgresql.sh
```diff
@@ -98,6 +109,14 @@ function initialize_database() {
 # additional custom settings and is included from $PGDATA/postgresql.conf.
 envsubst < ${POSTGRESQL_CONFIG_FILE}.template > ${POSTGRESQL_CONFIG_FILE}
 
+# Generate passwd file based on current uid
+export USER_ID=$(id -u)
+envsubst < ${HOME}/passwd.template > ${HOME}/passwd
+cat ${HOME}/passwd
```
bparees: debugging remnants?

9.2/test/run
```diff
@@ -79,9 +79,10 @@ function test_postgresql() {
 
 function create_container() {
 	local name=$1 ; shift
+        local cargs=${CONTAINER_ARGS:-}
```
bparees: alignment

9.2/run-postgresql.sh
```diff
@@ -98,6 +109,14 @@ function initialize_database() {
...
+export LD_PRELOAD=libnss_wrapper.so
+export NSS_WRAPPER_PASSWD=/var/lib/pgsql/passwd
+export NSS_WRAPPER_GROUP=/etc/group
```
bparees: why did we end up needing all this? postgres insists on having a user named postgres?

9.2/Dockerfile.rhel7
```diff
@@ -26,13 +26,20 @@ RUN yum install -y yum-utils gettext && \
...
     yum clean all && \
+    chmod go+rx /var/lib/pgsql && \
+    chmod go+w /var/run/postgresql && \
     mkdir -p /var/lib/pgsql/data && chown postgres.postgres /var/lib/pgsql/data && \
     test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)"
 
 COPY run-postgresql.sh /usr/local/bin/
 COPY contrib /var/lib/pgsql/
 
+RUN chmod -R go+rwx /var/lib/pgsql
```
bparees: isn't this redundant to line 33 above?

9.2/run-postgresql.sh
csrwng: postgres insists on looking up the current user and getting an entry back

9.2/Dockerfile.rhel7
csrwng: It's after we copy stuff over. (from contrib)

9.2/run-postgresql.sh
csrwng: In the 9.2 source base, this is the code in initdb.c:
```c
static char *
get_id(void)
{
#ifndef WIN32

	struct passwd *pw;

	if (geteuid() == 0)			/* 0 is root's uid */
	{
		fprintf(stderr,
				_("%s: cannot be run as root\n"
				  "Please log in (using, e.g., \"su\") as the "
				  "(unprivileged) user that will\n"
				  "own the server process.\n"),
				progname);
		exit(1);
	}

	pw = getpwuid(geteuid());
	if (!pw)
	{
		fprintf(stderr,
			  _("%s: could not obtain information about current user: %s\n"),
				progname, strerror(errno));
		exit(1);
	}
```
In later versions, it's been replaced by another function (get_username_or_exit) but it's the same deal:
http://doxygen.postgresql.org/username_8c.html#a0e84f99bf0c1826e3f629131a7719aac

9.2/run-postgresql.sh
csrwng: Yes

9.2/Dockerfile.rhel7
csrwng: removed first one

csrwng:
Comments addressed

9.2/Dockerfile
```diff
@@ -24,14 +24,18 @@ EXPOSE 5432
 # to make sure of that.
 RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/postgresql92/epel-7-x86_64/download/rhscl-postgresql92-epel-7-x86_64.noarch.rpm && \
-    yum -y --setopt=tsflags=nodocs install gettext postgresql92 && \
+    yum -y --setopt=tsflags=nodocs --enablerepo=centosplus install gettext postgresql92 epel-release && \
+    yum -y --setopt=tsflags=nodocs install nss_wrapper && \
     yum clean all && \
+    chmod go+w /var/run/postgresql && \
```
mfojtik: why you are doing this here and at line 37? ;-)

9.2/run-postgresql.sh
```diff
@@ -1,12 +1,23 @@
...
+export HOME=/var/lib/pgsql
+
 # For SCL enablement
 source $HOME/.bashrc
 
 set -eu
 
 # Data dir
-export PGDATA=$HOME/data
+if [ -O $HOME/data ]
```
mfojtik: please use `if cond; then` consistently

bparees:
lgtm.
add to whitelist
ok to test

9.2/Dockerfile
csrwng: Not the same directory. The reason I need to do the other one at line 37 is that I need to set permissions on the stuff we copy over from the contrib directory.

9.2/run-postgresql.sh
csrwng: In the case where $HOME/data is on an NFS mount, ideally the uid that's running the container owns the directory. If at worst they only have write permission, then the same logic applies... a subdirectory will be created that's owned by the current uid.

9.2/Dockerfile
bparees: any reason not to do them both on line 37 so we group the chmod operations?

csrwng: No reason, but now I've run into a weird bug. Only happens locally on docker 1.6:
If I don't chmod right after the yum install for /var/lib/pgsql, then I can't create files in it. I've examined the permissions and they are right (notice permissions on pgsql):
```
bash-4.2$ ls -al /var/lib
total 44
drwxr-xr-x 15 root     root     4096 Apr 29 13:58 .
drwxr-xr-x 25 root     root     4096 Apr 29 13:58 ..
drwxr-xr-x  2 root     root     4096 Apr 15 14:29 alternatives
drwxr-xr-x  2 root     root     4096 Jun 10  2014 games
drwxr-xr-x  2 root     root     4096 Jun 10  2014 misc
drwxrwxrwx  4 postgres postgres 4096 Apr 29 13:58 pgsql
drwxr-xr-x  2 root     root     4096 Apr 29 13:55 rpm
drwxr-xr-x  2 root     root     4096 Jun 10  2014 rpm-state
drwxr-xr-x  2 root     root     4096 Mar  6 05:04 selinux
drwxr-xr-x  4 root     root     4096 Apr 15 14:29 systemd
drwxr-xr-x 10 root     root     4096 Apr 29 13:56 yum
bash-4.2$ touch /var/lib/pgsql/testfile
touch: cannot touch '/var/lib/pgsql/testfile': Permission denied
bash-4.2$
```
If I do the chmod right after yum install, AND after I copy over the contrib directory, then I don't see this issue.

jhadvig:
LGTM :+1:

csrwng:
[test]

9.2/Dockerfile
csrwng: Just to follow up... the images built on the AWS vm work just fine with the chmod after everything, even when I pull the built images and run them locally on my boot2docker 1.6 vm. The issue happens only when I build the image locally. Moved the chmod to the single statement after things are copied in.

csrwng:
@bparees is this ready to merge?

bparees:
@csrwng our pull bot is confused because you weren't on the list when it opened, i think i've manually kicked off a test here:
https://ci.openshift.redhat.com/jenkins/job/postgresql/11/

if it passes, i'll merge.

PR #13, opened 2015-05-03
add streaming replication and openshift replca example

jmccormick2001:
I've added streaming replication to the run-postgresql.sh script, also a replica example of using replication controllers in Openshift, I borrowed from the mysql example.

### DIFF
--- a/9.2/run-postgresql.sh
+++ b/9.2/run-postgresql.sh
@@ -13,6 +13,9 @@ else
   fi
   export PGDATA=$HOME/data/userdata
 fi
+
+export PGDATA=$HOME/data
+POSTGRESQL_RECOVERY_FILE=$HOME/openshift-custom-recovery.conf
 POSTGRESQL_CONFIG_FILE=$HOME/openshift-custom-postgresql.conf
 
 # Configuration settings.
@@ -33,6 +36,8 @@ function usage() {
   echo "  POSTGRESQL_DATABASE (regex: '$psql_identifier_regex')"
   echo "Optional:"
   echo "  POSTGRESQL_ADMIN_PASSWORD (regex: '$psql_password_regex')"
+  echo "  POSTGRESQL_REPLICA (true or false)"
+  echo "  POSTGRESQL_MASTER (host or ip address of master)"
   echo "Settings:"
   echo "  POSTGRESQL_MAX_CONNECTIONS (default: 100)"
   echo "  POSTGRESQL_SHARED_BUFFERS (default: 32MB)"
@@ -54,12 +59,21 @@ function check_env_vars() {
   fi
 }
 
+function check_env_vars_for_replica() {
+	if ! [[ -v POSTGRESQL_MASTER && -v POSTGRESQL_REPLICA ]]; then
+		usage
+	fi
+
+}
+
 # Make sure env variables don't propagate to PostgreSQL process.
 function unset_env_vars() {
   unset POSTGRESQL_USER
   unset POSTGRESQL_PASSWORD
   unset POSTGRESQL_DATABASE
   unset POSTGRESQL_ADMIN_PASSWORD
+  unset POSTGRESQL_REPLICA
+  unset POSTGRESQL_MASTER
 }
 
 function initialize_database() {
@@ -86,18 +100,55 @@ EOF
 
 # Allow connections from all hosts.
 host all all all md5
+# Allow replication connections from all hosts.
+host replication all all md5
 EOF
 
   pg_ctl -w start
   createuser "$POSTGRESQL_USER"
   createdb --owner="$POSTGRESQL_USER" "$POSTGRESQL_DATABASE"
   psql --command "ALTER USER \"${POSTGRESQL_USER}\" WITH ENCRYPTED PASSWORD '${POSTGRESQL_PASSWORD}';"
+  psql --command "ALTER USER \"${POSTGRESQL_USER}\" WITH REPLICATION ;"
 
   if [ -v POSTGRESQL_ADMIN_PASSWORD ]; then
     psql --command "ALTER USER \"postgres\" WITH ENCRYPTED PASSWORD '${POSTGRESQL_ADMIN_PASSWORD}';"
   fi
 
   pg_ctl stop
+}
+function initialize_replica() {
+	echo "initialize replica called"
+	check_env_vars
+	check_env_vars_for_replica
+	cd /tmp
+	pwd
+	ls -al
+	ps ax
+	cat >> ".pgpass" <<-EOF
+	*:*:*:*:${POSTGRESQL_PASSWORD}
+	EOF
+	chmod 0600 .pgpass
+	export PGPASSFILE=/tmp/.pgpass
+	rm -rf $PGDATA/*
+	chmod 0700 $PGDATA
+	pg_basebackup -x --no-password --pgdata $PGDATA --host=$POSTGRESQL_MASTER --port=5432 -U $POSTGRESQL_USER
+	# PostgreSQL recovery configuration.
+	cat >> "$PGDATA/recovery.conf" <<-EOF
+
+		# Custom OpenShift recovery configuration:
+		include '../openshift-custom-recovery.conf'
+	EOF
+
+	pg_ctl -w start
+while :
+do
+	echo "Press [CTRL+C] to stop.."
+	sleep 1000
+done
+
+	cat $PGDATA/pg_log/*
+
+	pg_ctl stop
 }
 
 # New config is generated every time a container is created. It only contains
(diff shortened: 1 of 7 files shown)


smarterclayton:
@mfojtik

smarterclayton:
or @mnagy

README.md
```diff
@@ -141,3 +143,39 @@ User can choose between testing PostgreSQL based on RHEL or CentOS image.
...
+which would point to the master.
+
+The environment variable POSTGRESQL_MASTER is passed to the replica upon
+startup which will be used to perform the connection from the replica to
+the master.  This value can be an IP address or a host name if you are
+running within a Docker DNS orchestration environment (e.g. Openshift).
+
+The environment variable POSTGRESQL_REPLICA when set, is used to denote
+that you want to configure this Postgres instance as a replica and not a master.
+For example, a recovery.conf file is created and placed in PGDAT when this
```
mnagy: PGDATA

mnagy:
Please use `PostgreSQL` instead of `Postgres` in docs and comments, for consistency.

9.2/run-postgresql.sh
```diff
@@ -27,12 +28,34 @@ function usage() {
...
+        export RC=$?
+
+        echo "master connectivity test return code " $RC
+        if [ $RC = "0" ] ; then
+                echo "connection reached to master"
+                return 1
+        fi
+        echo "waiting for master pg to start..."
+        sleep 5
+done
```
mnagy: These should be tabs. I'll probably want to change indentation to two spaces anyway, so not a big deal.

9.2/run-postgresql.sh
```diff
@@ -93,14 +129,50 @@ function initialize_database() {
...
+	else
+		initialize_database
+	fi
 fi
 
 unset_env_vars
+
 exec "$@"
+
+
```
mnagy: No trailing empty lines please.

mnagy:
@jmccormick2001 Can you please add test cases? I imagine that running one master, two replicas, writing something and then verifying that the change propagated to master and both replicas would be an ideal test case.

jmccormick2001:
I'll incorporate these changes and do another PR this week.  thanks

rhcarvalho:
@jmccormick2001 I'm looking forward to see this PR merged :smile:. Instead of creating a new one, you can work on the same branch/PR. Just rebase it to openshift:master and then push force your changes to your jmccormick2001:master branch, Github will smartly update this PR. I'd be happy to help you with that if you need a hand.

mnagy:
@jmccormick2001 let me know if you have trouble with the tests. You can create a new function in 9.2/test/run (e.g. `run_replication_tests`) that would be called at the end of the script. You would probably want to use `create_container master [-e ARG1=foo -e ARG2=bar..]` and same for slave1 and slave2, with appropriate env vars set.

Alternatively, I can do the tests for you if you like.

Also, another nit: I noticed just now that you added files `master` and `replica`. I'm guessing these were just utility scripts for testing? I think we can remove them before merging this PR.

jmccormick2001:
Martin,  I just pushed a set of updates, hopefully I didn't miss anything, thanks for looking into writing the test, that would be very helpful if you could take that on.  regards.

mnagy:
Thanks, I'll take a look tomorrow.

9.2/run-postgresql.sh
```diff
@@ -3,6 +3,9 @@
 # Set HOME variable
 export HOME=/var/lib/pgsql
 
+# Set HOME variable
+export HOME=/var/lib/pgsql
+
```
mnagy: This seems to be duplicated by accident.

9.2/run-postgresql.sh
```diff
@@ -90,12 +107,16 @@ function initialize_database() {
...
+		# Allow replication connections from all hosts.
+		host replication all all md5
 	EOF
 
 	pg_ctl -w start
 	createuser "$POSTGRESQL_USER"
 	createdb --owner="$POSTGRESQL_USER" "$POSTGRESQL_DATABASE"
 	psql --command "ALTER USER \"${POSTGRESQL_USER}\" WITH ENCRYPTED PASSWORD '${POSTGRESQL_PASSWORD}';"
+	psql --command "ALTER USER \"${POSTGRESQL_USER}\" WITH REPLICATION ;"
```
mnagy: Isn't this dangerous? According to http://www.postgresql.org/docs/9.2/static/sql-createrole.html:

> REPLICATION
> NOREPLICATION
> 
>    These clauses determine whether a role is allowed to initiate streaming
> replication or put the system in and out of backup mode. A role having the
> REPLICATION attribute is a very highly privileged role, and should only be
> used on roles actually used for replication. If not specified,
> NOREPLICATION is the default.

I don't have a lot of experience with replication, but maybe there should be a separate user for this, or just use the admin one? @praiskup any thoughts on this?

praiskup: There are probably better ways to avoid shell/SQL injection.  Discussed
some time ago off-github.  If we are talking only about replication
itself, I'm afraid I can't say something is good or wrong, or dangerous -
especially in paas.  It usually depends on the context.

Somebody who will maintain such complicated scenarios like SR/failover
should design optimal configuration - which will be useful for every user
of the image.

praiskup: Maybe @jmlich, @hhorak fyi.

mnagy: I'm not that much concerned about shell/SQL injection. Those are really hard to do right, from what I've tried, at least for MySQL.

Anyway, my main concern here is that the REPLICATION role is enabled by default, even if replication is not used. @jmccormick2001 can we at least put an if there and only alter the role if replication is used?

9.2/run-postgresql.sh
```diff
@@ -17,6 +20,9 @@ else
   mkdir $HOME/data/userdata
   export PGDATA=$HOME/data/userdata
 fi
+
+export PGDATA=$HOME/data
```
mnagy: This line is causing tests to fail, please remove.

mnagy:
@jmccormick2001 sorry for getting back to you so late. I had some troubles with the tests and then got sidetracked and forgot about this PR. I've spent today some time reviewing and testing, and found out the (ordinary) tests were failing because of the mentioned issue with $PGDATA. Hopefully I'll manage to finish those replication tests.

9.2/run-postgresql.sh
hhorak: My paas thinking is still in early development, but in container we treat `POSTGRESQL_USER` as kind of admin, right? So having `REPLICATION` role set for this user doesn't seem too risky. It shouldn't be set if it is not necessary though, as proposed by @mnagy.

mnagy: That's the thing, POSTGRESQL_USER is supposed to be an ordinary user without privileges. But from what you're saying, I guess the best approach would be to use the `postgres` account and give it the `REPLICATION` role, unless it already has it. We'd then require the admin password for replication to work.

@jmccormick2001 I don't want to waste your time with this. I'll be happy to rework this logic later, but I would like there to be a test for replication enablement, so that we don't have any impact on non-replication use cases.

hhorak: @mnagy wrote:
> But from what you're saying, I guess the best approach would be to use the postgres account and give it the REPLICATION role, unless it already has it.

Actually this makes the problem even worse, i.e. solving problems with too strong role by using even more strong role... I don't think this is what we want. What would be imho better would be creating a user dedicated for replication only.

mnagy: OK, that sounds like an ideal solution (hope it won't take too long to implement). I'll open an issue for that after this PR is merged.

jhadvig:
@jmccormick2001 could you please squash all the commits into one? Would be more readable if we need to look only on one commit instead of several commits. Thanks :+1:

mfojtik:
@jeffmc ping?

jmccormick2001:
ping

soltysh:
@jmccormick2001 see @jhadvig [comment](https://github.com/openshift/postgresql/pull/13#issuecomment-105481164) and rebase and squash your changes into one commit please.

mnagy:
@jmccormick2001 apologies for the delay. Will merge this after OS v3 is released.

smarterclayton:
Let's get this merged soon

smarterclayton:
Martin, can you take this and make any remaining fixes necessary (I think the replica example needs to be updated to v1/v1beta3)

mnagy:
Yes, on it.

mnagy:
@jmccormick2001 I've opened #35 as a follow up for this PR. It contains your commit and some additional stuff, mostly updated template and overall repo structure. I'll close this one and let's follow up there.

mgudmund:
I would review the location of the trigger_file.
Putting it in /tmp is bad practice from a security standpoint, no matter if you run it a container or not.

PR #17, opened 2015-05-12
squash on build

bparees:


hack/build.sh
```diff
@@ -9,6 +9,23 @@ declare -a VERSIONS=(9.2)
 OS=$1
 VERSION=$2
 
+function squash {	
+	base=$(grep FROM Dockerfile | awk '{print $2}')
```
mfojtik: grep "^FROM"

hack/build.sh
```diff
@@ -9,6 +9,23 @@ declare -a VERSIONS=(9.2)
...
+	# install the docker layer squashing tool
+	easy_install --user pip
+	rm -rf docker-scripts
+	git clone https://github.com/goldmann/docker-scripts/
```
mfojtik: git clone --branch 0.3.0

hack/build.sh
```diff
@@ -9,6 +9,23 @@ declare -a VERSIONS=(9.2)
...
+	pushd docker-scripts
+	git checkout 0.3.0
+	~/.local/bin/pip install --force --user .
```
mfojtik: s/~/$HOME/

bparees:
@mfojtik thanks, updated here and in the 7 other pulls :)

PR #21, opened 2015-06-01
Don't try to create the directory more then once

mnagy:
This'll make us play nicely with persistent volumes https://github.com/openshift/origin/pull/2644

### DIFF
--- a/9.2/run-postgresql.sh
+++ b/9.2/run-postgresql.sh
@@ -14,7 +14,9 @@ if [ -O $HOME/data ]; then
 else
   # If current user does not own data directory
   # create a subdirectory that the user does own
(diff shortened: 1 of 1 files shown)


mnagy:
@csrwng @mfojtik PTAL

csrwng:
LGTM, though most likely the ownership test above will succeed and there won't be a need to create a subdirectory.

mnagy:
@csrwng I'm testing persistent volumes in the mentioned PR and this was an issue. I used nfs exports, directories owned by root with 777 permissions and got fails with mkdir not succeeding.

Merging.

mnagy:
(Thanks :)

PR #24, opened 2015-06-09
Make psql command work out of box

mnagy:
Required setting some nev vars in the Dockerfile.

@mfojtik @hhorak @jhadvig @csrwng

### DIFF
--- a/9.2/Dockerfile
+++ b/9.2/Dockerfile
@@ -11,7 +11,10 @@ MAINTAINER  Martin Nagy <mnagy@redhat.com>
 #  * $POSTGRESQL_ADMIN_PASSWORD (Optional) - Password for the 'postgres'
 #                           PostgreSQL administrative account
 
-ENV POSTGRESQL_VERSION=9.2
+ENV POSTGRESQL_VERSION=9.2 \
+    HOME=/var/lib/pgsql \
+    PGUSER=postgres \
+    BASH_ENV=/var/lib/pgsql/.bashrc
 
 LABEL k8s.io/description="PostgreSQL is an advanced Object-Relational database management system" \
       k8s.io/display-name="PostgreSQL 9.2" \
--- a/9.2/Dockerfile.rhel7
+++ b/9.2/Dockerfile.rhel7
@@ -11,7 +11,10 @@ MAINTAINER  Martin Nagy <mnagy@redhat.com>
 #  * $POSTGRESQL_ADMIN_PASSWORD (Optional) - Password for the 'postgres'
 #                           PostgreSQL administrative account
 
-ENV POSTGRESQL_VERSION=9.2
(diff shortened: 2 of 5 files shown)


mnagy:
@csrwng says we can do another approach. Not setting PGUSER in Dockerfile, but setting LD_PRELOAD in dockerfile. Which is better/worse?

@bparees @mfojtik ?

mfojtik:
i would say LD_PRELOAD is worse

csrwng:
@mnagy I guess I misunderstood, can we do either PGUSER or LD_PRELOAD ( I thought we needed both) ? if things work ok with PGUSER, then I think  that's fine as well.

mnagy:
@csrwng yes, we only need PGUSER and it works fine.

mfojtik:
@csrwng @mnagy I don't think we are going to run postgresql with other user than PGUSER.
@csrwng does the artibrary UID apply to databases as well? I guess yes as we mount volumes... in that case we just need to make sure PGUSER maps to UID

csrwng:
@mfojtik we won't know the uid until the container is running... when can we set it?

mfojtik:
@csrwng if PGUSER will be set to 'postgres' then we just need to make sure it maps to that UID. I guess we do that.

mnagy:
@mfojtik it will not map to it during docker exec. Do you want that? I don't think it's necessary just to run psql. I guess it might be needed for other debugging work..? Not sure.

mfojtik:
@mnagy ok, maybe I misunderstood, so why do we need to set PGUSER? Is that needed for .bashrc to work? Is it needed for `docker exec CID /bin/bash -c psql` ? Can we just export it in .bashrc instead of setting it globally in container?

csrwng:
@mfojtik I don't think .bashrc gets picked up when you do docker exec, but would need to check

mfojtik:
it is not... but we are trying to make docker exec /bin/bash -c psql work
with SCL...

without  /bin/bash we will need a bash wrapper script....

csrwng:
@mfojtik hmm, seems that maybe it could be easier to do the LD_PRELOAD? What could go wrong :-)

mnagy:
@mfojtik PGUSER is needed for psql. Otherwise it dies with an error about user not specified.

mnagy:
@mfojtik Good idea. We could export in .bashrc as wel. It will get picked up with bash -c, since BASH_ENV is now set.

bparees:
so it doesn't matter that pguser isn't going to be resolvable in /etc/passwd as the current uid when we run as a random uid?

mnagy:
Not when tou run psql, from what I've tested. No reason it should, it just
uses the user as a name when connecting.

mfojtik:
LGTM

PR #25, opened 2015-06-10
Added mandatory labels

eliskasl:
@mfojtik @brenton, please have a look

### DIFF
--- a/9.2/Dockerfile.rhel7
+++ b/9.2/Dockerfile.rhel7
@@ -18,6 +18,12 @@ LABEL k8s.io/description="PostgreSQL is an advanced Object-Relational database m
       openshift.io/expose-services="5432:postgresql" \
       openshift.io/tags="database,postgresql,postgresql92"
 
+# Labels consumed by Red Hat build service
+LABEL Component="postgresql92" \
+      Name="openshift3/postgresql-92-rhel7" \
+      Version="9.2" \
+      Release="1"
+
 EXPOSE 5432
 
 COPY run-postgresql.sh /usr/local/bin/


9.2/Dockerfile.rhel7
```diff
@@ -16,7 +16,11 @@ ENV POSTGRESQL_VERSION=9.2
 LABEL k8s.io/description="PostgreSQL is an advanced Object-Relational database management system" \
       k8s.io/display-name="PostgreSQL 9.2" \
       openshift.io/expose-services="5432:postgresql" \
-      openshift.io/tags="database,postgresql,postgresql92"
+      openshift.io/tags="database,postgresql,postgresql92" \
+      LABEL BZComponent="postgresql92" \
```
mfojtik: nuke the `LABEL` instruction (the LABEL support multi-line assignment)

mfojtik:
@eliskasl please add comment to Dockerfile explaining why we need this labels.

mfojtik:
@eliskasl actually I think it would be better to move RCM labels into new layer (we are going to squash that anyway). But it makes it more readable in sense that we know who owns what labels. Also you can add comment on top of that layer, saying that these labels belongs to RCM.

mfojtik:
cc @bparees

mfojtik:
LGTM

bparees:
same question here on whether openshift3 is the right value for the name field, vs openshift.

brenton:
As best I can understand our official naming policy I believe `/openshift3` is the most correct.  For example we were asked to change from `/openshift` to `/openshift3_beta` for the beta period.

The `BZComponent` label is new to me.  I thought it was simply `Component`.

bparees:
ok, that has an impact on our templates:
https://github.com/openshift/origin/blob/master/examples/image-streams/image-streams-rhel7.json#L14
(I guess they needed to be updated anyway)

so if we're sure it's going to be openshift3 i'll start updating them.

brenton:
thx ben.  I think it's safe to update those templates now.

mfojtik:
@brenton @bparees we still use Docker Hub centos images in those templates so I think we should also get the "openshift3" namespace in Docker Hub.

eliskasl:
BZComponent changed to Component. Thanks Brenton!

PR #29, opened 2015-06-12
Pulling nss_wrapper from RHEL

brenton:


9.2/Dockerfile.rhel7
```diff
@@ -38,9 +38,7 @@ RUN yum install -y yum-utils gettext && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
     yum install -y --setopt=tsflags=nodocs postgresql92 && \
-    curl -O https://dl.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-5.noarch.rpm && \
-    rpm -Uvh epel-release-7-5.noarch.rpm && rm epel-release-7-5.noarch.rpm && \
-    yum install -y --enablerepo=epel --setopt=tsflags=nodocs nss_wrapper && \
+    yum install -y --disablerepo="epel" --setopt=tsflags=nodocs nss_wrapper && \
```
brenton: For now I think the safest thing to do is explicitly disable epel in the RHEL build.  This will likely cause OSBS builds to fail since they won't have the repo and yum will bail.  I'm a little worried about epel being available and some day shipping a newer version of some of these packages and silently shadowing the build we really want.

brenton:
@mfojtik, @bparees, PTAL

bparees:
thanks @brenton!

PR #42, opened 2015-07-27
Generate locale archive

mnagy:
Recent RHEL7 image doesn't include the locale archive. Generate it
during Docker build. This does not affect the CentOS image, however,
generate it there as well just to be safe for the future.

Fixes RH bug 1245444.


mnagy:
@bparees please review?

bparees:
lgtm, thanks for the quick fix @mnagy!

PR #45, opened 2015-08-03
update maintainer and remove dead code

bparees:


bparees:
@mfojtik ptal pending maintainer

PR #47, opened 2015-08-07
switch to an arbitrary USER

bparees:


bparees:
@mnagy ptal

hhorak:
@bparees there is not reasoning why we should do this, please explain.

mnagy:
@hhorak The reason behind this PR is that the database images are supposed to be able to run under an arbitrary user ID.

Let's keep the discussion at one place, please: https://github.com/openshift/mongodb/pull/67

PR #48, opened 2015-08-18
Improve replication documentation

rhcarvalho:
@bparees @mfojtik @mnagy PTAL

### DIFF
--- a/9.2/contrib/common.sh
+++ b/9.2/contrib/common.sh
@@ -12,7 +12,7 @@ function usage() {
   if [ $# == 2 ]; then
     echo "error: $1"
   fi
-  echo "You must specify following environment variables:"
+  echo "You must specify the following environment variables:"
   echo "  POSTGRESQL_USER (regex: '$psql_identifier_regex')"
   echo "  POSTGRESQL_PASSWORD (regex: '$psql_password_regex')"
   echo "  POSTGRESQL_DATABASE (regex: '$psql_identifier_regex')"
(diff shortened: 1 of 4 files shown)


bparees:
should there be some discussion of what "recovery" looks like for this, in case you lose the master?

rhcarvalho:
I think for this template we don't aim to facilitate promoting a slave to master. To do that, one would need at least to edit a running slave pod to change labels (so that it is picked up by the `postgresql-master` service), plus PostgreSQL recovery steps.
Unfortunately the slaves use ephemeral storage, so promoting them to master means giving up the persistent storage until something like https://github.com/kubernetes/kubernetes/issues/260 lands (same problem as in MongoDB replication, how to have *N* replicated pods with independent persistent storage).

What is left about recovery is Kubernetes rescheduling the postgresql-master pod in a new node (in case the problem was a whole node down). In that case, you have downtime for writes for as long as rescheduling and restarting takes, while no read downtime (if reading from slaves). Recovery only makes sense if you have persistent storage for the master (still need to add it to this replication example).

Am I missing something?

rhcarvalho:
Is there any other recovery scenario we can write about other than the standard healing behavior of k8s?

bparees:
so basically the value of using this comes down to:
you can enable read from slave.

if you don't do that, there is no situation in which this provides any value, right?

rhcarvalho:
There's the original idea that these replication examples were meant for "backup", though in this case we're backing up data in ephemeral storage...

So other than ephemeral backup, proof of "yes, it is possible to setup replication in OpenShift" and reading from slaves, I don't see other uses right now.

bparees:
k, thanks.  lgtm.

PR #49, opened 2015-08-18
Make replication template use persistent storage

rhcarvalho:
@bparees @mfojtik @mnagy PTAL

### DIFF
--- a/examples/replica/README.md
+++ b/examples/replica/README.md
@@ -23,8 +23,40 @@ A standby server can also be used for read-only queries.
 
 ## Deployment
 
-To create a PostgreSQL replicated database cluster in OpenShift v3, you can use the template
-included in this repository and create a new deployment right away:
+This example uses a [PersistentVolumeClaim](https://docs.openshift.org/latest/architecture/additional_concepts/storage.html#persistent-volume-claims)
+to request persistent storage for the primary PostgreSQL server.
+
+You need to have persistent volumes configured and available in your project in
+order to continue. For trying out this example in a single node testing
+environment, you can create a temporary volume with:
+
+```
+$ oc create -f - <<EOF
+{
+  "kind": "PersistentVolume",
+  "apiVersion": "v1",
+  "metadata": {
+    "name": "postgres-data-volume"
+  },
+  "spec": {
+    "capacity": {
+      "storage": "512Mi"
+    },
+    "hostPath": {
+      "path": "`mktemp -d --tmpdir pg-data.XXXXX | tee >(xargs chmod a+rwx)`"
+    },
+    "accessModes": [
+      "ReadWriteOnce"
+    ]
+  }
+}
+EOF
+```
+
+It is recommended, however, that you use other [type of PersistentVolume](https://docs.openshift.org/latest/architecture/additional_concepts/storage.html#types-of-persistent-volumes)
+such as NFS.
+
+Now you can create a new database deployment:
 
 ```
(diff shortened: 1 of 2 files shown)


bparees:
i thought this won't work because the volume is shared by all the instances?  won't they clobber each other's data?

rhcarvalho:
This is different than MongoDB. The layout here is 2 services and 2 DCs, master and slave. We can have persistent volume in the master, as it is not meant to be scaled. Slaves replicate to ephemeral storage because of the limitation on how to specify their storage independently.

In MongoDB there is a single service and single DC.

examples/replica/README.md
```diff
@@ -23,8 +23,40 @@ A standby server can also be used for read-only queries.
...
+      "storage": "512Mi"
+    },
+    "hostPath": {
+      "path": "`mktemp -d --tmpdir pg-data.XXXXX | tee >(xargs chmod a+rwx)`"
+    },
+    "accessModes": [
+      "ReadWriteOnce"
+    ]
+  }
+}
```
bparees: nice!

bparees:
sorry, missed that you were only changing the master service.

would it make sense to use an emptydir volume for the slave pods for now, so we're in a good position to switch to persistent in the future?

bparees:
otherwise lgtm.

rhcarvalho:
@bparees you mean adding the volume without a claim? We could do that. Do you know if Docker will spare a volume in `/var/lib/docker/{volumes,vfs}` in that case?
Last week I ran into a problem of docker eating 25 GB of disk on useless old volumes.

bparees:
i mean like this:
            "volumes": [
              {
                "name": "${JENKINS_SERVICE_NAME}-data",
                "emptyDir": {
                  "medium": ""
                }
              }

rhcarvalho:
will that use the same "emptyDir" for every replica?

bparees:
every pod will get a unique empty directory volume.

rhcarvalho:
if the volumes will be shared by all replicas, then it won't work... I will try.

rhcarvalho:
@bparees emptyDir volume added.

bparees:
lgtm, will merge after tests pass.

rhcarvalho:
Answering the doubt I had about were the data of the emptyDir volumes go...

```
# tree -L 2 /home/vagrant/openshift.local.volumes/pods/**/volumes/kubernetes.io~empty-dir
/home/vagrant/openshift.local.volumes/pods/54c0002f-45d6-11e5-bf45-080027c5bfa9/volumes/kubernetes.io~empty-dir
└── postgresql-data
    └── userdata
/home/vagrant/openshift.local.volumes/pods/9f741235-45d6-11e5-bf45-080027c5bfa9/volumes/kubernetes.io~empty-dir
└── postgresql-data
    └── userdata
/home/vagrant/openshift.local.volumes/pods/9f77d3fd-45d6-11e5-bf45-080027c5bfa9/volumes/kubernetes.io~empty-dir
└── postgresql-data
    ├── slave-1         # I've created this file to test that the volumes of each replica are indeed independent
    └── userdata
/home/vagrant/openshift.local.volumes/pods/b550c0d0-41da-11e5-8411-080027c5bfa9/volumes/kubernetes.io~empty-dir
└── registry-storage
```

It goes under `openshift.local.volumes`, the place where the OpenShift server stores volumes (the actual path varies depending on how the server was started).

PR #50, opened 2015-08-19
Support change passwords flow via env vars

rhcarvalho:
@bparees @mfojtik @mnagy PTAL

### DIFF
--- a/9.2/run-postgresql.sh
+++ b/9.2/run-postgresql.sh
@@ -29,8 +29,13 @@ fi
 generate_postgresql_config
 generate_passwd_file
 
-if [ "$1" = "postgres" -a ! -f "$PGDATA/postgresql.conf" ]; then
-  initialize_database
+if [ "$1" = "postgres" ]; then
+  if [ ! -f "$PGDATA/postgresql.conf" ]; then
+    initialize_database
+  fi
+  pg_ctl -w start
+  set_passwords
+  pg_ctl stop
 fi
 
 unset_env_vars
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ $ docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL
 This will create a container named `postgresql_database` running PostgreSQL with
 database `db` and user with credentials `user:pass`. Port 5432 will be exposed
 and mapped to the host. If you want your database to be persistent across container
-executions, also add a `-v /host/db/path:/var/lib/pgsql/data` argument. This will be 
+executions, also add a `-v /host/db/path:/var/lib/pgsql/data` argument. This will be
 the PostgreSQL database cluster directory.
 
 If the database cluster directory is not initialized, the entrypoint script will
@@ -114,6 +114,19 @@ The admin account `postgres` has no password set by default, only allowing local
 connections.  You can set it by setting the `POSTGRESQL_ADMIN_PASSWORD` environment
 variable when initializing your container. This will allow you to login to the
 `postgres` account remotely. Local connections will still not require a password.
+
+
+Changing passwords
+------------------
+
+When you use a persistent volume to store your data, you may want to update
+passwords of an existing database. Since passwords are part of the image
+configuration, you are not supposed to change any database password through
+SQL statements or any database client whatsoever.
+
+The only supported method to change passwords for the database user (defined
+by `POSTGRESQL_USER`) and admin ('postgres') is by changing the environment
+variables `POSTGRESQL_PASSWORD` and `POSTGRESQL_ADMIN_PASSWORD`, respectively.
 
 
 Test
--- a/9.2/contrib/common.sh
+++ b/9.2/contrib/common.sh
@@ -107,10 +107,20 @@ EOF
   pg_ctl -w start
   createuser "$POSTGRESQL_USER"
   createdb --owner="$POSTGRESQL_USER" "$POSTGRESQL_DATABASE"
-  psql --command "ALTER USER \"${POSTGRESQL_USER}\" WITH ENCRYPTED PASSWORD '${POSTGRESQL_PASSWORD}';"
 
   if [ -v POSTGRESQL_MASTER_USER ]; then
     createuser "$POSTGRESQL_MASTER_USER"
+  fi
+
+  set_passwords
+
+  pg_ctl stop
+}
+
+function set_passwords() {
+  psql --command "ALTER USER \"${POSTGRESQL_USER}\" WITH ENCRYPTED PASSWORD '${POSTGRESQL_PASSWORD}';"
+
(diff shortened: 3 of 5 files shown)


bparees:
@rhcarvalho this is going to set the password every time the pod starts, isn't it?  (even if it's unchanged)

that means if someone edits the password manually and then restarts the pod, the password will be reverted.

here's how i tackled this in jenkins:

https://github.com/openshift/jenkins/blob/master/1.6/run.sh#L29-L34

basically it stores a hash of the password set by the environment variable.  on startup it checks if the hash of the current env value matches the old hash.  if it doesn't, only then does it change the password on the volume.

rhcarvalho:
@bparees I see how you might have arrived into that path with Jenkins, as there is an easy-to-spot way to change the password via UI.

But looking back at the [Trello card](https://trello.com/c/ZNMI7j81) that originates and motivates this PR,    I fear that going into that direction will make implementation more complex, and the whole flow harder to document. I didn't see much documented on [our Jenkins](https://github.com/openshift/jenkins) apart from mentioning the env var `JENKINS_PASSWORD`; specifically, there's nothing stating that you can use it to change Jenkins' password, nor what will be the state of the system once you change it via UI.

The card specifically mentions the "problem" of having the pod template contain an outdated password. It's not much of a problem if we remember that the password is not required to run the container after the first run, but it is a problem of inconsistency -- so I understand consistency is the reason why we went on to propose that

> the user will **have to** change the passwords manually via `oc edit`

As a side note, instead of suggesting `oc edit` in the docs, I went on with `oc env` which is a less error-prone and easier to document solution (and a "must-have" for a pain-free change of `POSTGRESQL_MASTER_PASSWORD`).

I'm willing to update my opinion in the light of new specs and an updated description in the card.

mfojtik:
@rhcarvalho we need to document what consequences the users should expect if he changes the password manually (by connecting to MySQL or via phpMyAdmin?). If user does that, we need to document what he should expect when `mysql` container is 'redeployed' (iow. you app will not be able to connect because env vars are wrong).

rhcarvalho:
@mfojtik hmm... the env vars in the "app" are independent than those of the ones in the postgresql-master/slave DCs... -- different DCs won't keep common vars in sync. Either that or we have to define better what is "app".

What should the user expect when the `postgresql-master` is redeployed after she changed the password manually?

mfojtik:
@bparees I think we are missing something like 'another deployment changed trigger' in deployment ;-) The use case here is:

1. I have two deployments: "mysql" and "app"
2. I change password in "mysql" deployment (oc env)
3. The "mysql" will redeploy with updated password
4. The "app" is now broken as it has old environment variables for MySQL

What I think we should do is "app" deployment has "DeploymentChangeTrigger" set to "mysql" and it redeploys the "app" everytime the "mysql" changes.

Maybe we already have something like this, in that case nvmd this comment  ;-)

rhcarvalho:
@mfojtik redeploying the "app" is not enough, we'd need also to update the env var(s). And if the password was changed outside of OpenShift's eyes, there's not much we could do to update env vars anywhere :)

mfojtik:
@rhcarvalho redeploying an app means it pick up a new env vars if they are provided by the mysql service (or shared across project)

rhcarvalho:
@mfojtik it is not possible to do that right now. There's no such thing as "project env vars" nor any way to pick up vars from a different DC. I think there are Trello cards for those features, but AFAIK nothing implemented.

rhcarvalho:
@bparees before we merge this, I'll move the main bits about updating passwords to the top-level README, and document the effects of trying to change the password by other means.

rhcarvalho:
Added `Changing passwords` section to the main README and mentioned that no other method to change passwords is supported in replica/README and "On every deploy, passwords are reset to match the values in the environment variables of the DeploymentConfig 'postgresql-master'."

@bparees PTAL again :pray:

rhcarvalho:
Hmm, need one more fix to support the standalone case.

9.2/run-postgresql.sh
```diff
@@ -33,5 +33,9 @@ if [ "$1" = "postgres" -a ! -f "$PGDATA/postgresql.conf" ]; then
   initialize_database
 fi
 
+pg_ctl -w start
+set_passwords
+pg_ctl stop
+
```
rhcarvalho: Sadly there are two code flows for postgresql-master (replication example) and standalone.
We can look into refactoring this later.

rhcarvalho:
[test]

rhcarvalho:
@bparees we should be good to go from my side.

README.md
```diff
@@ -116,6 +116,19 @@ variable when initializing your container. This will allow you to login to the
...
+When you use a persistent volume to store your data, you may want to update
+passwords of an existing database. Since passwords are part of the image
+configuration, you are not supposed to change any database password through
+SQL statements or any database client whatsoever.
+
+The only supported method to change passwords for the database user (defined
+by `POSTGRESQL_USER`) and admin ('postgres') is by changing the environment
+variables `POSTGRESQL_PASSWORD` and `POSTGRESQL_ADMIN_PASSWORD`, respectively.
+
+
```
bparees: @rhcarvalho  this will need to go into the openshift docs too.

bparees:
failed rhel testing.
[test]

bparees:
[test]

bparees:
@rhcarvalho looks like you're going to need to look into the rhel failures.

rhcarvalho:
I have no idea about what is actually failing.... [test]
Don't know how the changes in this PR could work in centos and not in rhel.

rhcarvalho:
Ok ok... way down in the output I see:

```
pg_ctl: could not start server
Examine the log output.
```

Hmmm, now on to investigate why it would fail to start the server and where might be the "log output"...

README.md
rhcarvalho: Yep

tnguyen-rh:
On my (old, pre-systemd) Debian system, i see that Postgres 8.4 keeps logs in /var/log/postgresql.

mfojtik:
@rhcarvalho @tnguyen-rh we should log everything to stdout

rhcarvalho:
[test] with some debugging code added...

rhcarvalho:
please [test] again

rhcarvalho:
Test passed for RHEL, now running on CentOS. Seems I fixed the problem :cookie:

rhcarvalho:
:green_heart: green tests, going home!

bparees:
lgtm, i'm going to merge, but @rhcarvalho it would be good to add a test around this, like in jenkins:

https://github.com/openshift/jenkins/blob/master/1.6/test/run#L122-L136

rhcarvalho:
@bparees FYI I'm looking into writing such a test.

PR #51, opened 2015-08-24
Refactor entrypoint

### LINKED ISSUE #6: Don't open port during initialization
When creating the user, we shouldn't open the postgres port.

rhcarvalho:
Unify flows for standalone execution and master of a master-slave setup.

This allows starting with standalone and later adding slaves, as well as remove one extra code path to maintain.

Also fixes #6.

Commits are split logically, but I can rearrange them if needed.

### DIFF
--- a/9.2/run-postgresql-master.sh
+++ b/9.2/run-postgresql-master.sh
@@ -5,19 +5,11 @@ set -eu
 source ${HOME}/common.sh
 
 check_env_vars
-
-generate_postgresql_config
 generate_passwd_file
+generate_postgresql_config
 
 if [ ! -f "$PGDATA/postgresql.conf" ]; then
   initialize_database
-  cat >> "$PGDATA/pg_hba.conf" <<EOF
-
-# Allow replication connections from all hosts.
-host replication all all md5
-EOF
-  # FIXME: ^^^ only allowing replication connections from specific hosts
-  #            would be a nice-to-have.
 fi
 
 pg_ctl -w start
--- a/9.2/run-postgresql-master.sh
+++ b/9.2/run-postgresql-master.sh
@@ -12,9 +12,7 @@ if [ ! -f "$PGDATA/postgresql.conf" ]; then
   initialize_database
 fi
 
-pg_ctl -w start
 set_passwords
-pg_ctl stop
 
 unset_env_vars
 exec postgres "$@"
--- a/9.2/run-postgresql.sh
+++ b/9.2/run-postgresql.sh
@@ -2,8 +2,6 @@
 
 set -eu
 
-source ${HOME}/common.sh
-
 # Data dir
 if [ -O $HOME/data ]; then
   export PGDATA=$HOME/data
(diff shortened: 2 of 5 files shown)


rhcarvalho:
@bparees @mfojtik @mnagy PTAL.

[test]

9.2/run-postgresql-master.sh
```diff
@@ -2,27 +2,16 @@
 
 set -eu
 
-source ${HOME}/common.sh
+source $HOME/common.sh
```
mfojtik: it is a good practice in bash to use "${HOME}"

bparees:
lgtm.

9.2/run-postgresql-master.sh
rhcarvalho: Hmm, I noticed we use `${HOME}` in some places, `$HOME` in others -- while the curly braces are not needed in that particular case.
What's this good practice about?

rhcarvalho:
[test] again, please

9.2/run-postgresql-master.sh
mfojtik: https://google-styleguide.googlecode.com/svn/trunk/shell.xml#Variable_expansion

rhcarvalho:
@bparees seems we have some infra problem with tests, failure:

> This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.

9.2/run-postgresql-master.sh
rhcarvalho: Thanks for pointing that out. I didn't know we followed anything specific like that.

> In order of precedence: Stay consistent with what you find; ...; prefer "${var}" over "$var"

In multiple places in the same file and group of files we're using a mix of both styles. @mfojtik @bparees what do you think about a follow up making those consistent?!

One way or another is fine for me, as long as we keep ourselves consistent.

mfojtik: not a high prio... i am fixing those when i touching files :)

rhcarvalho:
[test]

9.2/run-postgresql.sh
```diff
@@ -16,27 +14,13 @@ else
...
-if [ "$1" = "postgres" ]; then
-  if [ ! -f "$PGDATA/postgresql.conf" ]; then
-    initialize_database
-  fi
-  pg_ctl -w start
-  set_passwords
-  pg_ctl stop
-fi
+case "${1:-}" in
+  'postgres-slave' )
```
mfojtik: `postgresql-slave` keep it consistent

9.2/test/run
```diff
@@ -198,6 +197,12 @@ function run_tests() {
   if [ -v ADMIN_PASS ]; then
     envs="$envs -e POSTGRESQL_ADMIN_PASSWORD=$ADMIN_PASS"
   fi
+  if [ -v POSTGRESQL_MAX_CONNECTIONS ]; then
+    envs="$envs -e POSTGRESQL_MAX_CONNECTIONS=$POSTGRESQL_MAX_CONNECTIONS"
+  fi
+  if [ -v POSTGRESQL_SHARED_BUFFERS ]; then
+    envs="$envs -e POSTGRESQL_SHARED_BUFFERS=$POSTGRESQL_SHARED_BUFFERS"
```
mfojtik: envs+=" -e POSTGRESQL_"

9.2/run-postgresql.sh
rhcarvalho: not so simple, see https://github.com/openshift/postgresql/pull/51#discussion-diff-38196215L24
It was already spelled that way :/

mfojtik: I can't see the original comment :-) Why it is not simple? I guess this is just a helper to start the postgresql image in the "right" mode (iow. a `Cmd` string).

rhcarvalho: This is backward incompatible change.
Before `docker run ... postgresql-92-centos7 postgres-slave` (or postgres-master) would work, if we change it, it won't.
Existing deployed templates refer to the old name, if they get a new image is breaks the deployment.

If you don't care about anybody using this image in the wild... then it is indeed "simple": change it here and in the template, and perhaps in documentation if there is any.

rhcarvalho:
please [test] again

rhcarvalho:
and now... [test]

mfojtik:
LGTM [merge]

PR #52, opened 2015-08-24
Add test case for changing passwords

rhcarvalho:
This is a follow up to #50 

@mfojtik @bparees PTAL

### DIFF
--- a/9.2/test/run
+++ b/9.2/test/run
@@ -218,6 +218,72 @@ function run_tests() {
   test_postgresql $name
 }
 
+function run_change_password_test() {
+  local name="change_password"
+
+  local database='db'
+  local user='user'
+  local password='password'
+  local admin_password='adminPassword'
+
+  local volume_dir
+  volume_dir=`mktemp -d --tmpdir pg-testdata.XXXXX`
+  chmod a+rwx ${volume_dir}
+
+  DOCKER_ARGS="
+-e POSTGRESQL_DATABASE=${database}
+-e POSTGRESQL_USER=${user}
+-e POSTGRESQL_PASSWORD=${password}
+-e POSTGRESQL_ADMIN_PASSWORD=${admin_password}
+-v ${volume_dir}:/var/lib/pgsql/data
+" create_container ${name}
+
+  # need to set these because `postgresql_cmd` relies on global variables
+  USER=${user}
+  PASS=${password}
+
+  # need this to wait for the container to start up
+  CONTAINER_IP=$(get_container_ip ${name})
+  test_connection ${name}
+
+  echo "  Testing login"
+
+  assert_login_access ${user} ${password} true
+  assert_login_access 'postgres' ${admin_password} true
+
+  echo "  Changing passwords"
+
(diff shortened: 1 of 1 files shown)


9.2/test/run
```diff
@@ -212,6 +212,71 @@ function run_tests() {
...
+  test_connection ${name}
+
+  echo "  Testing login"
+  assert_login_access ${user} ${password} true
+  USER='postgres'
+  PASS=${admin_password}
+  assert_login_access 'postgres' ${admin_password} true
+
+  echo "  Changing passwords"
+  docker stop $(get_cid ${name})
```
rhcarvalho: For some reason my local tests fail after this, and I cannot start the new container getting:

> FATAL:  configuration file "/var/lib/pgsql/data/userdata/postgresql.conf" contains errors

Quickly glanced the file and it seemed ok, though.

9.2/test/run
```diff
@@ -212,6 +212,71 @@ function run_tests() {
...
+
+  # need this to wait for the container to start up
+  CONTAINER_IP=$(get_container_ip "${name}_NEW")
+  test_connection "${name}_NEW"
+
+  echo "  Testing login with new passwords"
+  assert_login_access ${user} "NEW_${password}" true
+  USER='postgres'
+  PASS="NEW_${admin_password}"
+  assert_login_access 'postgres' "NEW_${admin_password}" true
```
bparees: make sure the old password is invalid, too?

9.2/test/run
```diff
@@ -212,6 +212,71 @@ function run_tests() {
...
+  docker stop $(get_cid ${name})
+  DOCKER_ARGS="
+-e POSTGRESQL_DATABASE=${database}
+-e POSTGRESQL_USER=${user}
+-e POSTGRESQL_PASSWORD=NEW_${password}
+-e POSTGRESQL_ADMIN_PASSWORD=NEW_${admin_password}
+-v ${volume_dir}:/var/lib/pgsql/data
+" create_container "${name}_NEW"
+
+  # need to set these because `postgresql_cmd` relies on global variables
```
rhcarvalho: This is ugly, but needed for now before I go on and fix existing code (we, in fact, never test the admin password, but test the user password all the time).

9.2/test/run
rhcarvalho: I'd say in general if one password works, the other, by exclusion, doesn't. But in this reality we have here testing both can make sense...

I'm almost sure `assert_login_access` always return 0, though. So all those tests always pass...

rhcarvalho: Now we check new and old passwords

jhadvig:
Shouldn't we step away from the bash test monster that we are slowly feeding in our images and rather focus on rewriting them to extended-tests, since we finally got them ?

bparees:
@jhadvig yes but first we'd need to actually change our image test scripts to run the extended tests on openshift.  running the extended tests in origin doesn't help us test image repo PRs.

so i think we're still a bit away from converting everything into extended tests.

bparees:
lgtm.

PR #53, opened 2015-08-24
Make assert_login_access fail when it should fail

### LINKED ISSUE #54: assert_login_access always pass when success=false
[`assert_login_access`](https://github.com/openshift/postgresql/blob/f12361c74a5818762220d91c9d92818cbac7d1e0/9.2/test/run#L99) always succeed.

That means that assertions like [`assert_login_access $USER "${PASS}_foo" false`](https://github.com/openshift/postgresql/blob/f12361c74a5818762220d91c9d92818cbac7d1e0/9.2/test/run#L202) never fail.

rhcarvalho:
Our `assert_login_access` used to always return 0 no matter what.

Fixes #54

### DIFF
--- a/9.2/test/run
+++ b/9.2/test/run
@@ -90,14 +90,26 @@ function assert_login_access() {
   local USER=$1 ; shift
   local PASS=$1 ; shift
   local success=$1 ; shift
+  (
+    set +e
+    postgresql_cmd <<< "SELECT 1;"
+    status=$?
 
-  if $success; then
-    postgresql_cmd <<< "SELECT 1;" &&
-      echo "    $USER($PASS) access granted as expected"
-  else
-    postgresql_cmd <<< "SELECT 1;" ||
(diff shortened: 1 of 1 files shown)


rhcarvalho:
[test] again

mfojtik:
yet another reason why not use bash and move these tests to extended tests in origin.

rhcarvalho:
@mfojtik yep... that's my point :) I uncovered at least two bugs yesterday.

rhcarvalho:
@bparees @mfojtik PTAL

bparees:
lgtm, will get it tested after the new ami is available (building now)

mfojtik:
LGTM (but I really wish we have these test as extended test ;-)

bparees:
[test]

bparees:
[test]

bparees:
lgtm, passed tests.

PR #55, opened 2015-08-26
Refactor assert_login_access

rhcarvalho:
No need to run in a subshell nor to use `set +e`.

Follows up #53 

@bparees @mfojtik PTAL


bparees:
much nicer, thanks!

PR #57, opened 2015-08-26
Reorder Dockerfile for better use of cache

rhcarvalho:
Every change to the scripts was causing yum install to run over and over.

@bparees PTAL

### DIFF
--- a/9.2/Dockerfile.rhel7
+++ b/9.2/Dockerfile.rhel7
@@ -29,9 +29,6 @@ LABEL Component="postgresql92" \
 
 EXPOSE 5432
 
-COPY run-*.sh /usr/local/bin/
-COPY contrib /var/lib/pgsql/
-
 # This image must forever use UID 26 for postgres user so our volumes are
 # safe in the future. This should *never* change, the last test is there
 # to make sure of that.
@@ -44,7 +41,14 @@ RUN yum install -y yum-utils gettext && \
     localedef -f UTF-8 -i en_US en_US.UTF-8 && \
     mkdir -p /var/lib/pgsql/data && chown postgres.postgres /var/lib/pgsql/data && \
     test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)" && \
-    chmod -R go+rwx /var/lib/pgsql && chmod go+w /var/run/postgresql
+    # Loosen permission bits to avoid problems running container with arbitrary UID
+    chmod -R a+rwx /var/run/postgresql
+
+COPY run-*.sh /usr/local/bin/
+COPY contrib /var/lib/pgsql/
+
+# Loosen permission bits to avoid problems running container with arbitrary UID
+RUN chmod -R a+rwx /var/lib/pgsql
 
 # When bash is started non-interactively, to run a shell script, for example it
 # looks for this variable and source the content of this file. This will enable
@@ -52,8 +56,6 @@ RUN yum install -y yum-utils gettext && \
 ENV BASH_ENV=/var/lib/pgsql/scl_enable \
     ENV=/var/lib/pgsql/scl_enable \
     PROMPT_COMMAND=". /var/lib/pgsql/scl_enable"
-
-
 
 VOLUME ["/var/lib/pgsql/data"]
 
--- a/9.2/Dockerfile
+++ b/9.2/Dockerfile
@@ -23,9 +23,6 @@ LABEL io.k8s.description="PostgreSQL is an advanced Object-Relational database m
 
 EXPOSE 5432
 
-COPY run-*.sh /usr/local/bin/
-COPY contrib /var/lib/pgsql/
-
 # This image must forever use UID 26 for postgres user so our volumes are
 # safe in the future. This should *never* change, the last test is there
 # to make sure of that.
@@ -37,7 +34,14 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     localedef -f UTF-8 -i en_US en_US.UTF-8 && \
     mkdir -p /var/lib/pgsql/data && chown postgres.postgres /var/lib/pgsql/data && \
     test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)" && \
-    chmod -R go+rwx /var/lib/pgsql && chmod go+w /var/run/postgresql
+    # Loosen permission bits to avoid problems running container with arbitrary UID
+    chmod -R a+rwx /var/run/postgresql
+
+COPY run-*.sh /usr/local/bin/
+COPY contrib /var/lib/pgsql/
+
+# Loosen permission bits to avoid problems running container with arbitrary UID
+RUN chmod -R a+rwx /var/lib/pgsql
 
 # When bash is started non-interactively, to run a shell script, for example it
 # looks for this variable and source the content of this file. This will enable
@@ -45,7 +49,6 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
 ENV BASH_ENV=/var/lib/pgsql/scl_enable \
     ENV=/var/lib/pgsql/scl_enable \
     PROMPT_COMMAND=". /var/lib/pgsql/scl_enable"
-
 
 VOLUME ["/var/lib/pgsql/data"]
 


9.2/Dockerfile.rhel7
```diff
@@ -46,6 +43,9 @@ RUN yum install -y yum-utils gettext && \
     test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)" && \
     chmod -R go+rwx /var/lib/pgsql && chmod go+w /var/run/postgresql
 
+COPY run-*.sh /usr/local/bin/
+COPY contrib /var/lib/pgsql/
```
bparees: i don't think you can do this..  you're chmod'ing the directory before you copy stuff into it now, which means the things you copy into it are not getting chmod'd.

you *could* break the RUN up to do the chmod after, but that adds layers and build time.  

how i've handled this in the past is to reorder it while i'm working on the image/scripts, but then put it back before committing the changes.  i know that's a bit of a pain, but it maintains a cleaner dockerfile.

jhadvig: Agreed, we cant move copying the files to the /var/lib/.../ after we chmod the dir.

I think we can probably close this.

rhcarvalho:
Here's how the file permissions look like after a docker build: https://gist.github.com/rhcarvalho/a1a7add390aa6e0624a7

As @mfojtik mentioned yesterday, COPY will keep the permissions of existing files in our repo. If we want to have write bits set in `common.sh`, `openshift-custom-postgresql.conf.template`, `openshift-custom-recovery.conf.template`, `passwd.template` and `scl_enable` we can just change it in our repo and benefit from faster builds, but those write bits are not strictly necessary to have a working image -- all templates just need to be readable, same for `common.sh` and `scl_enable`.

jhadvig:
Strictly speaking the build wont be faster. It will take the same amount of time, but if we are speaking about development it would save time due to the cache. 
On the other hand I don't like the idea of having files with different ownership and permissions in the `/var/lib/.../`. Then we need to keep in the back of our had that we don't have full control of the directory in our images and personally i think that if we introduce this change to all our images, I'm pretty sure that we will need to rollback the change sooner or later(at least for some images) which will lead to inconsistency.

I'd rather would keep in mind that for the development I should consider moving the COPY instructions after the RUN (if it doesn't break anything) then keeping in mind that we are not masters of our images.

rhcarvalho:
@jhadvig notice that the current Dockerfile do not change the ownership (`chown`), just the permission bits (`chmod`). Before and after the ownerships are the same... please let's take that out of the equation.

A few weeks ago you were proposing changes to improve the life of those working on these repos, and now seems you're going against it? What's up, my friend?

> I'd rather would keep in mind that for the development I should consider moving the COPY instructions after the RUN (...)

Keeping in our brains specific undocumented tricks for multiple repos we work on is counter-productive, unnecessary overhead.

IMHO assuming a developer will out of blue figure out those tricks also discourages any contribution from the open source community.

jhadvig:
If you would recall correctly I suggested to get rid of the `hack/build.sh` cause its the same for every image and if you want to change one you will have to copy the change into all other 7 images. Thats not the case of the `contrib` dir where we have templates/helpers and where other scripts are(could) living or is it ?

> Keeping in our brains specific undocumented tricks for multiple repos we work on is counter-productive, unnecessary overhead.
What would you like to document about it ? If user understand Docker and knows what he can/can't do with the image then I don't think this will cause any overhead. If he wouldn't, then I don't think he would be a contributor.

jhadvig:
Anyway I see that this is leading nowhere so let other express their opinions
@bparees @mfojtik thought?

9.2/Dockerfile.rhel7
mfojtik: @bparees I think it is worth to move the RUN after we add the script. Also add comment line about why we are making it world writeable.
I think @rhcarvalho patch is correct, he just needs to add that chmod after RUN.

I don't think that adding extra ~1m (if we do that for all images) to a build job is something that will hurt. The benefit is to have much friendlier developer workflow ;-)

mfojtik:
@jhadvig i'm +1 to this change (but it has to go to all images). If the time of build will be longer for ~1m, I can live with that since we will get better build times when developing the images them-self in return.

mfojtik:
(also not a high priority, just nice to have if you're the image developer ;-)

bparees:
1) separate run statements (RUN, COPY, RUN) will build slower than (RUN, COPY) because a) each layer takes time to create in the docker build process and b) we have to squash more layers at the end.  Is it measurable or relevant?  maybe not, but just want to correct the technical point here.

2) chown is, as @rhcarvalho notes, pointless, since we don't care who owns the files, they need to be world writable to ensure any uid can access them, so chmod a+rwx is the requirement (and chmod o+rwx isn't good enough because if you do happen to be a member of the group that owns the file, you will end up with different/potentially lesser permissions)

3) I am ultimately fine with the argument from @mfojtik and @rhcarvalho that saving developer time in caching is more important than saving jenkins time by having fewer layers and forcing developers to temporarily reorder the statements while working on the scripts.

4) I strongly dislike the idea of setting the permissions in the github repo and relying on that to avoid a chmod, that is begging to cause problems down the road when someone introduces a new file and doesn't know to do that.  chmod is explicit and makes the dockerfile self-contained, so we should be doing that.

so in sum, if we want to go forward with this sort of change, i'm fine with it but we do need to move the chmod statements to a new RUN command after the COPY.

sound good?

rhcarvalho:
@bparees sounds good. Question, why do those particular files (the ones in 9.2/contrib, in this repo) need to be word-writable, and not just world-readable?

bparees:
Future proofing and a lack of doing a thorough investigation, mostly.
Depending what's in there someone might have a need/reason to exec in and
tweak them and I'd rather not block them from being able to.

Ben Parees | OpenShift

rhcarvalho:
@bparees, I have no problem with adding the extra RUN chmod line, but it's our common sense that nobody should be editing those files (nor possibly anything outside of a PV), as every tweaking is gone after a redeploy.

mfojtik:
@rhcarvalho they can override the files, correct?

```
FROM openshift/postgresql
ADD foo /bar
```

If the `/bar` is not world writeable, you will have to first change the USER to root and then change it back to the original user. Otherwise you will get permission denied.

rhcarvalho:
@mfojtik perhaps I did something wrong in this experiment... but:

```
[vagrant@openshiftdev docker-tmp]$ ls
Dockerfile  foo
[vagrant@openshiftdev docker-tmp]$ cat Dockerfile 
FROM centos
USER 26
COPY foo /bin/ls
[vagrant@openshiftdev docker-tmp]$ docker build -t copycopy .
Sending build context to Docker daemon  2.56 kB
Sending build context to Docker daemon 
Step 0 : FROM centos
latest: Pulling from docker.io/centos

f1b10cd84249: Already exists 
c852f6d61e65: Already exists 
7322fbe74aa5: Already exists 
Digest: sha256:90305c9112250c7e3746425477f1c4ef112b03b4abe78c612e092037bfecc3b7
Status: Downloaded newer image for docker.io/centos:latest
 ---> 7322fbe74aa5
Step 1 : USER 26
 ---> Running in fd7583e3bca6
 ---> bb71e1cb2cd4
Removing intermediate container fd7583e3bca6
Step 2 : COPY foo /bin/ls
 ---> d8f294669ae6
Removing intermediate container 61984b80ffba
Successfully built d8f294669ae6
[vagrant@openshiftdev docker-tmp]$ docker run --rm copycopy ls
exec: "ls": executable file not found in $PATH
FATA[0000] Error response from daemon: Cannot start container 5ac2373a445750e06d345d450594f9b68a68603946f50a969da9c1fb0edc6ef9: [8] System error: exec: "ls": executable file not found in $PATH 
```

Seems the file was copied, even though I was not root and user 26 cannot write to `/bin`?!

rhcarvalho:
`chmod -R a+rwx` both `/var/run/postgresql` and `/var/lib/pgsql`, added comments, and RUN - COPY - RUN.

@bparees @mfojtik PTAL

bparees:
@rhcarvalho COPY probably ignores permission issues.  what about "RUN touch /tmp/foo; cp foo /usr/bin/ls"  ?

that is where you'll have problems.

bparees:
changes lgtm but i see tests are failing...looks like  a network issue
[test]

9.2/Dockerfile.rhel7
```diff
@@ -44,7 +41,14 @@ RUN yum install -y yum-utils gettext && \
     localedef -f UTF-8 -i en_US en_US.UTF-8 && \
     mkdir -p /var/lib/pgsql/data && chown postgres.postgres /var/lib/pgsql/data && \
     test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)" && \
-    chmod -R go+rwx /var/lib/pgsql && chmod go+w /var/run/postgresql
+    # Loosen permission bits to avoid problems running container with arbitrary UID
```
bparees: i think this comment is messing up the dockerfile/RUN command.

bparees: oops, nm, was looking at the output from the jenkins test job...was running both in parallel.

bparees:
lgtm and passed test.

rhcarvalho:
> COPY probably ignores permission issues. what about "RUN touch /tmp/foo; cp foo /usr/bin/ls" ?

@bparees I think that deviates way too much from the reality we were discussing ;)
But yes, it fails:

> Step 3 : RUN touch /tmp/bar && cp /tmp/bar /usr/bin/pwd
 ---> Running in c90332e9956e
cp: cannot create regular file '/usr/bin/pwd': Permission denied

bparees:
I don't think it does deviate... The point in was originally trying to make
(maybe not clearly) was about what happens at runtime when someone execs
into the container and wants to edit things, or when the process itself
needs to touch those files. Which is basically the same as the RUN test.

Ben Parees | OpenShift

rhcarvalho:
I do agree that in the future we could add to `contrib/` files that would benefit from being editable by the current user in a running container. But that's not the case for the files currently there.

`common.sh` -- used by the entrypoint script
`*.template` -- not meant to be touched
`scl_enable` -- not meant to be touched

And files outside of a volume edited during a bash session in a running container are ephemeral... we don't promote that sort of usage, do we?

bparees:
We don't promote it but it's virtually inevitable and might be useful for a
user who is just trying to do some quick debugging and doesn't want to
rebuild the image to make changes (assuming the process in question
supported hot reload of config).

So yeah it's more a general future proofing and philosophy statement than
something that has an immediate impact on the existing images and contrib
files.

Ben Parees | OpenShift

PR #58, opened 2015-08-31
Update change password documentation

rhcarvalho:
Clarify documentation:
- The change password flow is valid not only when you have a persistent volume
- Passwords are always reset to match the env vars


rhcarvalho:
@mfojtik @bparees PTAL

jhadvig:
LGTM

PR #62, opened 2015-09-04
Setting groupID via env var

jhadvig:
@mfojtik PTAL
Also I looked into other images and I like the *_HOME (eg. JENKINS_HOME) vars that we are using. But I guess that just needed for Jenkins. Or ?


mfojtik:
@jhadvig sweet! thanks! let's wait for tests.

jhadvig:
@mfojtik @bparees need your blessing on this.

PR #63, opened 2015-09-11
move postgres 9.4 into rhscl namespace

bparees:


bparees:
@hhorak @praiskup @mfojtik fyi i'm aligning this with what was done for mongodb, are you ok with that?

bparees:
@hhorak @praiskup @mfojtik bump

hhorak:
@bparees I'm afraid it's not what we're going to do, the plan we have is centos/postgresql-94-centos7-candidate.

hhorak:
The mongodb will have to move as well at some point.

bparees:
@hhorak is the centos namespace available?  we need to put the postgres94 image somewhere.

bparees:
@hhorak note that you're making your life complicated by having a different namespace for the centos image and the rhel image.

hhorak:
RHSCL images based on centos might be for some time the only layered images available for centos and remember we always try to hide the fact those are based on RHSCL packages. Considering that I'd say we'd make more confusion if they would be available under rhscl/ on dockerhub.

PR #68, opened 2015-10-01
sync build-scripts with other images

praiskup:
- hack/ directory should be _just_ mirrored
- Makefile should be as declarative as possible
- build.sh should not rely on repository name


bparees:
[test]

PR #70, opened 2015-10-07
Improve the documentation on mounted directory permissions

strzibny:
I just came across this (even opened and closed an issue because of that) and feel like this should be mentioned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,13 @@ You can also set the following mount points by passing the `-v /host:/container`
 directory has the appropriate permissions and that the owner and group of the directory
 matches the user UID or name which is running inside the container.**
 
+Given that your mounted directory is `/data` run the following to set the right access rights:
+
+```
+# chmod a+rwx /data
+# chcon -t svirt_sandbox_file_t /data
+```
+
 Usage
 ----------------------
 


mnagy:
Hmm, not sure what to do with this one. Close?

bparees:
@mnagy I think it's a good thing to note though it should be clearer that it's referring to setting permissions in the host directory that you are bind mounting into the container for db storage. 

That said, I think we should let the scl team decide how to proceed with documenting it at this point. @praiskup.  The permissions need to at least ensure  the container uid can read/write the dir.

praiskup:
Well, are we talking about OpenShift environment, or regular 'docker run' use-case?

In the first case, OpenShift should take care of it, in the letter I would prefer something like
https://github.com/devexp-db/cont-postgresql/blob/master/share/cont-docs/postgresql/HOME.txt.tpl#L48
I bet I have a typo there however, there should be 'setfacl -m u:26:rwx' probably.

praiskup:
It is not a typo.

bparees:
@praiskup yes this is only relevant to manual docker run, all the more reason I defer to the SCL team.

omron93:
[test]

omron93:
[test]

omron93:
[test]

omron93:
[test]

omron93:
[test]

PR #72, opened 2015-10-19
Use explicit rhel7 version

hhorak:
Using `rhel7` only doesn't have to correspond with what we expect (latest rhel7), so we need to be more explicit here.


bparees:
[test]

PR #75, opened 2015-10-23
Use proper RHEL7 labels in v9.4

praiskup:
Also don't --disablerepo=epel because that command fails if there
is no epel repository configured (default).

### DIFF
--- a/9.4/Dockerfile.rhel7
+++ b/9.4/Dockerfile.rhel7
@@ -20,10 +20,11 @@ LABEL io.k8s.description="PostgreSQL is an advanced Object-Relational database m
       io.openshift.tags="database,postgresql,postgresql94,rh-postgresql94"
 
 # Labels consumed by Red Hat build service
-LABEL Component="rh-postgresql94" \
-      Name="openshift3/postgresql-94-rhel7" \
+LABEL Name="rhscl/postgresql-94-rhel7" \
+      BZComponent="rh-postgresql94-docker" \
       Version="9.4" \
-      Release="1"
+      Release="1" \
+      Architecture="x86_64"
 
 EXPOSE 5432
 


9.4/Dockerfile.rhel7
```diff
@@ -34,7 +35,7 @@ RUN yum install -y yum-utils gettext && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
     yum install -y --setopt=tsflags=nodocs bind-utils rh-postgresql94 && \
-    yum install -y --disablerepo="epel" --setopt=tsflags=nodocs nss_wrapper && \
```
mfojtik: @praiskup this is needed for the jenkins build... we have epel repo enabled on the host when we build and we don't want to install the nss_wrapper from epel (we want to install from official channels). This will prevent yum using EPEL.

mfojtik: @praiskup and it used not to fail :-)

praiskup: Weird:
```
$ sudo yum install something-that-does-not-exist --disablerepo epel
Error: Unknown repo: 'epel'
$ echo $?
1
```

praiskup:
ptal

mfojtik:
LGTM @hhorak we need to do this for all image, correct?

9.4/Dockerfile.rhel7
praiskup: @mfojtik wrote:
> @praiskup this is needed for the jenkins build... we have epel repo enabled
> on the host

What are the reasons for this anyway?

hhorak:
@mfojtik yes, I hoped this should already be fixed in most of them, but found it's not still for perl, python  and ruby.. (the new versions)
@praiskup could I ask you to use your scripting skills and submit a PRs for that?

hhorak:
[test]

praiskup:
FYI
https://raw.githubusercontent.com/praiskup/temporary-helpers/fcde79895785be9a87eb306aac308d6399652d61/rh-os-label-generator

hhorak:
@praiskup thx, it seems helpful :)

mfojtik:
@hhorak @praiskup lets hope we will not need to change labels too often ;-)

praiskup:
As far as we do not need to keep "Release" label in sync with RHEL's dist-git ..

hhorak:
I guess Release label may stay as downstream-only difference.

PR #77, opened 2015-11-05
sync container principles with mysql

praiskup:
@hhorak FYI


hhorak:
[test]

hhorak:
Script `scl_enable` can use `ENABLED_COLLECTIONS` env. variable, so we can share the scripts with 9.2.

hhorak:
@mfojtik I see the replica example uses `postgres-master` and `postgres-slave` as image arguments -- is it necessary to include symlinks `/usr/bin/postgres-master -> run-postgresql-master` and  `/usr/bin/postgres-slave -> run-postgresql-slave` to maintain compatibility for now or is it enough to change the replica example?

mfojtik:
@hhorak it is OK to change the replica example to reflect the new names

mfojtik:
[test]

9.4/root/usr/bin/container-entrypoint
```diff
@@ -0,0 +1,3 @@
+#! /bin/bash
```
mfojtik: the white space here is intentional?

9.4/root/usr/bin/run-postgresql
```diff
@@ -2,8 +2,9 @@
 
 set -eu
 
-source ${HOME}/common.sh
+source "$CONTAINER_SCRIPTS_PATH"/common.sh
```
mfojtik: i'm not bash expert but shouldn't this be `source "${CONTAINER_SCRIPTS_PATH}/common.sh"` ?

9.4/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -137,3 +139,18 @@ function set_passwords() {
 
   pg_ctl stop
 }
+
+# NOTE: https://github.com/openshift/postgresql/issues/76
```
mfojtik: can you please replace the note with some description? you can still link the issue, but would be great to have something that does not force us to open the browser ;-)

mfojtik:
small nits, but LGTM

@bparees || @rhcarvalho  you want to have second look?

9.4/root/usr/bin/container-entrypoint
praiskup: Yes (according to my habits from autotools).  Shoul I remove it?

9.4/root/usr/bin/run-postgresql
praiskup: Those two are equivalent.  Again just a habit, I'm fine to fix.

mfojtik: https://google.github.io/styleguide/shell.xml?showone=Quoting#Quoting

I'm usually follow this :-) Maybe it is wrong, but they say:

```
# "quote variables"
echo "${flag}"
```

9.4/root/usr/bin/container-entrypoint
mfojtik: for sake of consistency with other script I would prefer to remove it, unless you have good reason to keep it, but then you're committing yourself to fix it everywhere :-P

hhorak: @mfojtik I appreciate your focus on detail, but I don't think keeping the shebang format same is something we need to keep same across images, in case it works the same.

mfojtik: @hhorak just want to avoid future questions about that :-) i'm OK with merging this as it is.

hhorak:
> @hhorak it is OK to change the replica example to reflect the new names

@praiskup please, don't forget to change the "arg" in the replica example, I've learned in the mysql example that it may do some troubles..

Otherwise it looks fine to me, LGTM.

9.4/root/usr/bin/run-postgresql
praiskup: If I read correctly, its about *quoting* strings containing *variables*, which I actually done because the original code suffered from it.  I'll update according to your style .. a sec please.

praiskup:
ping

mfojtik:
@praiskup can you squash please? @bparees merge this at your will :-)

hhorak:
[test]

mfojtik:
@bparees i'm merging this to unblock @hhorak folks, this should not break anything (hopefully)

PR #78, opened 2015-11-06
Added fix for permissions, owner to pgsql directory

### LINKED ISSUE #64: Permissions 0777 set on some directories within Dockerfile
As discussed in openshift/mysql/pull/94 and mysql/pull/92#issuecomment-139270899.

Fwiw "Reorder Dockerfile for better use of cache" commit message did not help me
to understand the reasoning for the `chmod a+rwx ..` change...

Akasurde:
Fixes #64 
Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

### DIFF
--- a/9.2/Dockerfile
+++ b/9.2/Dockerfile
@@ -32,13 +32,13 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     yum -y --setopt=tsflags=nodocs install nss_wrapper && \
     yum clean all && \
     localedef -f UTF-8 -i en_US en_US.UTF-8 && \
-    mkdir -p /var/lib/pgsql/data && chown postgres.postgres /var/lib/pgsql/data && \
-    test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)" && \
-    # Loosen permission bits to avoid problems running container with arbitrary UID
-    chmod -R a+rwx /var/run/postgresql
+    mkdir -p /var/lib/pgsql/data && \
+    test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)"
 
 # Loosen permission bits to avoid problems running container with arbitrary UID
-RUN chmod -R a+rwx /var/lib/pgsql
+ADD root /
+RUN /usr/libexec/fix-permissions /var/lib/pgsql && \
+    /usr/libexec/fix-permissions /var/run/postgresql
 
 # Get prefix path and path to scripts rather than hard-code them in scripts
 ENV CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/postgresql \
@@ -51,7 +51,6 @@ ENV BASH_ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
     ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
     PROMPT_COMMAND=". ${CONTAINER_SCRIPTS_PATH}/scl_enable"
 
-ADD root /
 
 VOLUME ["/var/lib/pgsql/data"]
 
--- a/9.2/root/usr/libexec/fix-permissions
+++ b/9.2/root/usr/libexec/fix-permissions
@@ -1,0 +1,7 @@
+#!/bin/sh
+# Fix permissions on the given directory to allow group read/write of
+# regular files and execute of directories.
+chown -R postgres "$1"
+chgrp -R 0 "$1"
+chmod -R g+rw "$1"
+find "$1" -type d -exec chmod g+x {} +
--- a/9.4/root/usr/libexec/fix-permissions
+++ b/9.4/root/usr/libexec/fix-permissions
@@ -1,0 +1,7 @@
+#!/bin/sh
+# Fix permissions on the given directory to allow group read/write of
+# regular files and execute of directories.
+chown -R postgres "$1"
+chgrp -R 0 "$1"
+chmod -R g+rw "$1"
+find "$1" -type d -exec chmod g+x {} +
--- a/9.2/Dockerfile.rhel7
+++ b/9.2/Dockerfile.rhel7
@@ -37,13 +37,13 @@ RUN yum install -y yum-utils gettext && \
     yum install -y --disablerepo="epel" --setopt=tsflags=nodocs nss_wrapper && \
     yum clean all && \
     localedef -f UTF-8 -i en_US en_US.UTF-8 && \
-    mkdir -p /var/lib/pgsql/data && chown postgres.postgres /var/lib/pgsql/data && \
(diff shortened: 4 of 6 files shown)


mfojtik:
@praiskup review pls

mfojtik:
[test]

praiskup:
I dislike the 'a+rwx' a lot -- this sounds like step forward, but reasoning
why this change is needed and why exactly `postgres.0`, etc.. is needed.

mfojtik:
@praiskup the chgroup 0 is needed for this image to run in OpenShift IMHO (the user is random, but the group is always 0)

mfojtik:
@praiskup in other words, restricting write to the root group is better than having it world writable ;-)

praiskup:
I see, but mysql has e.g. in-code comment:
https://github.com/openshift/mysql/blob/master/5.6/root/usr/libexec/container-setup#L53

9.2/Dockerfile
```diff
@@ -32,13 +32,13 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     yum -y --setopt=tsflags=nodocs install nss_wrapper && \
     yum clean all && \
     localedef -f UTF-8 -i en_US en_US.UTF-8 && \
-    mkdir -p /var/lib/pgsql/data && chown postgres.postgres /var/lib/pgsql/data && \
+    mkdir -p /var/lib/pgsql/data && chown -R postgres.0 /var/lib/pgsql && \
     test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)" && \
     # Loosen permission bits to avoid problems running container with arbitrary UID
     chmod -R a+rwx /var/run/postgresql
```
bparees: why not change this one too?

bparees:
two main comments:
1) please update the Dockerfile.rhel7 files too
2) we now have a fix_permissions script we use for stuff like this, it would be better to use it so we can have some consistency in how permissions get set: 
https://github.com/openshift/sti-base/blob/master/bin/fix-permissions
(you'll have to copy that script into this repo since this image isn't based on the sti-base image)

once that script is in the image, you can invoke it with the directory to "fix" as the parameter.

Akasurde:
OK, I will do this and revert back.

Akasurde:
Ping

bparees:
lgtm, but can you please squash your commits?  thanks.
[test]

praiskup:
@bparees wrote:
> lgtm, but can you please squash your commits? thanks.

FWIW:  Does not look good to me.

We should probably move the `fix-permissions` script at least into
`root/usr/libexec`, or if it will stay in $PATH (and not removed before the
image is squashed) could we name it like `artificial-uid-perm-fix`?

Akasurde:
@praiskup We can move `fix-permissions` to `/usr/libexec`  directory and run it from there.

praiskup:
+1 if you think about new `libexec` subdir in https://github.com/openshift/postgresql/tree/master/9.4/root/usr

praiskup:
Is this fine to [test]?  I would quote variables in fix-permissions script, but not
a problem with actual hard-wired calls from Dockerfiles.

bparees:
[test]

bparees:
`failed: mkdir: cannot create directory '/var/lib/pgsql/data/userdata': Permission denied`

ironically when you actually run as the "postgres" user, you are not a member of the "root" group because uid 26 can actually be found in /etc/passwd with a group of postgres.

the easiest fix is probably to set the user ownership to postgres, in fix-permissions.  That way the files will be writable by either the postgres user, or anyone else (because anyone else will be a member of the root group)

Akasurde:
@bparees 

Adding following line in `fix-permissions` fixes issue

`chown -R postgres $1`

Let me know if this is OK?

bparees:
[test]

bparees:
I don't see the change, did you push it?

praiskup:
@Akasurde Note the missing quotation around $1.

bparees:
@mnagy is this PR necessary to fix replication, or is this just clean up?  I don't want to merge anything we don't need.

9.2/root/usr/libexec/fix-permissions
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+# Fix permissions on the given directory to allow group read/write of
+# regular files and execute of directories.
+chown -R postgres "$1"
+chgrp -R 0 "$1"
+chmod -R g+rw "$1"
+find $1 -type d -exec chmod g+x {} +
```
mnagy: Quotes around $1

mnagy: please add

9.4/root/usr/libexec/fix-permissions
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+# Fix permissions on the given directory to allow group read/write of
+# regular files and execute of directories.
+chown -R postgres "$1"
+chgrp -R 0 "$1"
+chmod -R g+rw "$1"
+find $1 -type d -exec chmod g+x {} +
```
mnagy: same here

mnagy:
Shouldn't we call the fix-permission script from the entrypoint as well?

Akasurde:
@mnagy  Do we need to call `fix-permissions` entrypoint as well ?

praiskup:
I don't think so.  You even do not have usually permissions to change ownership of file.

bparees:
not needed for replication fixing, deferring this until we get replication straightened out.

Akasurde:
@bparees ping

mnagy:
I'd like to note that it would be a very good idea to test this PR with OS extended tests we now have for replication.

bparees:
[test]

9.4/root/usr/libexec/fix-permissions
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+# Fix permissions on the given directory to allow group read/write of
+# regular files and execute of directories.
+chown -R postgres "$1"
+chgrp -R 0 "$1"
+chmod -R g+rw "$1"
+find "$1" -type d -exec chmod g+x {} +
```
mnagy: Are these commands allowed to fail? If not, we should add `set -e`. We should also make sure $1 is defined. Together with what we had previously, `set -eu` is ideal.

bparees:
@mnagy is testing this now for replication.  going forward we're going to ask that the SCL team start running the openshift extended image tests(which include replication) on these PRs (the PR tester does a more basic test, it doesn't test replication scenarios for example).

we can discuss how best to approach this on tuesday.

mnagy:
FYI, verified these changes and tests passed.

9.4/root/usr/libexec/fix-permissions
bparees: it's not in our sti-base script either:
https://github.com/openshift/sti-base/blob/master/bin/fix-permissions

updating this script here and elsewhere can be a follow up.

PR #80, opened 2015-11-12
Fix paths to config templates

mnagy:
@praiskup @bparees fyi

### DIFF
--- a/9.2/root/usr/share/container-scripts/postgresql/common.sh
+++ b/9.2/root/usr/share/container-scripts/postgresql/common.sh
@@ -80,6 +80,12 @@ function generate_postgresql_config() {
       > "${POSTGRESQL_CONFIG_FILE}"
 }
 
+function generate_postgresql_recovery_config() {
+  envsubst \
+      < "${CONTAINER_SCRIPTS_PATH}/openshift-custom-recovery.conf.template" \
+      > "${POSTGRESQL_RECOVERY_FILE}"
+}
+
 # Generate passwd file based on current uid
 function generate_passwd_file() {
   export USER_ID=$(id -u)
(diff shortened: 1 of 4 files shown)


praiskup:
Yeah, one issue I see here that we generate configuration file which
is not by default on persistent storage.  Would this be possible to move into
/var/lib/pgsql/data/userdata?

praiskup:
- s/$CONTAINER_SCRIPTS_PATH/$PGDATA/
- fix envsubst

mnagy:
We do not want that. We want the configuration file to be regenerated every time the container is started, putting it in persistent volume doesn't make much sense. That takes care of changing environment variable changes, image updates and so on.

mfojtik:
@mnagy we should allow volume with user specified configuration file and in that case, our entrypoint will just skip all env variables and leave the user with his pain :-)

mnagy:
@mfojtik that sounds reasonable. Not in this PR though :)

praiskup:
I still don't see a reason why we have to have another fix-permissions'ed variable
user potentially have to care about?

mnagy:
@praiskup I don't understand. What do you mean by fix-permisions'ed variable?

praiskup:
s/variable/directory/

mnagy:
I still don't follow. Can you elaborate?

praiskup:
To pass the testsuite, you need to allow group 0 to write into
/usr/share/container-scripts/postgresql
-- so (a) /var/lib/pgsql, (b) /var/run/postgresql and now (c).

mnagy:
PTAL

praiskup:
Looks fine.

mfojtik:
@mnagy is there any test case for this?

mnagy:
@mfojtik I've tested it manually. There is a PR for extended tests here: https://github.com/openshift/origin/pull/5874

9.2/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -80,6 +80,12 @@ function generate_postgresql_config() {
       > "${POSTGRESQL_CONFIG_FILE}"
 }
 
+function generate_postgresql_recovery_config() {
+  envsubst \
+      < "${CONTAINER_SCRIPTS_PATH}/openshift-custom-recovery.conf.template" \
```
bparees: i don't see this file, where is it coming from?

PR #81, opened 2015-11-12
Add function waiting for PostgreSQL master to start

mnagy:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1263671

### DIFF
--- a/9.2/root/usr/bin/run-postgresql-slave
+++ b/9.2/root/usr/bin/run-postgresql-slave
@@ -9,7 +9,6 @@ set_pgdata
 function initialize_replica() {
   echo "Initializing PostgreSQL slave ..."
   chmod 0700 $PGDATA
-  export MASTER_FQDN=$(postgresql_master_addr)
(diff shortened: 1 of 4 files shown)


bparees:
how is restarting it periodically, only to have it hit the timeout and fail again, better than a single run that waits forever?  

It will eventually start if the master appears, right?  So i see no benefit to timing out only to restart.  But it should print something to the logs to indicate what it is waiting for.

mnagy:
@bparees I agree. And it prints out every second that it waits.

mnagy:
..including the master FQDN

mfojtik:
@bparees what about the case when the master never starts... you want that container run forever in loop? When the container restarts for a long time, it is a good indication that something is wrong (eg. master did not started properly).

bparees:
> When the container restarts for a long time, it is a good indication that something is wrong (eg. master did not started properly)

that is what readiness checks are for, so if we're going to do that, it should be a readiness check, not baked into the run script.

bparees:
or liveness rather.

bparees:
does this not affect 9.4?

mnagy:
@bparees yes, it does. I forgot to apply those changes there as well, good catch.

mnagy:
Fixed, PTAL

PR #82, opened 2015-11-12
Remove slave data before starting

mnagy:
This is required for us, because when an OpenShift pod gets restarted, sometimes
it gets the data volume, with old data. It would be nice to not delete it every
time, but no time to get fancy right now.


bparees:
[test]

bparees:
[test]

PR #83, opened 2015-11-12
Add a PostgreSQL version parameter to the replica template

mnagy:
Since we only have one version of the PostgreSQL template and I want to be able to test both our versions in my tests, I decided to introduce a POSTGRESQL_VERSION template parameter to be able to specify it. I think it is better in that it saves us the duplication. I think it is worse in that we do not use such a parameter anywhere else. And we do not have any parameter validation, so if user puts in something else than 92 or 94, he will have weird errors.

Maybe we should have the whole image as a parameter? Comments welcome.

### DIFF
--- a/examples/replica/postgresql_replica.json
+++ b/examples/replica/postgresql_replica.json
@@ -62,6 +62,12 @@
       "required": true
     },
     {
+      "name": "POSTGRESQL_IMAGE",
+      "description": "PostgreSQL Docker image to use",
+      "value": "openshift/postgresql-92-centos7",
+      "required": true
+    },
+    {
       "name": "VOLUME_CAPACITY",
       "description": "Volume space available for data, e.g. 512Mi, 2Gi",
       "value": "512Mi",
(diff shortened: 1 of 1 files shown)


mnagy:
@bparees @mfojtik @rhcarvalho I'd like to know your thoughts

bparees:
We actually do already to this elsewhere...make it an image name parameter instead, so you can also use it for the rhel images:
https://github.com/openshift/origin/blob/master/examples/jenkins/jenkins-ephemeral-template.json#L136-L139
https://github.com/openshift/origin/blob/master/examples/jenkins/jenkins-ephemeral-template.json#L91

Note that i have pretty mixed feelings about using the "image" field on a DC instead of an imagechangetrigger.  By doing this, your deployment will not get updated when new images become available.  But since we're already doing it elsewhere (and were doing it here even w/o the parameter), we can press on for now.

rhcarvalho:
+1 for specifying the whole image as a parameter. Apart from the advantage @bparees mentioned, you could also theoretically use the template with images derived from our images, as long as the derive image stays compatible.

mnagy:
PTAL.

Also, @bparees mind opening issues for the DC issue you mentioned? No reason we can't change that later.

mfojtik:
LGTM

bparees:
@mnagy on further reflection we probably don't want to setup an imagechange trigger for the DBs because a redeploy of a DB pod using emptydir storage will result in data loss.

PR #89, opened 2015-12-18
always use $HOME/data/userdata

bparees:
fixes https://github.com/openshift/postgresql/issues/76

### DIFF
--- a/9.2/root/usr/share/container-scripts/postgresql/common.sh
+++ b/9.2/root/usr/share/container-scripts/postgresql/common.sh
@@ -164,19 +164,12 @@ function set_passwords() {
 
 function set_pgdata ()
 {
-  # TODO:  Remove this.
-  # Container should not choose different PGDATA location based on the way how
-  # data directory is "mounted";  mount permissions might change among different
-  # 'docker run' invocations (openshift/postgresql/issues/76).
-
-  if [ -O "${HOME}/data" ]; then
+  # backwards compatibility case, we used to put the data here
+  if [ -e ${HOME}/data/PG_VERSION ]; then
     export PGDATA=$HOME/data
-  else
-    # If current user does not own data directory
-    # create a subdirectory that the user does own
-    if [ ! -d "${HOME}/data/userdata" ]; then
-      mkdir "${HOME}/data/userdata"
-    fi
+  else 
+    # create a subdirectory that the user owns
+    mkdir -p "${HOME}/data/userdata"
     export PGDATA=$HOME/data/userdata
   fi
 }
--- a/9.4/root/usr/share/container-scripts/postgresql/common.sh
+++ b/9.4/root/usr/share/container-scripts/postgresql/common.sh
@@ -164,19 +164,12 @@ function set_passwords() {
 
 function set_pgdata ()
 {
-  # TODO:  Remove this.
-  # Container should not choose different PGDATA location based on the way how
-  # data directory is "mounted";  mount permissions might change among different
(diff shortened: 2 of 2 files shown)


bparees:
@praiskup @rhcarvalho @mnagy ptal.  logically this seems safe, but maybe i'm missing something?

9.2/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -164,21 +164,12 @@ function set_passwords() {
...
-    export PGDATA=$HOME/data
-  else
-    # If current user does not own data directory
-    # create a subdirectory that the user does own
-    if [ ! -d "${HOME}/data/userdata" ]; then
-      mkdir "${HOME}/data/userdata"
-    fi
-    export PGDATA=$HOME/data/userdata
+  # If current user does not own data directory
+  # create a subdirectory that the user does own
```
rhcarvalho: This comment doesn't make sense.

9.2/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -164,21 +164,12 @@ function set_passwords() {
...
+  if [ ! -d "${HOME}/data/userdata" ]; then
+    mkdir "${HOME}/data/userdata"
```
rhcarvalho: If you do `mkdir -p` you don't need to test for existence.

rhcarvalho:
The existing scheme of using one dir or another is already problematic, we could in consecutive runs end up with data in `/data` and next in `/data/userdata`.
Now, why do we make up the dir name "userdata"?

rhcarvalho:
The only potential problem I see with this change is breaking people that fell on the first branch of the conditional and have data stored under `/data`.

bparees:
@rhcarvalho good point about compatibility, i've checked in an attempt to include backwards compatibility in the logic and fixed your other comments, ptal.

rhcarvalho:
I'm not sure about the `${HOME}/data/PG_VERSION`. Assuming you tested that, LGTM.

BTW -- what I've run into in the past with misconfigured permissions is that `data/userdata` might exist and NOT be owned by the user...

As an improvement, we might want to test that the directory we are setting up as PG_DATA is writable by the user, and if not provide an error message instead of failing later with cryptic messages.

bparees:
> I'm not sure about the ${HOME}/data/PG_VERSION. Assuming you tested that, LGTM.

i started up an image and confirmed that file gets created in that dir, that was the basis for the logic.

praiskup:
Looks OK, though I would probably rather return non-zero value (or exit
immediately with "fatal" error) after 'mkdir -p' failure.

praiskup:
@rhcarvalho wrote
> Now, why do we make up the dir name "userdata"?

That is good question, btw.

mfojtik:
i thought we are not using $HOME but absolute path as $HOME is something that might change and depend on the user environment.

praiskup:
In case of 'postgres' user, it is created by RPM and $HOME directory should
be properly pre-defined (and I would really discourage to change the default).
From this POV it looks fine, but I do not object if you plan to change this
to absolute path.

bparees:
@praiskup 
> Looks OK, though I would probably rather return non-zero value (or exit
immediately with "fatal" error) after 'mkdir -p' failure.

the start scripts that invoke this run with -e, so this should exit on error.

@rhcarvalho 
> Now, why do we make up the dir name "userdata"?

because it's the data dir and its owned by the user.

PR #92, opened 2016-01-12
Verify all installed packages using rpm -V

mfojtik:


praiskup:
I'm just curious what was the trigger to this change.  I doubt this helps
much because e.g. if you install 'postgresql92' not only that package gets
installed (and I don't think users are ever  going to change something
within postgresql92 meta-package).

mfojtik:
@praiskup this was because `yum install <list>` silently skip packages that are not installed. This caused nss_wrapper package not installed in RHEL7 images without us knowing (we don't inspect the build logs). I added this to postgresql to keep this consistent with other repositories. This will be replaced with `yum install --strict` once that will land in RHEL7.X world and centos.

hhorak:
I think we should merge this, it can catch potential issues sooner.

PR #94, opened 2016-02-11
move data from old pgdata dir to new location

bparees:
after discussion with @pmorie we think it would be safer to move user data to the common user-owned directory rather than keep using the mounted path, because fsGroup changes might end up breaking the permissions on that directory.

### DIFF
--- a/9.2/root/usr/share/container-scripts/postgresql/common.sh
+++ b/9.2/root/usr/share/container-scripts/postgresql/common.sh
@@ -10,6 +10,11 @@ postinitdb_actions=
 psql_identifier_regex='^[a-zA-Z_][a-zA-Z0-9_]*$'
 psql_password_regex='^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$'
 
+# match . files when moving userdata below
+shopt -s dotglob
+# extglob enables the !(userdata) glob pattern below.
+shopt -s extglob
+
 function usage() {
   if [ $# == 1 ]; then
     echo >&2 "error: $1"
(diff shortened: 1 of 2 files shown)


bparees:
@rhcarvalho ptal

@pmorie fyi

@hhorak i think we're passed the agreed upon freeze date for the new SCL images, however it would be really great if we could include this change in the refresh of the SCL postgres9.4 image, as otherwise we're going to have some compatibility issues.  Is it possible?

hhorak:
I think it should be doable. Hopefully this change won't break anything.

bparees:
> I think it should be doable. Hopefully this change won't break anything.

yeah..you and me both.  But this path should only be taken in the case where someone has an existing persistent volume that has data on it in the "/data" path (not the "/data/userdata" path) which should be exceedingly rare.

bparees:
[test]

bparees:
[test]

praiskup:
+1 for moving one level down below mount point

rhcarvalho:
@bparees any chance the `mv` can fail because of file permissions on the parent dir?

rhcarvalho:
Assuming that `shopt` + `!(userdata)` + moving all kinds of files was tested, LGTM, and +1 to using the same directory in all cases.

mnagy:
[test][openshift]

mnagy:
LGTM, but let's wait for tests.

bparees:
@rhcarvalho if mv would fail due to permissions, creating the directory would also fail, as would postgres trying to use the directory, so i think it's safe.

PR #98, opened 2016-02-15
add tar and rsync to image

bparees:


mnagy:
LGTM

praiskup:
Changes like this are asking for reasoning in commit message.
Consider you want to make the image smaller but you are worried
about possible consequences..  and `git blame` does not help.

I'm asking for better commit messages, PR discussion is decoupled
from the change itself and thus might be considered non-persistent.

bparees:
@praiskup good point, sorry about that.  In this particular case we added rsync and tar because they are needed to support the openshift "rsync" capability for copying files into a container:
https://docs.openshift.org/latest/dev_guide/copy_files_to_container.html

That said, git blame isn't likely to help because we're going to update that line repeatedly.  But certainly the commit history could help.

It might make even more sense to just have a comment in the dockerfile that lists each package we install and why.

PR #100, opened 2016-02-23
Ensuring PGDATA sane permissions

caruccio:
Looks like [this](https://github.com/kubernetes/kubernetes/blob/dff7490c57f309e8ba29563a35d9e2147f241f9a/pkg/volume/volume_linux.go#L74) is something postgres [doesn't like much](https://github.com/postgres/postgres/blob/REL9_4_STABLE/src/backend/postmaster/postmaster.c#L1462-L1467).

### DIFF
--- a/9.2/root/usr/share/container-scripts/postgresql/common.sh
+++ b/9.2/root/usr/share/container-scripts/postgresql/common.sh
@@ -183,6 +183,8 @@ function set_pgdata ()
     mkdir -p "${HOME}/data/userdata"
   fi
   export PGDATA=$HOME/data/userdata
+  # ensure sane perms for postgresql startup
+  chmod 700 "$PGDATA"
 }
 
 function wait_for_postgresql_master() {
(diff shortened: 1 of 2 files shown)


spinolacastro:
Just adding a few more context:
When the pvc bounds to a ceph pv, the plugin ensures the permissions, so the mount point starts with (0660) mode.
As postgres expects 0700 at pgdata, it refuses to start (if we stop/start the pod).

mfojtik:
[test]

mfojtik:
@praiskup PTAL

praiskup:
I don't think the "ensure ..." comment is correct, as there is `|| true` (so no guarantees).  Is that what
we want?  I would rather fail sooner than later.

caruccio:
@praiskup exactly! That `|| false` shouldn't be there

caruccio:
Fixed and squashed to a single commit.

praiskup:
Looks OK.

PR #104, opened 2016-03-18
Tune shared_buffers and effective_cache_size based on given memory

mfojtik:
@praiskup @bparees initial database tuning PR.

We already chat about this with @praiskup and his concerns are that we should not be changing the defaults and let users to provide the 'correct' value. 
My thoughts are that we as a platform can deterministically get the memory (based on the resource quota set for the Pod) so we can provide better "default" based on upstream formulas. If we will not do it, users will have to change them every time they  adjust them in OpenShift. The assumption here is that they will follow the upstream formulas and so we can make their life easier by doing that for them. 

Note that this auto-tuning **only** take account when you run the container with `--memory=512MB` or so. By default we don't do anything. Also note that users can still override the values manually (the user  provided env vars take precedence)

@hhorak ptal as well, we might do something similar to mysql :-)

For the reference, there is Trello card for this: https://trello.com/c/cmY00byV/675-8-update-images-to-autoconfigure-based-on-available-memory-template-evg

### DIFF
--- a/9.2/root/usr/bin/cgroup-limits
+++ b/9.2/root/usr/bin/cgroup-limits
@@ -1,0 +1,92 @@
+#!/usr/bin/python
+
+"""
+Script for parsing cgroup information
+
+This script will read some limits from the cgroup system and parse
+them, printing out "VARIABLE=VALUE" on each line for every limit that is
+successfully read. Output of this script can be directly fed into
+bash's export command. Recommended usage from a bash script:
+
+    set -o errexit
+    export_vars=$(cgroup-limits) ; export $export_vars
+
+Variables currently supported:
+    MAX_MEMORY_LIMIT_IN_BYTES
+        Maximum possible limit MEMORY_LIMIT_IN_BYTES can have. This is
+        currently constant value of 9223372036854775807.
+    MEMORY_LIMIT_IN_BYTES
+        Maximum amount of user memory in bytes. If this value is set
+        to the same value as MAX_MEMORY_LIMIT_IN_BYTES, it means that
+        there is no limit set. The value is taken from
+        /sys/fs/cgroup/memory/memory.limit_in_bytes
+    NUMBER_OF_CORES
+        Number of detected CPU cores that can be used. This value is
+        calculated from /sys/fs/cgroup/cpuset/cpuset.cpus
+    NO_MEMORY_LIMIT
+        Set to "true" if MEMORY_LIMIT_IN_BYTES is so high that the caller
+        can act as if no memory limit was set. Undefined otherwise.
+"""
+
+from __future__ import print_function
+import sys
+
+
+def _read_file(path):
+    try:
+        with open(path, 'r') as f:
+            return f.read().strip()
+    except IOError:
+        return None
+
+
+def get_memory_limit():
+    """
+    Read memory limit, in bytes.
+    """
+
+    limit = _read_file('/sys/fs/cgroup/memory/memory.limit_in_bytes')
+    if limit is None or not limit.isdigit():
+        print("Warning: Can't detect memory limit from cgroups",
+              file=sys.stderr)
+        return None
+    return int(limit)
+
+
+def get_number_of_cores():
+    """
+    Read number of CPU cores.
+    """
+
+    core_count = 0
+
+    line = _read_file('/sys/fs/cgroup/cpuset/cpuset.cpus')
+    if line is None:
+        print("Warning: Can't detect number of CPU cores from cgroups",
+              file=sys.stderr)
+        return None
+
+    for group in line.split(','):
+        core_ids = list(map(int, group.split('-')))
+        if len(core_ids) == 2:
+            core_count += core_ids[1] - core_ids[0] + 1
+        else:
+            core_count += 1
+
+    return core_count
+
+
+if __name__ == "__main__":
+    env_vars = {
+        "MAX_MEMORY_LIMIT_IN_BYTES": 9223372036854775807,
+        "MEMORY_LIMIT_IN_BYTES": get_memory_limit(),
+        "NUMBER_OF_CORES": get_number_of_cores()
+    }
+
+    env_vars = {k: v for k, v in env_vars.items() if v is not None}
+
+    if env_vars.get("MEMORY_LIMIT_IN_BYTES", 0) >= 92233720368547:
+        env_vars["NO_MEMORY_LIMIT"] = "true"
+
+    for key, value in env_vars.items():
+        print("{0}={1}".format(key, value))
--- a/9.2/root/usr/share/container-scripts/postgresql/common.sh
+++ b/9.2/root/usr/share/container-scripts/postgresql/common.sh
@@ -1,6 +1,22 @@
 # Configuration settings.
 export POSTGRESQL_MAX_CONNECTIONS=${POSTGRESQL_MAX_CONNECTIONS:-100}
-export POSTGRESQL_SHARED_BUFFERS=${POSTGRESQL_SHARED_BUFFERS:-32MB}
+
+# Perform auto-tuning based on the container cgroups limits (only when the
+# limits are set).
+# Users can still override this by setting the POSTGRESQL_SHARED_BUFFERS
+# and POSTGRESQL_EFFECTIVE_CACHE_SIZE variables.
+if [[ "${NO_MEMORY_LIMIT}" == "true" ]]; then
+    export POSTGRESQL_SHARED_BUFFERS=${POSTGRESQL_SHARED_BUFFERS:-32MB}
+    export POSTGRESQL_EFFECTIVE_CACHE_SIZE=${POSTGRESQL_EFFECTIVE_CACHE_SIZE:-128MB}
+else
+    # Use 1/4 of given memory for shared buffers
+    shared_buffers_computed="$(($MEMORY_LIMIT_IN_BYTES/1024/1024/4))M"
+    # Setting effective_cache_size to 1/2 of total memory would be a normal conservative setting,
+    effective_cache="$(($MEMORY_LIMIT_IN_BYTES/1024/1024/2))M"
+    export POSTGRESQL_SHARED_BUFFERS=${POSTGRESQL_SHARED_BUFFERS:-$shared_buffers_computed}
+    export POSTGRESQL_EFFECTIVE_CACHE_SIZE=${POSTGRESQL_EFFECTIVE_CACHE_SIZE:-$effective_cache}
+
+fi
 
 export POSTGRESQL_RECOVERY_FILE=$HOME/openshift-custom-recovery.conf
 export POSTGRESQL_CONFIG_FILE=$HOME/openshift-custom-postgresql.conf
--- a/9.4/README.md
+++ b/9.4/README.md
@@ -24,6 +24,7 @@ The following environment variables influence the PostgreSQL configuration file.
 | :---------------------------- | ----------------------------------------------------------------------- | -------------------------------
 |  `POSTGRESQL_MAX_CONNECTIONS` | The maximum number of client connections allowed. This also sets the maximum number of prepared transactions. |  100
 |  `POSTGRESQL_SHARED_BUFFERS`  | Sets how much memory is dedicated to PostgreSQL to use for caching data |  32M
+|  `POSTGRESQL_EFFECTIVE_CACHE_SIZE`  | Set to an estimate of how much memory is available for disk caching by the operating system and within the database itself |  128M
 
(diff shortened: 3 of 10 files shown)


praiskup:
This is not a bad idea, but it is hard to decide what should be a 'correct' value.
As discussed over irc, https://github.com/gregs1104/pgtune might be interesting
project to cooperate with.

mfojtik:
@praiskup agree, using that script will make it better, but it has to support cgroups ;-) (it does not from the source code). I can add that as trello card/follow up.

hhorak:
Generally, I think we can do some reasonable auto-tuning, if it is properly documented, users understand what is going on and they have way to change it. I see this change is documented in README.md, but this file is not part of the resulting image.
Well, I don't say it must be implemented as part of this PR, but we should start using some standard way, once it is set up (https://github.com/projectatomic/container-best-practices/issues/38) and document things from README.md in the image as well.

9.2/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -1,6 +1,21 @@
...
+# limits are set).
+# Users can still override this by setting the POSTGRESQL_SHARED_BUFFERS
+# variable.
+if [[ "${MAX_MEMORY_LIMIT_IN_BYTES}" != "${MEMORY_LIMIT_IN_BYTES}" ]]; then
+    # Use 1/4 of given memory for shared buffers
+    shared_buffers_computed="$(python -c "print(int((${MEMORY_LIMIT_IN_BYTES}/(1024*1024))*0.25))")M"
+    # Setting effective_cache_size to 1/2 of total memory would be a normal conservative setting,
+    effective_cache="$(python -c "print(int((${MEMORY_LIMIT_IN_BYTES}/(1024*1024))*0.5))")M"
+    export POSTGRESQL_SHARED_BUFFERS=${POSTGRESQL_SHARED_BUFFERS:-$shared_buffers_computed}
+    export POSTGRESQL_EFFECTIVE_CACHE_SIZE=${POSTGRESQL_EFFECTIVE_CACHE_SIZE:-$effective_cache}
```
bparees: why the fallback to defaults here?  you know you have a value since you just computed it.

bparees: oops, nm.  just realized the fallback is to the computed value if no value was user-supplied.

9.4/README.md
```diff
@@ -59,6 +60,16 @@ or if it was already present, [`postgres`](http://www.postgresql.org/docs/9.4/st
 is executed and will run as PID 1. You can stop the detached container by running
 `docker stop postgresql_database`.
 
+PostgreSQL auto-tuning
+--------------------
+
+When the PostgreSQL image is ran with the `--memory` parameter set and if there
```
bparees: s/ran/run/

bparees:
update the 9.2 readme also?

mfojtik:
@bparees updated.

9.2/root/usr/bin/cgroup-limits
```diff
@@ -0,0 +1,83 @@
...
+        else:
+            core_count += 1
+
+    env_vars['NUMBER_OF_CORES'] = str(core_count)
+
+if __name__ == "__main__":
+    get_memory_limit()
+    get_number_of_cores()
+
+    print("MAX_MEMORY_LIMIT_IN_BYTES=9223372036854775807")
```
bparees: We need to make this more tolerant, given that we've already been screwing w/ these numbers.  We need to basically say "if the limit is more than some unreasonably high number, assume no limit is set"

9.2/root/usr/bin/cgroup-limits
```diff
@@ -0,0 +1,83 @@
...
+        return
+
+    for group in line.split(','):
+        core_ids = list(map(int, group.split('-')))
+        if len(core_ids) == 2:
+            core_count += core_ids[1] - core_ids[0] + 1
+        else:
+            core_count += 1
+
+    env_vars['NUMBER_OF_CORES'] = str(core_count)
```
php-coder: I would rather write it in a bit different way, by returning value instead of setting it inside of the function:
```python
env_vars['MEMORY_LIMIT_IN_BYTES'] = get_memory_limit()
env_vars['NUMBER_OF_CORES'] = get_number_of_cores()
```
WDYT?

9.2/root/usr/bin/cgroup-limits
php-coder: Why not put this value to the `env_vars` when we're declaring it?

mnagy:
@mfojtik I'm working for an update for the script for sti-base. It'll address issues raised by @bparees and @php-coder as well.

mfojtik:
@mnagy cool, can you tag me in the PR?

mnagy:
fyi, the PR for updated script is here: https://github.com/openshift/sti-base/pull/90

9.2/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -1,6 +1,21 @@
 # Configuration settings.
 export POSTGRESQL_MAX_CONNECTIONS=${POSTGRESQL_MAX_CONNECTIONS:-100}
-export POSTGRESQL_SHARED_BUFFERS=${POSTGRESQL_SHARED_BUFFERS:-32MB}
+
+# Perform auto-tuning based on the container cgroups limits (only when the
+# limits are set).
+# Users can still override this by setting the POSTGRESQL_SHARED_BUFFERS
+# variable.
```
php-coder: Probably, it's better to not mention variable names at all. Or we need to add POSTGRESQL_EFFECTIVE_CACHE_SIZE here too.

mfojtik: we should add it :-)

bparees: +1 to adding the other.

9.2/root/usr/bin/cgroup-limits
mfojtik: @php-coder you have to complain to @mnagy in sti-base :-)

php-coder: @mfojtik Yes, he has already corrected it in sti-base.

9.4/README.md
```diff
@@ -24,6 +24,7 @@ The following environment variables influence the PostgreSQL configuration file.
 | :---------------------------- | ----------------------------------------------------------------------- | -------------------------------
 |  `POSTGRESQL_MAX_CONNECTIONS` | The maximum number of client connections allowed. This also sets the maximum number of prepared transactions. |  100
 |  `POSTGRESQL_SHARED_BUFFERS`  | Sets how much memory is dedicated to PostgreSQL to use for caching data |  32M
+|  `POSTGRESQL_EFFECTIVE_CACHE_SIZE`  | Set to an estimate of how much memory is available for disk caching by the operating system and within the database itself |  128M
```
php-coder: Looks like that the same change for 9.2 is missing.

9.4/README.md
```diff
@@ -59,6 +60,16 @@ or if it was already present, [`postgres`](http://www.postgresql.org/docs/9.4/st
...
 
+PostgreSQL auto-tuning
+--------------------
+
+When the PostgreSQL image is run with the `--memory` parameter set and if there
+are no values provided for `POSTGRESQL_SHARED_BUFFERS` and
+`POSTGRESQL_EFFECTIVE_CACHE_SIZE` those values are automatically calculated
+based on the value provided in the `--memory` parameter.
+
+The values are calculated based on the [upstream](https://wiki.postgresql.org/wiki/Tuning_Your_PostgreSQL_Server) formulas. For the `shared_buffers` we use 1/4 of given memory and for the `effective_cache_size` we
```
php-coder: BTW, it would be good to split this long line.

9.2/root/usr/share/container-scripts/postgresql/common.sh
mfojtik: just a note, I just added it into the comment, the variable is already documented in README

mfojtik:
[test]

9.4/README.md
mfojtik: fixed, good catch.

9.2/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -1,6 +1,22 @@
...
+# Users can still override this by setting the POSTGRESQL_SHARED_BUFFERS
+# and POSTGRESQL_EFFECTIVE_CACHE_SIZE variables.
+if [[ "${NO_MEMORY_LIMIT}" == "true" ]]; then
+    export POSTGRESQL_SHARED_BUFFERS=${POSTGRESQL_SHARED_BUFFERS:-32MB}
+    export POSTGRESQL_EFFECTIVE_CACHE_SIZE=${POSTGRESQL_EFFECTIVE_CACHE_SIZE:-128MB}
+else
+    # Use 1/4 of given memory for shared buffers
+    shared_buffers_computed="$(python -c "print(int((${MEMORY_LIMIT_IN_BYTES}/(1024*1024))*0.25))")M"
+    # Setting effective_cache_size to 1/2 of total memory would be a normal conservative setting,
+    effective_cache="$(python -c "print(int((${MEMORY_LIMIT_IN_BYTES}/(1024*1024))*0.5))")M"
```
mnagy: Python is overly complicated, you can just use bash's $(()) notation for these.

9.2/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -1,6 +1,22 @@
...
+# Perform auto-tuning based on the container cgroups limits (only when the
+# limits are set).
+# Users can still override this by setting the POSTGRESQL_SHARED_BUFFERS
+# and POSTGRESQL_EFFECTIVE_CACHE_SIZE variables.
+if [[ "${NO_MEMORY_LIMIT}" == "true" ]]; then
+    export POSTGRESQL_SHARED_BUFFERS=${POSTGRESQL_SHARED_BUFFERS:-32MB}
+    export POSTGRESQL_EFFECTIVE_CACHE_SIZE=${POSTGRESQL_EFFECTIVE_CACHE_SIZE:-128MB}
+else
+    # Use 1/4 of given memory for shared buffers
+    shared_buffers_computed="$(python -c "print(int((${MEMORY_LIMIT_IN_BYTES}/(1024*1024))*0.25))")M"
```
php-coder: After I accidentally commited `cgroup-limits` file as non-executable I had an error
```
/usr/bin/run-mysqld: line 3: /usr/bin/cgroup-limits: Permission denied
  File "<string>", line 1
    print(int((/(1024*1024))*0.1))
               ^
SyntaxError: invalid syntax
```
So I also added additional check for emptiness of `MEMORY_LIMIT_IN_BYTES`. Maybe you also would like to add it here. Just to be sure.

mfojtik:
[test]

9.2/root/usr/share/container-scripts/postgresql/common.sh
mfojtik: fixed.

9.2/root/usr/share/container-scripts/postgresql/common.sh
mfojtik: fixed.

9.4/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -1,6 +1,21 @@
 # Configuration settings.
 export POSTGRESQL_MAX_CONNECTIONS=${POSTGRESQL_MAX_CONNECTIONS:-100}
-export POSTGRESQL_SHARED_BUFFERS=${POSTGRESQL_SHARED_BUFFERS:-32MB}
+
+# Perform auto-tuning based on the container cgroups limits (only when the
+# limits are set).
+# Users can still override this by setting the POSTGRESQL_SHARED_BUFFERS
+# and POSTGRESQL_EFFECTIVE_CACHE_SIZE variables.
+if [[ "${NO_MEMORY_LIMIT}" == "true" || -z "${MEMORY_LIMIT_IN_BYTES}" ]]; then
```
php-coder: Looks like we also should update if conditions for 9.2.

mfojtik: fixed.

mfojtik:
@php-coder @mnagy any final thoughts? ;-)

php-coder:
LGTM.

twk3:
Issue with the settings in common.sh: 
https://github.com/openshift/postgresql/issues/109

PR #105, opened 2016-03-23
Use ose-3.0 but not epel repos

hhorak:
nss_wrapper is available in ose-3.0 repository and we should make sure `docker build` does not fail if epel repository is not defined at all.

### DIFF
--- a/9.2/Dockerfile.rhel7
+++ b/9.2/Dockerfile.rhel7
@@ -30,11 +30,15 @@ EXPOSE 5432
 # This image must forever use UID 26 for postgres user so our volumes are
 # safe in the future. This should *never* change, the last test is there
 # to make sure of that.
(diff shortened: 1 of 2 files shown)


praiskup:
I'm not sure what to do with `--disable=epel` .. it it still happens that OpenShift's builders
have epel repository enabled by default it would not be 100% safe to _not disable_ it explicitly.

hhorak:
What I see in https://github.com/openshift/sti-python/blob/master/3.4/Dockerfile.rhel7#L25:
```
yum-config-manager --disable epel >/dev/null || : && \
```
should do the trick and not cause any troubles, if epel repo is not defined. Although I think disabling epel repo could be even removed completely, since it is not supported configuration IMO.

hhorak:
Closing this PR, since there is a new one that does the same thing #121.

PR #106, opened 2016-04-23
Support custom configuration file from external URL

nak3:


nak3:
This patch supports environmental variable `POSTGRESQL_CONFIG_FILE` to set custom configuration file.

__Example usage__

~~~
$ docker run -d --name postgresql_database \
   -e POSTGRESQL_CONFIG_URL=https://gist.githubusercontent.com/nak3/67788d11d66460f7fc41ed52719df4b1/raw/1195bab962b1ed3556328b524628a5865491651b/postgresql-download.conf \
   -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 centos/postgresql-94-centos7
~~~
- This command starts postgres with config by downloading from [external location](https://gist.githubusercontent.com/nak3/67788d11d66460f7fc41ed52719df4b1/raw/1195bab962b1ed3556328b524628a5865491651b/postgresql-download.conf).

__Other info__

- The current image has some mechanisms to configurations like discussed [bz#1326525#c2](https://bugzilla.redhat.com/show_bug.cgi?id=1326525#c2).

> - Set the config through environment vars for entries that have already been made configurable
> - Setup dockerbuilds within OpenShift to rebuild an additional layer with your config changes.  You can use an image change trigger to automate this.
> - Mount in config using a volume

- However, the method introduced by this patch is more useful way.

mnagy:
@bparees @mfojtik what do you guys think?

mfojtik:
@mnagy i don't love this idea :-) what if the external URL will be inaccessible (we will have container in restart loop?). Also have great potential to allow middle-man attacks without verifying the checksum or signing the configuration. The security guy in my head is screaming no :-)

bparees:
I can see the value for dev/debug usage.  But @mfojtik's points are spot on...probably better would be to make the postgres image into an s2i builder that can pick up configuration as the "source", similar to what we do with jenkins.  (so you can either use it as is, or you can easily apply your own configuration on top of it via s2i).

mfojtik:
The S2I idea is interesting... It means you will be able to install your own custom configuration or populate the database using assemble/run.

bparees:
you wouldn't want to populate the DB using this mechanism, because then you're using the container storage for your DB which is a Bad Idea[tm].

mfojtik:
@bparees i can seed the DB in the `run` script when the storage is attached. The script can check if the DB is seeded or not and I can have the initial dump stored in the repo with configuration.

bparees:
@mfojtik fair enough.  though you are bloating image sizes.  and it doesn't help w/ schema migration (if anything it might make schema migration more confusing since the user might update their image w/ the new schema, but when the image starts up it mounts the PV from the old instance and so it ignores the image db content since the volume is seeded)

mfojtik:
@bparees i think it depends on the application... take the baseball example marketing use... they just need to insert the stadium data when mongodb start up :-) if the data already exists (iow. there is PV, this will be no-op). I'm saying this is just an example of how the S2I scripts can be used, not saying is the best example everyone should follow :)

bparees:
@mfojtik yeah, i'm just worried about people picking up on it as a pattern.  i agree there can be some use cases.  In fact back when we were first starting this i argued that the DB images should be s2i images for pretty much that reason:  you might want to define the DB schema as part of your assemble script.  (I didn't consider loading data, but that's a small extension)

mfojtik:
@bparees trello: https://trello.com/c/knDOVefH/914-make-database-images-s2i-compatible

praiskup:
I also don't think downloading something from internet is not the best idea.
On the other hand, people still can bind-mount the some script or configuration
into container instead (into precisely defined place). That is done e.g. in
docker-library/postgres or fedora-cloud/Fedora-Dockerfiles/postgres

mfojtik:
@praiskup in openshift/kubernetes they can't unless they use `ConfigMap` which will take over the configuration directory entirely and there are no 'good tools' to manage them. 

Also from the operations point of view, the Docker containers should contain (that also include configuration for the application/service that the given image provide), so having to bind-mount directory with custom configuration might defeat the purpose of having docker image contain (and allow users to get the same behavior without sharing the piece of configuration separately)....

mnagy:
Suppose I wanted to create an image that downloads the config from the net
anyway. How hard would it be if we made postgres into an s2i image?

bparees:
if we made it an s2i image you could write a custom/override run script to pull config down from the internet when the image starts, if you really wanted to.

mfojtik:
@bparees after some thinking it might make sense, but not from internet but from some local internal service which distribute configuration files... still possible to use `run` script for that.

nak3:
Thank you for the discussion all. I agree so please feel free to close this request.

mfojtik:
@nak3 I think you'll be able to fetch the configuration in the "run" script once we make the database images S2I compatible, but I don't know when that happen.

nak3:
Oh, I hope it will be very soon ;) since all of current methods are pains in the neck for my testing...

PR #107, opened 2016-04-25
update instead of replacing passwd file

bparees:


bparees:
@mfojtik @mnagy ptal  (main concern is around the user 26 magic discussed in the Dockerfile)

9.2/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -93,13 +93,22 @@ function generate_postgresql_recovery_config() {
...
-  export GROUP_ID=$(id -g)
-  grep -v ^postgres /etc/passwd > "$HOME/passwd"
-  echo "postgres:x:${USER_ID}:${GROUP_ID}:PostgreSQL Server:${HOME}:/bin/bash" >> "$HOME/passwd"
-  export LD_PRELOAD=libnss_wrapper.so
-  export NSS_WRAPPER_PASSWD=${HOME}/passwd
-  export NSS_WRAPPER_GROUP=/etc/group
+  USER_ID=$(id -u)
+  GROUP_ID=$(id -g)
+
+  if [ x"$USER_ID" != x"0" -a x"$USER_ID" != x"1001" ]; then
```
mnagy: I think the x's are unnecessary here.

9.4/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -93,13 +93,22 @@ function generate_postgresql_recovery_config() {
...
-  export GROUP_ID=$(id -g)
-  grep -v ^postgres /etc/passwd > "$HOME/passwd"
-  echo "postgres:x:${USER_ID}:${GROUP_ID}:PostgreSQL Server:${HOME}:/bin/bash" >> "$HOME/passwd"
-  export LD_PRELOAD=libnss_wrapper.so
-  export NSS_WRAPPER_PASSWD=${HOME}/passwd
-  export NSS_WRAPPER_GROUP=/etc/group
+  USER_ID=$(id -u)
+  GROUP_ID=$(id -g)
+
+  if [ x"$USER_ID" != x"0" -a x"$USER_ID" != x"1001" ]; then
```
mnagy: same as above

9.2/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -93,13 +93,22 @@ function generate_postgresql_recovery_config() {
...
+    NSS_WRAPPER_GROUP=/etc/group
+
+    grep -v ^postgres /etc/passwd > $NSS_WRAPPER_PASSWD
+    echo "postgres:x:${USER_ID}:${GROUP_ID}:PostgreSQL Server:${HOME}:/bin/bash" >> $NSS_WRAPPER_PASSWD
+
+    export NSS_WRAPPER_PASSWD
+    export NSS_WRAPPER_GROUP
+
+    LD_PRELOAD=libnss_wrapper.so
+    export LD_PRELOAD
```
mnagy: Why not one line?

mnagy:
I'm a bit confused about the name of this PR. It seems like you are doing functionally the same thing wrt updating/replacing the passwd file, but you are using a different path.

And why the added userid check for 0 and 1001? Is that a bug fix for some problem?

9.2/root/usr/share/container-scripts/postgresql/common.sh
bparees: it's considered a good practice to ensure full compatibility:
http://stackoverflow.com/questions/174119/why-do-shell-script-comparisons-often-use-xvar-xyes

9.2/root/usr/share/container-scripts/postgresql/common.sh
bparees: symmetry w/ the others?  :) 

I think this was the result of some rework/reordering. i'll change it.

bparees:
@mnagy it's addressing this issue:
https://github.com/openshift/s2i-python/issues/98

the point is to start from the image's /etc/passwd file instead of from a hardcoded template that may not contain all the same users.

the 0 and 1001 thing is copied from the way it was done in the s2i images to avoid setting up nss when we're running as the image user (which is 1001) because the image user exists in /etc/passwd already.  I think it can be removed for this image so i'll do that.

bparees:
@mnagy hm.  ok now i'm not sure what i was doing here.  I think i was just blindly aligning this image w/ the other images i was updating and didn't stop to see this one was already doing the right thing.

will close this.

PR #110, opened 2016-05-06
Bug 1332756: Fix units in computed shared buffers and cache size

mfojtik:


mnagy:
LGTM, feel free to merge when tests pass

PR #111, opened 2016-05-06
README.md: make sure the readme file is distributed

praiskup:
Take this just as an idea ...  The problem is that the image does not
self-describes itself perfectly, pointed by @ppisar:

```
You must either specify the following environment variables:
  POSTGRESQL_USER (regex: '^[a-zA-Z_][a-zA-Z0-9_]*$')
  POSTGRESQL_PASSWORD (regex: '^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$')
  POSTGRESQL_DATABASE (regex: '^[a-zA-Z_][a-zA-Z0-9_]*$')
Or the following environment variable:
  POSTGRESQL_ADMIN_PASSWORD (regex: '^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$')
Or both.
Optional settings:
  POSTGRESQL_MAX_CONNECTIONS (default: 100)
  POSTGRESQL_SHARED_BUFFERS (default: 32MB)
```

The most correct documentation is in README.md at the time of image is built (later
it may not match 100%), so we could potentially at least distribute it with the image itself?

@eliskasl  mentioned the 'atomic help' idea, which is almost equivalent to README.md, some
kind of alternative documentation file in man format.

Or we can potentially have a look at documentation idea implemented in 'fedora/postgresql' or
'praiskup/postgresql' (documentation provided by image itself, hierarchical documentation
topics).


praiskup:
@pkajaba, @hhorak fyi

hhorak:
Maybe we can place it directly to some final destination, like `/usr/share/container-scripts/postgresql/` or something like that?

praiskup:
Make sense, but the path is not easy to remember - so I've edited the basic usage
info -- it now mentions where the README.md is located within container.

hhorak:
[test]

9.2/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -35,16 +35,21 @@ function usage() {
...
+Or the following environment variable:
+  POSTGRESQL_ADMIN_PASSWORD (regex: '$psql_password_regex')
+Or both.
+Optional settings:
+  POSTGRESQL_MAX_CONNECTIONS (default: 100)
+  POSTGRESQL_SHARED_BUFFERS (default: 32MB)
+
+For more info have a look at /usr/share/container-scripts/postgresql/README.md
+file within container or visit https://github.com/openshift/postgresql page.
+EOF
```
mnagy: +1 for this

mnagy:
I'm not against distributing the README.md, but I'm not sure how it would be useful for someone. I mean, how would a user find it? However, if we want to include it in the image, let's just add it in a Dockerfile and not move it to some obscure place in the repository.

@bparees thoughts on the readme distribution?

9.2/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -35,16 +35,21 @@ function usage() {
...
+  POSTGRESQL_PASSWORD (regex: '$psql_password_regex')
+  POSTGRESQL_DATABASE (regex: '$psql_identifier_regex')
+Or the following environment variable:
+  POSTGRESQL_ADMIN_PASSWORD (regex: '$psql_password_regex')
+Or both.
+Optional settings:
+  POSTGRESQL_MAX_CONNECTIONS (default: 100)
+  POSTGRESQL_SHARED_BUFFERS (default: 32MB)
+
+For more info have a look at /usr/share/container-scripts/postgresql/README.md
```
bparees: "For more information see /usr/share/container-scripts/postgresql/README.md within the container or visit https://github.com/openshift/postgresql."

bparees:
>  However, if we want to include it in the image, let's just add it in a Dockerfile and not move it to some obscure place in the repository.

not sure what you mean.  Even if you explicitly add it in the Dockerfile, you have to pick a path within the image to add it to.  The approach taken here avoids adding another layer/command to the dockerfile so i think it's preferred (even if we squash, layers take more time to build).

bparees:
I guess you're saying the Dockerfile should add it from its existing location in the git repo, into the image.  That does still require an additional ADD command in the Dockerfile, but I think I agree it is better than hiding the file within the repo.

mnagy:
Yes. Maybe we can move it and create a link so it's stil visible when you
are in the version dir.

praiskup:
I've added symbolic link in the original commit, or I probably miss the point here?
Can you elaborate what exactly is needed now?

bparees:
@praiskup sorry i missed the symlink, i think that's fine, though I think other people have sometimes expressed concerns about symlinks in repos.

can you squash the commits and assuming on one objects to the symlink we can merge this.

praiskup:
np, squashed

bparees:
[test]

mnagy:
LGTM

bparees:
lgtm, thanks @praiskup!

PR #119, opened 2016-06-07
set cgroup limit env variables in slave script

bparees:
should fix https://github.com/openshift/origin/issues/9188

@gabemontero


bparees:
[test]

PR #120, opened 2016-06-13
fix recursive handling of links

gabemontero:
@bparees PTAL

### DIFF
--- a/9.2/root/usr/libexec/fix-permissions
+++ b/9.2/root/usr/libexec/fix-permissions
@@ -1,7 +1,7 @@
 #!/bin/sh
-# Fix permissions on the given directory to allow group read/write of
+# Fix permissions on the given directory to allow group read/write of 
 # regular files and execute of directories.
-chown -R postgres "$1"
-chgrp -R 0 "$1"
-chmod -R g+rw "$1"
+find "$1" -exec chown postgres {} \;
+find "$1" -exec chgrp 0 {} \;
+find "$1" -exec chmod g+rw {} \;
 find "$1" -type d -exec chmod g+x {} +
--- a/9.4/root/usr/libexec/fix-permissions
+++ b/9.4/root/usr/libexec/fix-permissions
@@ -1,7 +1,7 @@
 #!/bin/sh
-# Fix permissions on the given directory to allow group read/write of
+# Fix permissions on the given directory to allow group read/write of 
 # regular files and execute of directories.
-chown -R postgres "$1"
-chgrp -R 0 "$1"
-chmod -R g+rw "$1"
+find "$1" -exec chown postgres {} \;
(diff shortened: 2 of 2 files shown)


bparees:
doesn't look good....
[test]

gabemontero:
There several of these:

```
psql: could not connect to server: Connection refused
	Is the server running on host "172.17.0.2" and accepting
	TCP/IP connections on port 5432?
```

And at the end, one of these:

```
mkdir: cannot create directory '/var/lib/pgsql/data/userdata': Permission denied
```

And there is this in the Dockerfile:
`/usr/libexec/fix-permissions /var/lib/pgsql`

So something is amiss with the finds

gabemontero:
I see what is wrong ... will be submitting update ... my diffs somehow missed the differences in the postgres fix-permission vs. the other ones

gabemontero:
[test]

gabemontero:
ok @bparees the test check is happy again

9.2/root/usr/libexec/fix-permissions
```diff
@@ -1,7 +1,7 @@
...
+# Fix permissions on the given directory to allow group read/write of 
 # regular files and execute of directories.
-chown -R postgres "$1"
-chgrp -R 0 "$1"
-chmod -R g+rw "$1"
-find "$1" -type d -exec chmod g+x {} +
+find $1 -exec chown postgres {} \;
+find $1 -exec chgrp 0 {} \;
+find $1 -exec chmod g+rw {} \;
+find $1 -type d -exec chmod g+x {} +
```
bparees: Quotes should remain for safety (path could contain a space)

gabemontero: quotes readded

bparees:
lgtm

praiskup:
This is not nice change :( I miss the reasoning.  Why is 'find -exec' better
than 'chown -R'?

bparees:
@praiskup chmod and chgrp and presumably chown, do not change the ownership of symlinked files when used with -R.

see:
https://github.com/openshift/jenkins/pull/110#discussion_r66736478

praiskup:
Did we found a reasons why coreutils  'chmod' does not support --[no]-dereference option, '-H' or '-L'? Is this really missing feature?

Why we don't use those options where possible?  When we don't have to care about portability,
find && xargs should be used (to not waste forks).

Is this really what we want?  To allow symbolic links and then to fix the destination?

What is the most important message, we should have documentation in commit message.

praiskup:
Note there is also '-exec {} +' variant.  Here is the bug asking for missing '-L' option in chmod:
https://bugzilla.redhat.com/show_bug.cgi?id=1346074

PR #121, opened 2016-06-17
Use OSE 3.2 repos for nss_wrapper

hhorak:
Use OSE 3.2 repos for nss_wrapper and do not disable epel repository, because it makes the build failing when epel is not defined at all.

### DIFF
--- a/9.2/Dockerfile.rhel7
+++ b/9.2/Dockerfile.rhel7
@@ -30,11 +30,15 @@ EXPOSE 5432
 # This image must forever use UID 26 for postgres user so our volumes are
 # safe in the future. This should *never* change, the last test is there
 # to make sure of that.
+# rhel-7-server-ose-3.2-rpms is enabled for nss_wrapper until this pkg is
+# in base RHEL
 RUN yum install -y yum-utils && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
+    yum-config-manager --enable rhel-7-server-ose-3.2-rpms && \
+    yum-config-manager --disable epel >/dev/null && \
     INSTALL_PKGS="rsync tar gettext bind-utils postgresql92 postgresql92-postgresql-contrib nss_wrapper" && \
-    yum install -y --disablerepo="epel" --setopt=tsflags=nodocs $INSTALL_PKGS && \
+    yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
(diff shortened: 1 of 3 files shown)


mnagy:
```
[91mwarning: /var/cache/yum/x86_64/7Server/epel/packages/nss_wrapper-1.1.3-1.el7.x86_64.rpm: Header V3 RSA/SHA256 Signature, key ID 352c64e5: NOKEY
[0mPublic key for nss_wrapper-1.1.3-1.el7.x86_64.rpm is not installed
```

hhorak:
Hm, my opinion is that this error is happening when building on machines that have epel enabled and nss_wrapper is taken from there. That is the use case I was missing, so we indeed need to disable epel, just doing it safely like this:
```
yum-config-manager --disable epel >/dev/null || : && \
```
Does it make sense?

mnagy:
@hhorak why do you want to ignore a possible error?

praiskup:
Because 'epel' is not available on correctly set-up builder -- and in this case
the yum woudl fail saying "there is no epel channel".

mnagy:
And what about the `--disablerepo="epel"` that was there before?

praiskup:
I actually checked the 'yum-config-manager --disable NONEXISTENT' and the error was not thrown out, exit 0.  Yum --disablerepo fails with NONEXISTENT repo.

mnagy:
ok, so we don't need to ignore an error, cool

praiskup:
> ok, so we don't need to ignore an error, cool

I would call it bug in yum-config-manager.  Nowadays you have 'dnf config-manager' which
properly fails with non-zero status.  Fixing 'yum-config-manager' is highly unlikely (deprecated
nowadays).

bparees:
lgtm, ready for merge?

praiskup:
+# rhel-7-server-ose-3.2-rpms is enabled for nss_wrapper until this pkg is
+# in base RHEL

.. FWIW, I still think we should remove the LD_PRELOAD hack completely.

bparees:
And do what instead?

Ben Parees | OpenShift

praiskup:
Ah, nostalgia in https://github.com/sclorg/postgresql-container/pull/10 :), there were some ideas.
I still think SELinux (mcs) is the answer for security concerns in arbitrary GID approach.

praiskup:
I meant https://github.com/sclorg/rhscl2dockerfile/issues/18

hhorak:
Already fixed by b3f42ca30ec6ae6.

PR #122, opened 2016-06-20
Add enable/disable replication option.

dtschan:
This pull request adds an option to enable or disable replication while keeping the current default configuration. Rationale:
The OpenShift Enterprise 3.2 postgresql-persistent templates use images built from this repository and persistent volumes with a default size between 0.5 to 1 GiB. However `openshift-custom-postgresql.conf.template` configures a WAL size of 6.25 GiB (400 segments of 16 MiB) for replication purposes which means that most users PostgreSQL containers will fail after some time with an out of disk error. Not all users have larger persistent volumes available so increasing the default persistent volume is not always be an option. On the other hand not all users and use cases require replication.

### DIFF
--- a/9.2/root/usr/bin/run-postgresql
+++ b/9.2/root/usr/bin/run-postgresql
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+export ENABLE_REPLICATION=${ENABLE_REPLICATION:-false}
+
 set -eu
 export_vars=$(cgroup-limits) ; export $export_vars
 
--- a/9.2/root/usr/bin/run-postgresql-master
+++ b/9.2/root/usr/bin/run-postgresql-master
@@ -1,1 +1,5 @@
-run-postgresql
+#!/bin/bash
+
+export ENABLE_REPLICATION=true
+
+exec run-postgresql "$@"
--- a/9.2/root/usr/bin/run-postgresql-slave
+++ b/9.2/root/usr/bin/run-postgresql-slave
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+export ENABLE_REPLICATION=true
+
 set -eu
 export_vars=$(cgroup-limits) ; export $export_vars
 
--- a/9.2/root/usr/share/container-scripts/postgresql/common.sh
+++ b/9.2/root/usr/share/container-scripts/postgresql/common.sh
@@ -106,6 +106,12 @@ function generate_postgresql_config() {
   envsubst \
       < "${CONTAINER_SCRIPTS_PATH}/openshift-custom-postgresql.conf.template" \
       > "${POSTGRESQL_CONFIG_FILE}"
+
+  if [ "${ENABLE_REPLICATION}" == "true" ]; then
+    envsubst \
+        < "${CONTAINER_SCRIPTS_PATH}/openshift-custom-postgresql-replication.conf.template" \
+        >> "${POSTGRESQL_CONFIG_FILE}"
+  fi
 }
 
 function generate_postgresql_recovery_config() {
--- a/9.2/root/usr/share/container-scripts/postgresql/openshift-custom-postgresql-replication.conf.template
+++ b/9.2/root/usr/share/container-scripts/postgresql/openshift-custom-postgresql-replication.conf.template
@@ -1,0 +1,7 @@
+# required on master for replication
+wal_level = hot_standby         # minimal, archive, hot_standby, or logical
+max_wal_senders = 6             # max number of walsender processes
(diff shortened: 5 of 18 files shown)


9.2/root/usr/share/container-scripts/postgresql/README.md
```diff
@@ -26,6 +26,7 @@ The following environment variables influence the PostgreSQL configuration file.
 |  `POSTGRESQL_MAX_PREPARED_TRANSACTIONS` | Sets the maximum number of transactions that can be in the "prepared" state. If you are using prepared transactions, you will probably want this to be at least as large as max_connections |  0
 |  `POSTGRESQL_SHARED_BUFFERS`  | Sets how much memory is dedicated to PostgreSQL to use for caching data |  32M
 |  `POSTGRESQL_EFFECTIVE_CACHE_SIZE`  | Set to an estimate of how much memory is available for disk caching by the operating system and within the database itself |  128M
+|  `POSTGRESQL_ENABLE_REPLICATION`  | Streaming replication is enabled by default and requires up to 6.25 GiB additional disk space. Set to 0 if you don't need replication and want to run PostgreSQL with a small persistent volume |  1
```
bparees: we've typically tried to use true/false instead of 0/1 for these sort of things.

bparees:
one nit, otherwise lgtm.
@hhorak you want me to start delegating these sort of PRs to your team for review/merge?

bparees:
[test]

hhorak:
@bparees I'll try ;)

hhorak:
@dtschan I might miss something here, but wouldn't it be better to not using the replication configuration by default (when running with defualt CMD) and use the replication config only for `run-postgresql-master` CMD?
@praiskup what do you think?

praiskup:
I'm probably lost, I was also about to believe that there is separate CMD for replication.

praiskup:
OK, master shares configuration with default non-replicating server, which is probably not
the right thing.  There are 'run-postgresql', 'run-postgresql-master' and 'run-postgresql-slave'
candidates for CMD, we shouldn't have another option to turn the replication on/off.

hhorak:
yeah, there is a symlink, because it seemed like master server will be the same as non-replication server. Since we see it is no longer truth, we shouldn't use the symlink and have a separate CMDs for both use cases.

dtschan:
@hhorak Ah, I didn't see that there are multiple CMDs available. Using replication config only for `run-postgresql-master` makes more sense of course. However there currently doesn't seem to be a way to override CMD in the standard OpenShift deployment strategies. What we'd like to have in the end is a way to offer stable PostgreSQL containers to users which only have small volumes available (i.e. without replication) based on the supported images that come with OpenShift Enterprise.

hhorak:
@dtschan This is the piece I'm not familiar with much so I might miss something, but I understand it like the CMD that is used even for small volumes (without replication) uses run-postgresql-master, which would be wrong.

bparees:
you can override the container entrypoint in the container definition within the podtemplate (which is part of a deploymentconfig definition)

phil-pona:
@bparees Thanks.
We tried overwriting the command in the deploymentconfig, http://kubernetes.io/docs/user-guide/containers/, which worked fine for us.

The only thing left to do is to provide a separate CMD to be able to run postgresql in master and non replication mode as described in the notes above.

omron93:
[test]

dtschan:
Updated pull request to current state of discussion, particulary comments https://github.com/sclorg/postgresql-container/pull/122#issuecomment-227340527

>  OK, master shares configuration with default non-replicating server, which is probably not the right thing. There are 'run-postgresql', 'run-postgresql-master' and 'run-postgresql-slave' candidates for CMD, we shouldn't have another option to turn the replication on/off.

 and https://github.com/sclorg/postgresql-container/pull/122#issuecomment-227344919

> yeah, there is a symlink, because it seemed like master server will be the same as non-replication server. Since we see it is no longer truth, we shouldn't use the symlink and have a separate CMDs for both use cases.

`run-postgresql` now no longer has replication enabled, while `run-postgresql-master` and `run-postgresql-slave` still have. Please review again.

praiskup:
[test]

praiskup:
ACK here.

dtschan:
Forgot to pass arguments of {{run-postgresql-master}} along to {{run-postgresql}}, fixed in commit 0b7ae60.

praiskup:
Thanks for re-reveiw :) can you squash?

dtschan:
Squashed.

praiskup:
Thanks. ACK.

praiskup:
[test]

hhorak:
LGTM.

PR #124, opened 2016-07-12
Added cccp.yml files for CentOS Container Pipeline

mohammedzee1000:
This is for building containers on the centos container
pipeline.


hhorak:
This PR does not influence building, so merging.

PR #126, opened 2016-08-01
fix owner of mount directory for persistant.

minimaru:
I am using NFS for persistant volume.
However, can't write data.
Cause was the permission of the directory.
So, I fixed owner of directory "/var/lib/pgsql/data".

### DIFF
--- a/9.2/Dockerfile
+++ b/9.2/Dockerfile
@@ -32,7 +32,7 @@ RUN yum install -y centos-release-scl epel-release && \
     rpm -V $INSTALL_PKGS && \
     yum clean all && \
     localedef -f UTF-8 -i en_US en_US.UTF-8 && \
-    mkdir -p /var/lib/pgsql/data && \
+    mkdir -p /var/lib/pgsql/data && chown postgres /var/lib/pgsql/data && \
     test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)"
 
 # Loosen permission bits to avoid problems running container with arbitrary UID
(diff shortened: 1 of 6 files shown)


bparees:
this is not the right solution to your problem.  When NFS volumes (and any persistent volume) are setup correctly in your openshift cluster, the container will be able to write to them without this change.

as you can see here, the image already creates a subdirectory under "data" that is owned by the container user (postgres):
https://github.com/sclorg/postgresql-container/blob/master/9.5/root/usr/share/container-scripts/postgresql/common.sh#L196

minimaru:
I use docker "PostgreSQL" container with "**OpenShift Origin**".

When "OpenShift" try to mount directory which is created by ROOT,
"OpenShift" can't mount that directory.

So, "MySQL" container, 
in "Dockerfile", change owner as following.

https://github.com/docker-library/mysql/blob/master/5.7/Dockerfile#L47

"PostgreSQL" container you published 
makes "/var/lib/pgsql/data" directory by "ROOT" not like above "MySQL" container,
so "OpenShift Origin" can't NFS-mount "data" directory.

bparees:
/var/lib/pgsql/data is being bind-mounted by your PV, so it is not going to be owned by root, the ownership will depend on the underlying NFS volume you are mounting and openshift permissions.

If you're not bind-mounting into /var/lib/pgsql/data then you are using the image incorrectly:
https://github.com/openshift/origin/blob/master/examples/db-templates/postgresql-persistent-template.json#L145-L150

PR #131, opened 2016-08-25
Add master-slave replication tests

pkubatrh:
This PR adds a simple test for the master-slave replication scenario and was written to be similar to mariadb's replication test.
@praiskup Please take a look at this


praiskup:
[test]

praiskup:
Thanks for the progress, looks like a very good start!

9.2/test/run
```diff
@@ -255,6 +255,66 @@ function run_tests() {
...
+
+  # Run the PostgreSQL master
+  docker run $cluster_args -p 5432  \
+    -d --cidfile ${CIDFILE_DIR}/master.cid $IMAGE_NAME run-postgresql-master >/dev/null
+  local master_ip
+  master_ip=$(get_container_ip master.cid)
+
+  # Run the PostgreSQL slave
+  docker run $cluster_args -e POSTGRESQL_MASTER_IP=${master_ip} \
+    -d --cidfile ${CIDFILE_DIR}/slave.cid $IMAGE_NAME run-postgresql-slave >/dev/null
```
praiskup: We should have somehow more "abstract" 'start_master' action and 'start_slave' functions
to be able easily change number of slaves.  But this might be extracted in follow up patches.

9.2/test/run
```diff
@@ -255,6 +255,66 @@ function run_tests() {
...
+  # Read value from slave and check whether it is expected
+  for i in $(seq $max_attempts); do
+    set +e
+    result="$(postgresql_cmd -c "select * from t1" | grep $value)"
+    set -e
+    if [[ ! -z "${result}" ]]; then
+      echo "${slave_ip} successfully got value from MASTER ${master_ip}"
+      break
+    fi
+    if [[ "${i}" == "${max_attempts}" ]]; then
```
praiskup: Just a nit:  I would use '-ge' operator here.

praiskup: Ah, ok -- that is not infinite loop.  Then this condition is probably not needed at all.

9.2/test/run
```diff
@@ -255,6 +255,66 @@ function run_tests() {
...
+  local value
+  value=24
+  # Do some real work to test replication in practice
+  postgresql_cmd -c "CREATE TABLE t1 (a integer); INSERT INTO t1 VALUES ($value);"
+
+  CONTAINER_IP=$slave_ip
+  # Read value from slave and check whether it is expected
+  for i in $(seq $max_attempts); do
+    set +e
+    result="$(postgresql_cmd -c "select * from t1" | grep $value)"
```
praiskup: Here you can use 'psql -tA', instead of "grep", then you can avoid also the 'set -e / set +e'.

bparees:
[test]

bparees:
@praiskup if you join the sclorg organization you should have test rights.

pkubatrh:
Incorporated suggestions from @praiskup 
Created run_master and run_slave functions, modified original test to be able to create multiple slaves.

praiskup:
'diff 9.{2,5}/test/run' command shows some differences ATM

pkubatrh:
@praiskup Fixed, thanks for noticing.

praiskup:
Just an idea, @pkubatrh could you make this patch
 https://github.com/praiskup/openshift-postgresql/commit/47f0841a25e18f
applied before/after your patch in this PR?  We can do it later of course.

pkubatrh:
Going to take a look at it

pkubatrh:
Incorporated praiskup's de-duplication patch into the PR.

praiskup:
We are not in sclorg yet, I don't see a "request for membership" button in sclorg... Can somebody test?

As the de-duplication patch is applied *after* master-slave, it is unnecessarily big set of patches :(.  But otherwise LGTM.  Please don't squash.

hack/run_test
```diff
@@ -0,0 +1,424 @@
...
+#
+# IMAGE_NAME specifies the name of the candidate image used for testing.
+# The image has to be available before this script is executed.
+#
+
+set -exo nounset
+shopt -s nullglob
+
+test -n "${IMAGE_NAME-}" || false 'make sure $IMAGE_NAME is defined'
+test -n "${VERSION-}" || false 'make sure $VERSION is defined'
```
praiskup: Ach, missed that you've changed the patch ... why is there the ${VERSION-} instead of ${VERSION}?

praiskup: Ah, the 'nounset' :( ok.

bparees:
@hhorak can @praiskup be added to sclorg so he can run tests?
[test]

hack/run_test
praiskup: Turns out that the $VERSION is probably not properly exported for 'make test' now.

praiskup:
We'll manage sclorg membership once Honza is back from PTO.

pkubatrh:
$VERSION is now set properly before calling tests.
@bparees please test again

bparees:
[test]

praiskup:
Just for inspiration, next time I would use something like this (see the latest 3 commits):
https://github.com/praiskup/openshift-postgresql/commits/easier-to-follow

That way anybody can very easily check that the commits don't do anything nasty :) (what
the changes actually really are).

That is the reason why I dislike squashing (what that makes sense of course).

hhorak:
@bparees I've added @pkubatrh and @praiskup to the org, so hopefully the testing should now work for them. Guys, feel free to try it.

hhorak:
[test]

hhorak:
I think this is a very good start for testing replication, more test cases (e.g. slave crash + simple failover) may be added later. LGTM

PR #132, opened 2016-08-30
don't use sudo in scripts

praiskup:
Sudo keeps asking for password (with the default config) which
makes the script interactive.


praiskup:
[test]

bparees:
lgtm

PR #133, opened 2016-08-31
Add a test for master restart after being killed

pkubatrh:
This PR adds a simple test to check if a master can be properly restarted after being killed in a replication scenario.
First commit makes some necessary changes to existing test code while the second commit adds the actual test.

### DIFF
--- a/hack/run_test
+++ b/hack/run_test
@@ -269,27 +269,14 @@ function run_master() {
     -d --cidfile ${CIDFILE_DIR}/master-${suffix}.cid $IMAGE_NAME run-postgresql-master >/dev/null
 }
 
-function run_replication_test() {
-  echo "Testing master-slave replication"
-  local cluster_args="-e POSTGRESQL_ADMIN_PASSWORD=pass -e POSTGRESQL_MASTER_USER=$PGUSER -e POSTGRESQL_MASTER_PASSWORD=$PASS"
+function test_slave_visibility() {
+  local cid_suffix="$1"; shift
   local max_attempts=30
-  local cid_suffix="basic"
 
-  # Run the PostgreSQL master
-  run_master $cid_suffix
-  local master_ip
-  master_ip=$(get_container_ip master-$cid_suffix.cid)
-
-  # Run the PostgreSQL slaves
-  local i
-  for i in $(seq ${SLAVE_NUM:-1}); do
-    run_slave $master_ip $cid_suffix-$i
-  done
-
-  # Check if the master knows about the slaves
-  CONTAINER_IP=$master_ip
   local f
   local slave_ip
+  local cidfile
+  local master_ip=$(get_container_ip master-$cid_suffix.cid)
   for f in `find ${CIDFILE_DIR} -type f -name "slave-$cid_suffix-*"`; do
     cidfile=$(basename $f)
     slave_ip=$(get_container_ip $cidfile)
@@ -308,21 +295,29 @@ function run_replication_test() {
       sleep 1
     done
   done
+}
 
+function test_value_replication() {
+  local table_name="$1"; shift
+  local max_attempts=30
+
+  # Setup the replication data
   local value
   value=24
-  # Do some real work to test replication in practice
-  postgresql_cmd -c "CREATE TABLE t1 (a integer); INSERT INTO t1 VALUES ($value);"
+  local table_name="t1"
+  postgresql_cmd -c "CREATE TABLE $table_name (a integer); INSERT INTO $table_name VALUES ($value);"
 
   # Read value from slaves and check whether it is expected
   local f
   local slave_ip
+  local cidfile
+  local master_ip=$(get_container_ip master-$cid_suffix.cid)
   for f in `find ${CIDFILE_DIR} -type f -name "slave-*"`; do
     cidfile=$(basename $f)
     slave_ip=$(get_container_ip $cidfile)
     CONTAINER_IP=$slave_ip
     for i in $(seq $max_attempts); do
-      result="$(postgresql_cmd -At -c "select * from t1")"
+      result="$(postgresql_cmd -At -c "select * from $table_name")"
       if [[ ! -z "${result}" ]]; then
         echo "${slave_ip} successfully got value from MASTER ${master_ip}"
         break
@@ -336,6 +331,36 @@ function run_replication_test() {
       sleep 1
     done
   done
+}
+
+function setup_replication_cluster() {
+  # Run the PostgreSQL master
+  run_master $cid_suffix
+  local master_ip
+  master_ip=$(get_container_ip master-$cid_suffix.cid)
+
+  # Run the PostgreSQL slaves
+  local i
+  for i in $(seq ${SLAVE_NUM:-1}); do
+    run_slave $master_ip $cid_suffix-$i
+  done
+
+  # Check if the master knows about the slaves
+  CONTAINER_IP=$master_ip
+  test_slave_visibility $cid_suffix
+}
+
+function run_replication_test() {
+  echo "Testing master-slave replication"
+  local cluster_args="-e POSTGRESQL_ADMIN_PASSWORD=pass -e POSTGRESQL_MASTER_USER=$PGUSER -e POSTGRESQL_MASTER_PASSWORD=$PASS"
+  local cid_suffix="basic"
+  local table_name="t1"
+
+  # Setup the cluster
+  setup_replication_cluster
+
+  # Do some real work to test replication in practice
+  test_value_replication $table_name
 }
 
 function run_change_password_test() {
--- a/hack/run_test
+++ b/hack/run_test
@@ -32,6 +32,17 @@ function cleanup() {
 }
 trap cleanup EXIT
 
+function cleanup_volume() {
+  local volume_dir="$1"; shift
+  local cid="$1"; shift
+  local volume_args="-v ${volume_dir}:/var/lib/pgsql/data:Z"
+
+  docker stop $cid
+  local rm_cmd='/bin/rm -rf /var/lib/pgsql/data/*'
+  docker run $volume_args --rm $IMAGE_NAME /bin/sh -c "$rm_cmd"
+  rmdir ${volume_dir}
+}
+
 function get_cid() {
   local id="$1" ; shift || return 1
   echo $(cat "$CIDFILE_DIR/$id")
@@ -97,6 +108,13 @@ function create_container() {
   # create container with a cidfile in a directory for cleanup
   docker run $cargs --cidfile $cidfile -d $IMAGE_NAME "$@"
   echo "Created container $(cat $cidfile)"
+}
+
+function create_volume() {
+  local volume_dir
+  volume_dir=`mktemp -d --tmpdir pg-testdata.XXXXX`
+  chmod a+rwx ${volume_dir}
+  echo $volume_dir
 }
 
 function assert_login_access() {
@@ -269,7 +287,8 @@ function run_slave() {
 
 function run_master() {
   local suffix="$1"; shift
-  docker run $cluster_args -p 5432  \
+  master_args=${master_args-}
+  docker run $cluster_args $master_args \
     -d --cidfile ${CIDFILE_DIR}/master-${suffix}.cid $IMAGE_NAME run-postgresql-master >/dev/null
 }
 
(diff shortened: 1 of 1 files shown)


pkubatrh:
[test]

bparees:
[test]

praiskup:
Thanks.  I wouldn't use a word "failover" as that usually means different usecase in PostgreSQL context.  I'll do a better review tomorrow.

hack/run_test
```diff
@@ -27,6 +24,13 @@ function cleanup() {
     docker rm $CONTAINER
     rm $cidfile
     echo "Done."
+}
+
+function cleanup() {
+  for cidfile in $CIDFILE_DIR/* ; do
+    CONTAINER=$(cat $cidfile)
+
+    cleanup_container
```
praiskup: I would move this into the second patch, otherwise this "function split" looks like
unnecessary (when somebody looks at the patch separately).

hack/run_test
```diff
@@ -259,7 +263,7 @@ function run_tests() {
 function run_slave() {
   local master_ip="$1"; shift
   local suffix="$1"; shift
-  docker run $cluster_args -e POSTGRESQL_MASTER_IP=${master_ip} \
+  docker run $cluster_args -p 5432 -e POSTGRESQL_MASTER_IP=${master_ip} \
```
praiskup: Why is this needed? (missing comment probably).

Can we connect the master/slave containers without exposing ports from host system?

hack/run_test
```diff
@@ -311,6 +311,52 @@ function setup_replication_cluster() {
...
 
+function run_failover_test {
+  echo "Testing failover"
+  local cluster_args="-e POSTGRESQL_ADMIN_PASSWORD=pass -e POSTGRESQL_MASTER_USER=$PGUSER -e POSTGRESQL_MASTER_PASSWORD=$PASS"
+  local cid_suffix="failover"
+
+  # Setup the cluster
+  setup_replication_cluster
+
+  # Kill the master
```
praiskup: Could we use something like `: Kill the master`, instead of `# Kill the master`...  so we can see from log what is happening during 'make test'.

hack/run_test
pkubatrh: Thanks for spotting this. This is actually leftover code from me trying to figure out why the the new slave doesn't connect to the master. 
Will remove the change in the next push as it is not needed.

hack/run_test
praiskup: Also, I would consider `docker kill` to really shoot the postgres server, `docker stop` rather carefully stops the server.

praiskup:
Ah, reading the patch carefully, there really is failover scenario tested (that is, however, probably not usable in OpenShift at least ATM).

praiskup:
IOW: promoting slave container to master container is hard task enough to rather start new
master...

pkubatrh:
Updated the PR with a new test case (master restart instead of failover).

hack/run_test
```diff
@@ -308,21 +294,28 @@ function run_replication_test() {
...
+
+function test_value_replication() {
+  local table_name="$1"; shift
 
+  # Setup the replication data
   local value
   value=24
-  # Do some real work to test replication in practice
-  postgresql_cmd -c "CREATE TABLE t1 (a integer); INSERT INTO t1 VALUES ($value);"
+  local table_name="t1"
```
praiskup: This overwrites previous variable assignment.

hack/run_test
```diff
@@ -338,6 +331,37 @@ function run_replication_test() {
...
+  max_attempts=30
+
+  # Setup the cluster
+  setup_replication_cluster
+
+  # Do some real work to test replication in practice
+  test_value_replication $table_name
+}
+
```
praiskup: Please don't move the functions, if possible.  It is harder to review (from the diff is not obvious what changed within function).

hack/run_test
```diff
@@ -308,21 +294,28 @@ function run_replication_test() {
...
   local slave_ip
+  local cidfile
+  local master_ip=$(get_container_ip master-$cid_suffix.cid)
   for f in `find ${CIDFILE_DIR} -type f -name "slave-*"`; do
     cidfile=$(basename $f)
     slave_ip=$(get_container_ip $cidfile)
     CONTAINER_IP=$slave_ip
     for i in $(seq $max_attempts); do
-      result="$(postgresql_cmd -At -c "select * from t1")"
+      result="$(postgresql_cmd -At -c "select * from $table_name")"
```
praiskup: Once there are variables, you can consider 'psql --set' option.  We are in tests only, so it is safe so far.

hack/run_test
```diff
@@ -338,6 +331,37 @@ function run_replication_test() {
...
+  CONTAINER_IP=$master_ip
+  test_slave_visibility $cid_suffix
+}
+
+function run_replication_test() {
+  echo "Testing master-slave replication"
+  local cluster_args="-e POSTGRESQL_ADMIN_PASSWORD=pass -e POSTGRESQL_MASTER_USER=$PGUSER -e POSTGRESQL_MASTER_PASSWORD=$PASS"
+  local cid_suffix="basic"
+  local table_name="t1"
+  max_attempts=30
```
praiskup: Not a local variable, it is not obvious this is expected.

pkubatrh:
New push, fixed issues pointed out by @praiskup (overwritten variable + added a comment to global variable)

hack/run_test
```diff
@@ -308,21 +294,28 @@ function run_replication_test() {
...
+  local cidfile
+  local master_ip=$(get_container_ip master-$cid_suffix.cid)
   for f in `find ${CIDFILE_DIR} -type f -name "slave-*"`; do
     cidfile=$(basename $f)
     slave_ip=$(get_container_ip $cidfile)
     CONTAINER_IP=$slave_ip
     for i in $(seq $max_attempts); do
-      result="$(postgresql_cmd -At -c "select * from t1")"
+      result="$(postgresql_cmd -At -c "select * from $table_name")"
       if [[ ! -z "${result}" ]]; then
```
praiskup: This is probably OK, but I would rather fix it sooner than later ... checking psql output 
for (non)empty-ness is not perfect, because error messages comes to standard output ->
thus empty string comes from invalid query.  We are fine because 'set -e' is activated ..
but that could change in future...  I would vote for explicit string to compare $result with.

hack/run_test
```diff
@@ -348,11 +350,51 @@ function setup_replication_cluster() {
...
+  docker run $master_args --rm $IMAGE_NAME /bin/sh -c "$rm_cmd"
+  rmdir ${volume_dir}
+}
+
 function run_replication_test() {
   echo "Testing master-slave replication"
   local cluster_args="-e POSTGRESQL_ADMIN_PASSWORD=pass -e POSTGRESQL_MASTER_USER=$PGUSER -e POSTGRESQL_MASTER_PASSWORD=$PASS"
   local cid_suffix="basic"
   local table_name="t1"
+  # Global variable for setup_replication_cluster and test_value_replication
```
praiskup: Why you can't use local variable?

pkubatrh: With a setting such as this that is used in polling loops and probably won't be changed much from its default value, I felt like it would be best to use it as a global variable. Although with the addition of another test I am now setting it in two places so I will fix that.

pkubatrh: Scratch that, I see it's found in other parts of the code as well. Will make it local then.

pkubatrh:
Rewrote non-emptyness check into an expected value check, max_attempts variable no longer global.

hack/run_test
```diff
@@ -350,6 +352,43 @@ function setup_replication_cluster() {
...
+  test_slave_visibility $cid_suffix
+
+  # Check if the replication works
+  test_value_replication $table_name
+
+  # Cleanup the volume
+  docker stop $(cat $cidfile)
+  local rm_cmd='/bin/rm -rf /var/lib/pgsql/data/*'
+  docker run $master_args --rm $IMAGE_NAME /bin/sh -c "$rm_cmd"
+  rmdir ${volume_dir}
```
praiskup: Nit, can we do this at one place (looks like code duplication).

hack/run_test
```diff
@@ -350,6 +352,43 @@ function setup_replication_cluster() {
...
+  local table_name="t1"
+
+  local volume_dir
+  volume_dir=`mktemp -d --tmpdir pg-testdata.XXXXX`
+  chmod a+rwx ${volume_dir}
+  master_args="-v ${volume_dir}:/var/lib/pgsql/data:Z"
+
+  # Setup the cluster
+  local slave_num=2
+  setup_replication_cluster $slave_num
```
praiskup: Looks like setup_replication_cluster takes the $slave_num twice, once as variable, once as argument.

hack/run_test
```diff
@@ -265,7 +265,8 @@ function run_slave() {
 
 function run_master() {
   local suffix="$1"; shift
-  docker run $cluster_args -p 5432  \
+  master_args=${master_args-}
+  docker run $cluster_args $master_args -p 5432  \
```
praiskup: I'm still not sure whether we need to '-p 5432'.

hack/run_test
```diff
@@ -308,21 +295,29 @@ function run_replication_test() {
...
-  postgresql_cmd -c "CREATE TABLE t1 (a integer); INSERT INTO t1 VALUES ($value);"
+  local table_name="t1"
+  postgresql_cmd -c "CREATE TABLE $table_name (a integer); INSERT INTO $table_name VALUES ($value);"
 
   # Read value from slaves and check whether it is expected
   local f
   local slave_ip
+  local cidfile
+  local master_ip=$(get_container_ip master-$cid_suffix.cid)
   for f in `find ${CIDFILE_DIR} -type f -name "slave-*"`; do
```
praiskup: Here is probably missing "suffix" in find command.  Btw., I would much rather see the `run_slave` method "appending" the IP of newly created slave into let's say $slave_addresses variable.  Then you can simply iterate like `for i slave in $slave_addresses`.

pkubatrh: Good idea. It would be better to take this one step further though and append the container's ID instead. We only need the IPs now, but we might need something else in the future.
We will still need unique file names for the cid files in order for the containers to be cleaned up properly, so the $suffix will probably need to stay in.

hack/run_test
pkubatrh: setup_replication_cluster takes the local $slave_num variable as an optional argument and sets up a local variable with the same name inside the function, defaulting to 1 if called without the argument.
As both of the variables are local, there should not be any value transfer by variable going on.

praiskup: It then just looks weird (or complicated), because you are overwriting the `slave_num` from `run_master_restart_test` with another local variable (named exactly the same) ...  But the local variable from `run_master_...` is available in called `setup_replication_cluster` function, too.   Also, the local variable is not needed:
https://github.com/praiskup/openshift-postgresql/commit/c1b99705a2

hack/run_test
praiskup: > We only need the IPs now, but we might need something else in the future.

The point was not to docker inspect all the time, but that is just preference.  Container's id is better than doing error-prone find command.

> We will still need unique file names for the cid files in order for the containers to be cleaned up properly, so the $suffix will probably need to stay in.

Right, the suffix is already there.  But this particular find could go away completely.

pkubatrh:
Incorporated @praiskup's suggestions - new functions for the creation and removal of volumes, got rid of some redundant local variables, changed loops to use container IDs instead of calling 'find' on .cid files.

praiskup:
Weird.  On my Fedora 24 box, running 'make test VERSION=9.4' fails very early with:
+ timeout -s 9 --preserve-status 60s docker run --rm -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass centos/postgresql-94-centos7-candidate
/bin/bash: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: Permission denied

Actually, the docker build created unusable image, the source 'centos:centos7' works fine.

praiskup:
[test]

praiskup:
My mistake, more info in #136, I'll continue with the review.

hack/run_test
```diff
@@ -309,7 +328,7 @@ function test_value_replication() {
     CONTAINER_IP=$slave_ip
     for i in $(seq $max_attempts); do
       result="$(postgresql_cmd -At -c "select * from $table_name")"
```
praiskup: I'm not sure the loop here has some effect.  When the value is not there yet, it is
very likely the $table_name table does not exist yet too.  Then, postgresql_cmd fails
and this would fail the whole command, so no other attempt would me made.

Probably `|| :` is missing.

praiskup: s/fail the whole command/exit the whole script/

pkubatrh:
Pushed new changes - ignore psql errors when checking for value replication, master_args variable now local

praiskup:
LGTM.

praiskup:
@pkubatrh looks like this is pretty atomic change, would it be fine to squash the commits?

bparees:
squash merged.

Issue #116, opened 2016-05-26
centos/postgresql-94-centos7 fails to start on OpenShift cluster


csrwng:
Using the template from https://github.com/openshift/rails-ex, the postgresql db pod fails to start with the following error:

```
/usr/share/container-scripts/postgresql/common.sh: line 121: /var/lib/pgsql/passwd: Permission denied
```

csrwng:
The image used by the example comes from the one defined in the openshift image streams file: https://github.com/openshift/origin/blob/master/examples/image-streams/image-streams-centos7.json
@bparees is that the right postgresql image?

bparees:
that's the right image, and i just tried it and it worked fine for me, can you pull the latest centos/postgresql-94-centos7 image down?

bparees:
`oc new-app https://raw.githubusercontent.com/openshift/rails-ex/master/openshift/templates/rails-postgresql.json`

is what i ran.

csrwng:
@bparees I pulled this image fresh into a new VM, but I think I know what's going on. This likely one of those cases where the image will run fine on RHEL/Fedora but not on a non-RedHat OS.

csrwng:
So running on the boot2docker OS, it looks like with an arbitrary uid, even though my gid is set to 0, I cannot write to the home directory where permissions are:
```
bash-4.2$ ls -al
total 12
drwxrwx---  4 postgres root 4096 May 17 14:40 .
```

csrwng:
Actually setting the permissions to 777 on that directory doesn't help either. Maybe has something to do with how the image was built?

bparees:
nothing magical about how that image is built, we aren't even squashing anymore.  can you build it yourself and see if the behavior is any different?

csrwng:
I get the same behavior when built locally. However, if I chmod 777 the home directory right after creating it, then things work. Would that be an acceptable fix?

bparees:
chmod 777 instead of chgrp/chmod g+rwx ?

i'd like to understand why, when you're a member of the root group, you can't write to a directory owned by the root group.

csrwng:
doing additional testing to better understand the problem

praiskup:
Can SELinux be the issue?

bparees:
any results?

csrwng:
will revisit this issue today. I want to get a set a reproducible steps before I ask the wider community for help.

csrwng:
@bparees Latest update... 

[...]

PR #134, opened 2016-09-01
create dir and set permissions in a single layer to work around aufs limitations

bparees:
fixes https://github.com/sclorg/postgresql-container/issues/116

@csrwng ptal

### DIFF
--- a/9.2/Dockerfile
+++ b/9.2/Dockerfile
@@ -23,6 +23,8 @@ LABEL io.k8s.description="PostgreSQL is an advanced Object-Relational database m
 
 EXPOSE 5432
 
+ADD root /
+
 # This image must forever use UID 26 for postgres user so our volumes are
 # safe in the future. This should *never* change, the last test is there
 # to make sure of that.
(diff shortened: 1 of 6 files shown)


hhorak:
Even if it's not clear to me why something like that might make the difference, I understand we need to get rid of fail. If this patch helps, LGTM.

bparees:
@hhorak see https://github.com/sclorg/postgresql-container/issues/116#issuecomment-232845287

csrwng:
@bparees it's still broken for me ... I believe because /var/lib/pgsql gets created in the previous layer

However, this works:
https://github.com/sclorg/postgresql-container/compare/master...csrwng:fix_aufs_perms?expand=1

bparees:
updated, @csrwng ptal again

csrwng:
LGTM
tested all versions on a Mac with Docker for Mac and they all work.

bparees:
thanks @csrwng, will wait for tests to pass and then merge.

PR #137, opened 2016-09-06
Fix permissions for using persistent storage

minimaru:
Postgresql pod is CrashLoopBackOff if using persistent storage.

Please, see postgresql container logs.

    Success. You can now start the database server using:

        postgres -D /var/lib/pgsql/data/userdata
    or
        pg_ctl -D /var/lib/pgsql/data/userdata -l logfile start

    WARNING: enabling "trust" authentication for local connections
    You can change this by editing pg_hba.conf or using the option -A, or
    --auth-local and --auth-host, the next time you run initdb.
    waiting for server to start....FATAL:  data directory "/var/lib/pgsql/data/userdata" has wrong ownership
    HINT:  The server must be started by the user that owns the data directory.
    .... stopped waiting
    pg_ctl: could not start server
    Examine the log output.

### DIFF
--- a/9.2/root/usr/libexec/fix-permissions
+++ b/9.2/root/usr/libexec/fix-permissions
@@ -4,4 +4,4 @@
 find "$1" -exec chown postgres {} \;
 find "$1" -exec chgrp 0 {} \;
 find "$1" -exec chmod g+rw {} \;
-find "$1" -type d -exec chmod g+x {} +
+find "$1" -type d -exec chmod +xw {} +
--- a/9.4/root/usr/libexec/fix-permissions
+++ b/9.4/root/usr/libexec/fix-permissions
@@ -4,4 +4,4 @@
 find "$1" -exec chown postgres {} \;
 find "$1" -exec chgrp 0 {} \;
 find "$1" -exec chmod g+rw {} \;
-find "$1" -type d -exec chmod g+x {} +
+find "$1" -type d -exec chmod +xw {} +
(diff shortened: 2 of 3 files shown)


praiskup:
@minimaru, thanks for the patch.

Please correct me if I'm wrong, but setting write access for everybody by default
is not good.  Also, the mounted storage (and it's permissions) has nothing to do with
`fix-permissions` script which is called only in Dockerfile (thus during `docker build` time).

It looks like you've re-used persistent storage against different container, unfortunately
I'm not sure how this is handled in Openshift.  Can anyone help?

bparees:
yeah this is not the right, or necessary, fix.  the image already works with persistent storage when used properly.  one frequent cause of issues w/ persistent storage in openshift is when you are re-using an existing volume that already has a "userdata" directory in it.  The image wants to create a new userdata directory that it owns, so if there is one already in the volume, that can cause problems.

praiskup:
@bpares, but I can imagine user's want to re-use data directory, what is the expected
workflow then (I guess the UID is always changed with the new container)?

bparees:
@praiskup the uid changes if you have a new pod.  if you have a use case where you're going to delete the pod but want to reuse the data, you should be using the DB tools to export the data and then re-import it.  Otherwise, there's no guarantee who's going to get your persistent storage volume after you're done with it.  it could be assigned to any new pod.

praiskup:
I would expect that app needs to be updated (new app container), while the database data are just migrated (by running some database migration scripts like django migration, or alembic from app container).

The point probably is that one pod **always** survives during the whole life-time of the hosted app, so there is no need to migrate DB to another pod, or restart the whole pod.

minimaru:
I face the problem with not re-using existing NFS volume but "clean" NFS volume.

It appears that this problem is same as fix "BZ - 1282733" of RedHat
errata "[RHSA-2016:1064 - Security Advisory](https://bugzilla.redhat.com/show_bug.cgi?id=1282733)".

Has the fix "BZ - 1282733" already applied to Openshift Origin ?

bparees:
based on the age i would expect so, yes.

PR #138, opened 2016-09-06
build: move common parts out from this project

praiskup:
@hhorak, @pkubatrh, @bparees WDYT?  Please don't merge until we have
appropriate git repository in sclorg namespace.


praiskup:
[test]

praiskup:
The CI needs to be fixed to call 'git submodule update --init' first.  It may be
called unconditionally because it is empty operation if there are no submodules.

praiskup:
@omron93, could we fix the CI?

praiskup:
Hms, I fail to see where the `w` make option comes from in CentOS ci.

Makefile
```diff
@@ -1,7 +1,12 @@
-# Variables are documented in hack/build.sh.
+# Variables are documented in common/build.sh.
 BASE_IMAGE_NAME = postgresql
 VERSIONS = 9.2 9.4 9.5
 OPENSHIFT_NAMESPACES = 9.2
 
 # Include common Makefile code.
-include hack/common.mk
+-include common/common.mk
+git_submodule_warning:
```
bparees: why is it called git_submodule_warning ?

bparees: and why isn't this part of common.mk itself?

Makefile
```diff
@@ -1,7 +1,12 @@
...
+	git submodule update --init
+
+test: git_submodule_warning
+	$(MAKE) check $(MAKEFLAGS)
```
bparees: are you renaming "test" to "check" in common.mk?

bparees:
i'm not a fan of submodule headaches but you guys own the repo.

Makefile
praiskup: nono, the second commit is there just to make it work without fixing CI ... if this gets merged, we should have CI doing the submodule-init explicitly.

praiskup: IOW: please ignore the second commit

Makefile
praiskup: Because until you init the submodule, you don't have common.mk.  That target is invoked only if the 'make' is called and the git submodule is not initialized.

praiskup:
For example, I like git submodules (pretty common way to share code within GNU).
Nowadays each submodule is fixed to concrete hash (until explicitly updated) so I'm not
familiar with particular issues.  Can you mention some, so we can reject this idea rather
sooner than later?

bparees:
I will admit i have used them very little, it just is another thing to manage in your workflow (keeping them up to date, if you need to make a change, making it in the appropriate repo and then reflecting the change back down to your submodules, etc).

the fact that "git clone ; make test" won't work now (and won't tell me why until i run "make", if i'm understanding this correctly) would seem to make the barrier to entry higher for users.

Again, my vote counts minimally at this point, it's an SCL repo, and certainly I like the value of sharing this common code between all these repos.

praiskup:
> I will admit i have used them very little, it just is another thing to manage in your workflow (keeping them up to date, if you need to make a change, making it in the appropriate repo and then reflecting the change back down to your submodules, etc).

Right, the submodule is like library, has it's own API and it should rarely change.  And if we change something, each project still needs to be explicitly updated (with actual proposal).

> the fact that "git clone ; make test" won't work now (and won't tell me why until i run "make", if i'm understanding this correctly) would seem to make the barrier to entry higher for users.

Well, newcomers won't have such issue (those will run 'make' if any).  Old maintainers are probably flexible enough.  But, if we consider this point serious enough, the second patch (with slight modification) can be applied to make the `make test` work.  Originally I thought I could just add the `make check` target as an alias for `make test`, because by habit I always hit `make check`... `make test` is non-standard to some extent.

omron93:
@praiskup  So should I change CI? Your second commit works fine and it is invisible for users.

pkubatrh:
This change makes sense to me and the change to the workflow seems bearable (if I understand it correctly that is). However, I am not that familiar with git submodules and as such am unaware of any issues that might come up after changing to them.

praiskup:
@omron93 I personally dislike the hack in 2nd commit (I wouldn't be proud to be author of such commit :)), it just does non-transparent things -- I would rather see rather change in CI doing automatically really innocent git submodule init.  Feel free to lift up pros/cons.

praiskup:
@omron93 also, don't you have an idea why ci.centos.org CI fails?

omron93:
[test]

omron93:
@praiskup Centos CI machines were updated today morning, so maybe this could cause this problem. This PR was the only one tested today.

omron93:
@praiskup Seems that Centos7 tests goes right. 'make test SKIP_SQUASH=1 UPDATE_BASE=1' resolves in 'make check w -- UPDATE_BASE=1 SKIP_SQUASH=1' and there is no target for 'w'...

praiskup:
@omron93 any idea where the 'w' comes from?

praiskup:
There is some `ssh` wrapper involved, for example.  Neither in `Makefile`, nor in `common.mk` occurs `w` character, so there must be something different?  Dunno. I'm not sure, am I able to have a closer look somehow?

> Personally I don't have a problem with this. However sounds better to me to keep current workflow (make test) with some deprecation warning. And after some time to remove this make target...

The `make test` is still there, no deprecation is expected.  Users just need to initialize submodules, **once** for the whole life of local repo copy.

praiskup:
There is one difference!  You are using `make -C` in CI for CentOS, this causes that `MAKEFLAGS` contain `w` character.  This looks like bug in GNU make, but TBH I'm not really sure.  As this is anyway about the second patch, I'll revert at least for now ... and in the meantime, I'll make sure what is wrong with GNU make.

omron93:
>  I'm not sure, am I able to have a closer look somehow?

Jenkins job builder configuration resides in https://github.com/sclorg/rhscl-container-ci (I haven't pushed RHEL config yet:-().

The most important config for CentOS PR testing is in https://github.com/sclorg/rhscl-container-ci/blob/master/centos-ci-jenkins/yaml/jobs/collection.yaml
Mainly `builders` section (`cico_install' and 'generate_ssh_config` is from https://github.com/sclorg/rhscl-container-ci/tree/master/centos-ci-jenkins/yaml/builders).

Output of the jenkins test contains output of all these commands.

omron93:
@praiskup I will update CI scripts tomorrow...

praiskup:
Thanks!

praiskup:
Turns out this is expected behavior of `$(MAKEFLAGS)` which is not needed in this case.
I've temporary re-enabled the hack.

Makefile
praiskup: I've noticed that the first commit pointed to "ssh-cloned" submodule, that was wrong because not everybody (nobody except for me in case of `praiskup` namespace) is able to clone like that. Moved to https and force pushed.

omron93:
> The make test is still there, no deprecation is expected. Users just need to initialize submodules, once for the whole life of local repo copy.

@praiskup I thought that second commit won't be there. And without it, is there a `make test` somewhere? (I don't understand make so well, but I see only `make check` in shared scripts)

Why you don't like calling `git submodule ...` from Makefile? (to automate it)

omron93:
@praiskup RHSCL CI is updated...

praiskup:
@omron93 I dislike second commit, and yes -> there is still `make test` target in shared common.mk, so once user initializes git submodules `make test` works "forever".
You can see that in the shared scripts directory is also one "hack" commit which should go away ...

praiskup:
Let me simplify things -> as there is no general disagreement with this proposal, I'll finalize it now (by creating the "common" repository within `sclorg` and, as we have fixed CI now, there will be only the one "good" commit, that will be less confusing.

praiskup:
Please have a look, in case you have already cloned within 'common' subdirectory, I suggest you to remove 'common' directory, edit '.git/modules/common/config' so it point's to correct project (https://github.com/sclorg/container-common-scripts) and hit `git submodule update --init`.

omron93:
@praiskup Sorry, there are some problems with RHEL7 jenkins, so I can't update jobs there. So that is why RHEL7 tests are failing...

praiskup:
NP, maybe we can gather more opinions until RHEL7 CI get's fixed.  @hhorak?

praiskup:
[test]

omron93:
RHEL7 test should be updated...

[test]

praiskup:
@omron93 thanks!  One suggestion, I've missed the `git submodule update --init` action in bulid log, would it be possible to trace (`sh -x`) the shell script performing git repository initialization?

omron93:
@praiskup `set -x` should be set in all jobs.

[test]

praiskup:
I was told that, even though it is not visible e.g. in this PR, the OpenShift CI is still running in background.  @bparees would that be OK to add `git submodule update --init` into openshift CI, too?

bparees:
@praiskup this is an example of why i didn't want to go down this path.  can't you just add the submodule update to the make logic so it happens automatically?

praiskup:
@bparees that would be really ugly :(. I was thinking about it, but I see no trivial way.  The previous "hack" solution was be to rename 'test' target into 'check' for example (in common.mk), and implement 'test' directly in per-container Makefile, and call 'check' after automatic submodule initialization.

What I'm asking for is something similar to https://github.com/sclorg/rhscl-container-ci/pull/6/commits/7bd43fd65ca1e92b1f0aa within openshift's CI, that is trivial and zero risk (I can submit PR if you show me where).  If this is not acceptable, then the hack is needed, at least until openshift's ci is active.

 into openshift CI then we probably would have to the non-preferred way.  so 
(which should stay as small as possible, ideally it should be rather configuration file only).

bparees:
@praiskup i think the changes need to go here:

https://github.com/openshift/vagrant-openshift/blob/master/lib/vagrant-openshift/command/test_origin_image.rb#L120
https://github.com/openshift/vagrant-openshift/blob/master/lib/vagrant-openshift/action/push_openshift_images.rb#L101-L103

hhorak:
I think this change shouldn't cause big troubles for end-users who want to build the containers themselves, provided we document necessary steps properly. That should be part of this PR, which seems to be not done, README.md does not include `git submodule update --init` step:
```
$ git clone https://github.com/openshift/postgresql.git
$ cd postgresql
$ make build VERSION=9.2
```
same for tests:
```
$ cd postgresql
$ make test VERSION=9.2
```

omron93:
[test]

praiskup:
[test]

praiskup:
[test]

omron93:
@praiskup coreservices-jenkins seems to be down now :-(

bparees:
[test]

omron93:
@bparees Tests are passing, so if you are fine with this change, I'll merge it.

@praiskup Could you please report an issue to remember to remove `.PHONY` call of submodules init (and sumarize the reasons)? How long to have it in Makefile to let users to notice the change?

bparees:
@omron93 it's fine with me.

praiskup:
@omron93  wrote:
> @praiskup Could you please report an issue to remember to remove .PHONY call of submodules init (and sumarize the reasons)?

One of the reasons is that we need to have this hack distributed everywhere .. (otherwise to reuse common.mk is just single include, so almost 100% code reusability).  And the need for this hack is at least doubtful (mostly it is because of robots in the wild and docs).

More for reasons: having implicit side-effects (this is reaching internet!) in make **targets** is ugly, but it is IMO self explaining in hack itself.  I'm not sure where to submit a bug, probably against all consumers ...?  Feel free, I don't think we'll forget about it.

> How long to have it in Makefile to let users to notice the change?

I'm not sure I got the question ... I'm afraid getting rid of the hack is not going to be easy (the reasons would be the same as now, IOW why we can not push without it..).

omron93:
> More for reasons: having implicit side-effects (this is reaching internet!) in make targets is ugly, but it is IMO self explaining in hack itself.

I totaly agree with you.

> I'm not sure I got the question ... I'm afraid getting rid of the hack is not going to be easy (the reasons would be the same as now, IOW why we can not push without it..).

@praiskup I thought that you added the hack because you wanted. (I thought we aggreed offline that we want this change even if users will have to change their git commands)

But I see a little "advantage" in a temporary hack. Users could see the change in README (so their git workflow should be changed) and they have some time to notice the README change (during this time using `make` works the same). But not sure if it could cause more confusion? What is your opinion?

So I vote for merge even without this hack.

praiskup:
I more or less agree but one would have to take care of the original jenkins CI as @bparees mentioned above;  which I'm not sure it is worth it.  @omron93 would you be able to have a look at relevant PRs and deal with potential issues?  Or could we consider the CI is down/obsoleted?

omron93:
>I more or less agree but one would have to take care of the original jenkins CI as @bparees mentioned above; which I'm not sure it is worth it. 

Original jenkins CI is going to be shut down in near future, so it does not worth to change it. So if we decide to not to use "make hack" I will talk to Ben to shut down it quicker.

>would you be able to have a look at relevant PRs and deal with potential issues?

@praiskup What relevant PRs?

praiskup:
Ok, so what would be the next step?  I guess I should drop the hack?
> What relevant PRs?
https://github.com/sclorg/postgresql-container/pull/138#issuecomment-246977051

bparees:
@praiskup @omron93 

i've shutdown our jenkins CI, so you should be free to do what you need to do.

omron93:
> Ok, so what would be the next step? I guess I should drop the hack?

Please drop the hack and add warning meesage again (that submodules have to be initialized).

omron93:
@praiskup Thanks for updates and this PR.

I hope last step, somewhere it was required to squash commits in PR (@bparees is it documented somewhere?), please do so...

praiskup:
I don't think that is general rule, but In this case squashing makes sense to me.

bparees:
> I hope last step, somewhere it was required to squash commits in PR (@bparees is it documented somewhere?)

squashing commits for PRs (especially when the commits are all related) is a best practice we follow in the openshift org, since you guys own these repos now you're welcome to do what you want.  There's no technical/functional requirement to do it.

praiskup:
[test]

omron93:
[test]

omron93:
Tests are passing + README for users will be updated, so I don't expect unexpected problems. So merging.

@praiskup Thaks for making this PR.
@bparees Thanks for cooperation.

Issue #140, opened 2016-09-07
Let's make the test-cases completely independent


praiskup:
So we can selectively run particular tests.

praiskup:
And run the tests in parallel, ideally through `make` which is proven to handle parallelism correctly.

pkubatrh:
I will try taking a look at this.
Running test cases in parallel will be problematic right now with the master_restart test relying on a specific IP of the master to synchronize properly (which might get 'stolen' by another running test case) so that will definitely need to be fixed before making the change.

praiskup:
Isn't it possible to start docker containers so they see only explicit subset of other containers?

Let's make the test-cases as much independent as possible first.  The parallel approach is something which deserves another issue.

pkubatrh:
>Isn't it possible to start docker containers so they see only explicit subset of other containers?

Right, creating a docker network for a single test case should resolve the issue. Will try it out later and focus on test case independency for now.

PR #141, opened 2016-09-07
simplify working with temporary volume directories

praiskup:
Instead of special-case the master-slave tests by (a) stopping the
container manually, (b) removing cidfile and (c) remove data
directory, from now the volume cleanup is "just" planned to be
executed after regular container cleanup in 'cleanup()'.


praiskup:
@pkubatrh, can you please have a look whether you agree with this?

praiskup:
[test]

praiskup:
Also, please - don't squash if you plan to merge this, thanks.

hack/run_test
```diff
@@ -29,18 +47,21 @@ function cleanup() {
...
-  local volume_dir="$1"; shift
-  local cid="$1"; shift
-  local volume_args="-v ${volume_dir}:/var/lib/pgsql/data:Z"
-
-  docker stop $cid
-  local rm_cmd='/bin/rm -rf /var/lib/pgsql/data/*'
-  docker run $volume_args --rm $IMAGE_NAME /bin/sh -c "$rm_cmd"
-  rmdir ${volume_dir}
+function cleanup_volume_dir () {
+  test ! -d "$volume_dir" && : "WARN: cleaned $volume_dir for some reason" && return 0
```
pkubatrh: This should be testing $1 instead of $volume_dir

praiskup: Good catch.

pkubatrh:
I'm not sure if changing the default field seperator to ':' is really needed when looping over temporary filepaths. But that is just a nit, otherwise LGTM.

praiskup:
@pkubatrh, truth - but promoting that to library code one day, it is worth having it general for any path.  
Btw,. that is inspired by http://git.savannah.gnu.org/cgit/gnulib.git/commit/?id=55c6f22b8043 so it is to some extent already tested code (though modified a bit as we don't have to be "that" portable here).

praiskup:
Argh, that code in gnulib is more powerful (see the `%dir%` part), but anyway.  It is not needed yet here.

pkubatrh:
> @pkubatrh, truth - but promoting that to library code one day, it is worth having it general for any path.

Right, if we want to use it for other path checks in the future I don't see a reason not to have it now.

pkubatrh:
@praiskup Just and idea: would it be possible to expand this from cleaning up volumes only to cleaning up after containers in general? This would be nice to have for #140

pkubatrh:
Although probably not needed, since every test case can have their own temporary directory to store their stuff in. Nevermind.

omron93:
+1 for adding into common testing library. It would be used also in MongoDB container.

praiskup:
Note that this requires rebase against fixed container-common-scripts first.

praiskup:
Rebased, PTAL.

praiskup:
PTAL

test/run_test
```diff
@@ -46,18 +51,22 @@ function cleanup() {
...
+cleanup_volume_dir ()
+{
+  test ! -d "$1" && : "WARN: cleaned $1 for some reason" && return 0
+  # When we run this test script as non-root (we should?), the PostgreSQL server
+  # within container is still run under 'postgres' user.  It means that, taking
+  # into account 0077 umask of PostgreSQL server, we are unable to remove files
+  # created by server.  That's why we need to let docker escalate the privileges
+  # again.
+  local datadir=/var/lib/pgsql/data
+  docker run -v "$1:$datadir:Z" --rm "$IMAGE_NAME" /bin/sh -c "/bin/rm -rf $datadir/userdata"
```
omron93: I thought that container is also tested with "random" user (`"-u 12345"`). I tried this manually and in that case this does not work for me. So I had to run cleaning with `-u 0`.

But tests are passing, so maybe I'm doing something wrong. @praiskup Or random uid is not tested with mounted volume?

test/run_test
```diff
@@ -127,13 +136,15 @@ function create_container() {
...
 }
 
-function create_volume() {
-  local volume_dir
+
+create_volume_dir ()
+{
   volume_dir=`mktemp -d --tmpdir pg-testdata.XXXXX`
   setfacl -m u:26:-wx "$volume_dir"
   echo "$volume_dir"
```
omron93: What does this echo do? When `create_volume_dir` is called output is not captured...

test/run_test
```diff
@@ -29,6 +33,7 @@ test -n "${OS-}" || false 'make sure $OS is defined'
 
 CIDFILE_DIR=$(mktemp --suffix=postgresql_test_cidfiles -d)
 
+volumes_to_clean=
```
omron93: This variable is not modified nowhere

test/run_test
```diff
@@ -385,7 +396,7 @@ function run_master_restart_test() {
   local master_ip=
   local slave_cids=
 
-  local volume_dir=$(create_volume)
+  create_volume_dir
```
pkubatrh: I may be missing something but should not this be added to `volumes_to_clean`?

praiskup:
Drat, what a OCD mess (rebase).  That's a day ... I'll fix this.

> I thought that container is also tested with "random" user ("-u 12345"). I tried this manually and in that case this does not work for me. So I had to run cleaning with -u 0.

I'm bit afraid of disasters caused by giving the container (== the `rm -rf` command) root privileges (e.g. if we mistakenly bind-mounted wrong directory ....).

praiskup:
Thanks for the review, @pkubatrh and @omron93.  I hope addressed the remarks, let's see the CI output.

praiskup:
After an hour, trying once more [test].
```
localedef -f UTF-8 -i en_US en_US.UTF-8 &&     test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)" &&     mkdir -p /var/lib/pgsql/data &&     /usr/libexec/fix-permissions /var/lib/pgsql &&     /usr/libexec/fix-permissions /var/run/postgresql
13:16:56  ---> Running in 8bad5673577e
13:16:57 Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
14:16:01 Build was aborted
14:16:01 Aborted by Pavel
```

test/run_test
omron93: Above mentioned problem is fixed now.
---------------------------------------------
Technically this is not needed - `ct_path_append` creates a variable if it doesn't exist...

omron93:
> I'm bit afraid of disasters caused by giving the container (== the rm -rf command) root privileges (e.g. if we mistakenly bind-mounted wrong directory ....).

OK. That's probably true.


LGTM. @pkubatrh If you don't have any other suggestions please merge this.

pkubatrh:
lgtm as well, thanks! Merging

PR #142, opened 2016-09-15
Make test cases independent

pkubatrh:
First try on independent tests. Related to #140 

This PR moves each of the existing test cases into their own file inside `hack/tests` directory and provides a Makefile to run them.

Open questions:
1. What do we do with leftover `$VERSION/test/run` soft links?
2. Do we leave the new test files in `hack/tests` directory or somewhere else?
3. Divide `general` tests? How?

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -92,8 +92,7 @@ for dir in ${dirs}; do
   fi
 
   if [[ -v TEST_MODE ]]; then
-    VERSION=$dir IMAGE_NAME=${IMAGE_NAME} test/run
-
+    make -C ../ ${TEST_CASE:-runtests} VERSION=$dir IMAGE_NAME=${IMAGE_NAME}
     if [[ $? -eq 0 ]] && [[ "${TAG_ON_SUCCESS}" == "true" ]]; then
       echo "-> Re-tagging ${IMAGE_NAME} image to ${IMAGE_NAME%"-candidate"}"
       docker tag -f $IMAGE_NAME ${IMAGE_NAME%"-candidate"}
--- a/hack/run_test
+++ b/hack/run_test
@@ -9,9 +9,19 @@
 set -exo nounset
 shopt -s nullglob
 
+TEST_LIST="\
+run_container_creation_tests
+run_general_tests
+run_change_password_test
+run_replication_test
+run_master_restart_test"
+
+test $# -eq 1 -a ${1-} == --list && echo "$TEST_LIST" && exit 0
 test -n "${IMAGE_NAME-}" || false 'make sure $IMAGE_NAME is defined'
 test -n "${VERSION-}" || false 'make sure $VERSION is defined'
+
 CIDFILE_DIR=$(mktemp --suffix=postgresql_test_cidfiles -d)
+
 
 function cleanup() {
   for cidfile in $CIDFILE_DIR/* ; do
@@ -357,6 +367,10 @@ function setup_replication_cluster() {
 }
 
 function run_master_restart_test() {
+  local DB=postgres
+  local PGUSER=master
+  local PASS=master
+
   echo "Testing failed master restart"
   local cluster_args="-e POSTGRESQL_ADMIN_PASSWORD=pass -e POSTGRESQL_MASTER_USER=$PGUSER -e POSTGRESQL_MASTER_PASSWORD=$PASS"
   local cid_suffix="mrestart"
@@ -394,6 +408,10 @@ function run_master_restart_test() {
 }
 
 function run_replication_test() {
+  local DB=postgres
+  local PGUSER=master
+  local PASS=master
+
   echo "Testing master-slave replication"
   local cluster_args="-e POSTGRESQL_ADMIN_PASSWORD=pass -e POSTGRESQL_MASTER_USER=$PGUSER -e POSTGRESQL_MASTER_PASSWORD=$PASS"
   local cid_suffix="basic"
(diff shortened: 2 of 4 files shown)


pkubatrh:
CI is failing because of nonexisting soft link targets. See open question 1.

praiskup:
Could we have everything in one flie and still make the test-cases independent?  Then you can have something like '--keywords "replication"'  or '--tests "IDA"' to select them.

praiskup:
Things that are general enough to cut them out to "library" can be solved as separate issue, similarly to fixing the existing Makefile to support parallelism.

pkubatrh:
All test cases now contained in one `run_test` file and can be run using `run_test test_case`. Made necessary tweaks to the Makefile.

praiskup:
Thats much better, thanks!  

Some notes: Could the `./run_tests` script run all the tests by default?  Then you don't have to define the list in `Makefile`.  What I would like to avoid is to do edits in `Makefile`.

hack/run_test
```diff
@@ -480,21 +490,21 @@ $volume_options
...
-PGUSER=user1 PASS=pass1 ADMIN_PASS=r00t run_tests admin
-DB=postgres ADMIN_PASS=r00t run_tests only_admin
-# Test with arbitrary uid for the container
-DOCKER_ARGS="-u 12345" PGUSER=user2 PASS=pass run_tests no_admin_altuid
-DOCKER_ARGS="-u 12345" PGUSER=user3 PASS=pass1 ADMIN_PASS=r00t run_tests admin_altuid
-DB=postgres DOCKER_ARGS="-u 12345" ADMIN_PASS=rOOt run_tests only_admin_altuid
-run_change_password_test
-DB=postgres PGUSER=master PASS=master run_replication_test
-DB=postgres PGUSER=master PASS=master run_master_restart_test
+test "$(type -t $1)" == "function" || false "$1 is not a test case."
```
praiskup: The `$1` should be in quotes, the `function` doesn't have to be.  I think, however that' doing
```
: "running $1"
"$1"
```
could be enough.

pkubatrh:
Well the list in the Makefile is there to just control whether there was an actual existing test case provided via the TEST_CASE. Which I was (basically) doing anyway in `run_test` now that I think about it. I will just remove it from the Makefile and provide a default `run_all_tests` function in the script.

hack/build.sh
```diff
@@ -92,12 +92,13 @@ for dir in ${dirs}; do
   fi
 
   if [[ -v TEST_MODE ]]; then
-    VERSION=$dir IMAGE_NAME=${IMAGE_NAME} test/run
-
+    TEST_DIR=../hack
+    pushd $TEST_DIR && make $TEST_CASE VERSION=$dir IMAGE_NAME=${IMAGE_NAME}
```
praiskup: (shrug), I'm not sure this is needed.

pkubatrh:
Although moving the list into `run_test` will make the default not able to be run in parallel in future via Makefile, if I understand it correctly

praiskup:
It should be possible.  There's one additional thing needed  -- to be able to get the list of tests somehow;  be that from separate file or e.g by `./run_test --list-tests`. 

Another thing is that you should be able to run testsuite without Makefile easily.

hack/build.sh
pkubatrh: Why not? Since we need a way to actually run `make` on the tests, we can't go through the soft link. Well, we could just run the test directly but that would defeat the purpose of the Makefile. Or should I just postpone introduction of the Makefile to a later patch?

praiskup: Right, I was confused.  Can we use `make -C` here?

praiskup: Actually, it is not really pretty calling the make here (but that is rather historic issue than problem with potential fix).  The `test/run` script should be run from `common.mk` somehow (if someone wanted to make that pretty).

pkubatrh: >The test/run script should be run from common.mk somehow (if someone wanted to make that pretty).

I agree this would be the best way to do it. Although I'm not sure how to do that without doing a bigger overhaul of `build.sh` and `common.mk`. Probably best to leave it for now and rewrite it as a part of a separate issue.

praiskup: Yup, I can help too once we are done with this PR.

pkubatrh:
Moved the list of test cases from the Makefile into the actual testing script while providing a way of getting it via `run_test --list`.
When `run_test` is called without any arguments, it starts all available test cases by default.
Now calling `make -C` instead of `pushd && popd`.

hack/run_test
```diff
@@ -9,10 +9,20 @@
 set -exo nounset
 shopt -s nullglob
 
+TEST_LIST="\
+run_container_creation_tests \
+run_general_tests \
+run_change_password_test \
+run_replication_test \
+run_master_restart_test"
```
praiskup: Please remove the trailing back-slashes, if possible.

hack/run_test
```diff
@@ -9,10 +9,20 @@
...
+
+test $# -eq 1 -a ${1-} == --list && echo $TEST_LIST && exit 0
 test -n "${IMAGE_NAME-}" || false 'make sure $IMAGE_NAME is defined'
```
praiskup: use 'echo "$TEST_LIST"' with the note above, so people can easily grep the output.

hack/Makefile
```diff
@@ -0,0 +1,11 @@
+tests = $(shell ./run_test --list 2>/dev/null)
+
```
praiskup: That's it.

hack/Makefile
```diff
@@ -0,0 +1,11 @@
...
+.PHONY: all
+
+all: $(tests)
+
+%:
+	@if [[ "$(tests)" =~ "$@" ]]; then \
+		echo Running test: $@; \
+		VERSION=$(VERSION) IMAGE_NAME=$(IMAGE_NAME) ./run_test $@; \
+	fi;
```
praiskup: I stlil don't think this needs to be in `Makefile` or `hack/Makefile`.  can not this be directly in common.mk?

praiskup: What is the check '=~' for?  It guards against misspelled test names?  Also, it is always best to write in POSIX shell if you do Makefiles.

pkubatrh: >can not this be directly in common.mk?
Will try to incorporate it.

>What is the check '=~' for? It guards against misspelled test names?
Yes, although changing it to multiple targets for the rule makes more sense so I will do that.

pkubatrh:
Moved `hack/Makefile` content into `hack/common.mk`.
Made `hack/run_test --list` output more grepable.

hack/run_test
```diff
@@ -480,21 +498,28 @@ $volume_options
...
-DOCKER_ARGS="-u 12345" PGUSER=user2 PASS=pass run_tests no_admin_altuid
-DOCKER_ARGS="-u 12345" PGUSER=user3 PASS=pass1 ADMIN_PASS=r00t run_tests admin_altuid
-DB=postgres DOCKER_ARGS="-u 12345" ADMIN_PASS=rOOt run_tests only_admin_altuid
-run_change_password_test
-DB=postgres PGUSER=master PASS=master run_replication_test
-DB=postgres PGUSER=master PASS=master run_master_restart_test
+# Run all tests when the test case is not explicitly provided
+test $# -ge 1 || { run_all_tests && exit 0; }
+# Run the chosen test.
+if [[ ! "$TEST_LIST" =~ "$1" ]]; then false "$1 is not a test case."; else $1; fi
```
praiskup: I really dislike the last line.   The `=~` operator is not needed, and is used not properly.  The second argument of `=~` should be unquoted (there are buggy bashes across our products for that), well, yes - you can use quotes, but for different purpose probably.  If you really want that `=~`, then please use correct regexp.

I see nothing wrong on running as it is 100% clear if it fails:
: "running test case $1"
"$1"

praiskup: By correct regexp I mean: 'run' string shouldn't match everything.

praiskup:
Except for the last note, It looks like a pretty good start.  Note that the only reason the testsuite is not parallel yet (with `make -j`) is the implicit make in make guard ... that shouldn't be problem.

hack/run_test
pkubatrh: >By correct regexp I mean: 'run' string shouldn't match everything.

Right, thanks for pointing this out. Will probably just get rid of it completely.

pkubatrh:
Removed checking for a correct test name before running the test.

praiskup:
Looks good, I think.  Note that some small edits will be needed in other containers, too.. but I'm not sure we want to sync before https://github.com/sclorg/postgresql-container/pull/138 .

hack/run_test
```diff
@@ -480,21 +498,30 @@ $volume_options
...
-DOCKER_ARGS="-u 12345" PGUSER=user2 PASS=pass run_tests no_admin_altuid
-DOCKER_ARGS="-u 12345" PGUSER=user3 PASS=pass1 ADMIN_PASS=r00t run_tests admin_altuid
-DB=postgres DOCKER_ARGS="-u 12345" ADMIN_PASS=rOOt run_tests only_admin_altuid
-run_change_password_test
-DB=postgres PGUSER=master PASS=master run_replication_test
-DB=postgres PGUSER=master PASS=master run_master_restart_test
+# Run all tests when the test case is not explicitly provided
+test $# -ge 1 || { run_all_tests && exit 0; }
+# Run the chosen test.
+: "Running test $1"
```
praiskup: I missed this `test $# -ge 1` before .. doesn't that mean that we should rather do:
```
for test in "$@"
do
   : "running test $test"
  "$test"
done
```

praiskup: or `TEST_LIST=$@ run_all_tests`

pkubatrh: I put `-ge 1` there as not to run every test in case multiple arguments were provided (as this is not used by make). However, looking at it now, it would be better to just try running all of the arguments like you suggest since it can be called this way without using `make`. Thanks.

pkubatrh:
Changed the script to use `run_all_tests` function to run test cases provided as arguments.

praiskup:
Looks good to me.

hhorak:
LGTM.

PR #144, opened 2016-09-23
Setup its own Docker network for 'master_restart' test case

pkubatrh:
As the `master_restart` test case's success depends on getting the same IP for the restarted master node, this PR adds the possibility for a test case to create its nodes on a Docker network unrelated to the default one (with its own IP range).

### DIFF
--- a/hack/run_test
+++ b/hack/run_test
@@ -22,7 +22,6 @@ test -n "${VERSION-}" || false 'make sure $VERSION is defined'
 
 CIDFILE_DIR=$(mktemp --suffix=postgresql_test_cidfiles -d)
 
-
 function cleanup() {
   for cidfile in $CIDFILE_DIR/* ; do
     CONTAINER=$(cat $cidfile)
@@ -38,6 +37,7 @@ function cleanup() {
     rm $cidfile
     echo "Done."
   done
+  test ! -z "${DOCKER_NETS-}" && docker network rm $DOCKER_NETS
   rmdir $CIDFILE_DIR
 }
 trap cleanup EXIT
@@ -60,16 +60,18 @@ function get_cid() {
 
 function get_container_ip() {
   local id="$1" ; shift
-  docker inspect --format='{{.NetworkSettings.IPAddress}}' $(get_cid "$id")
+  local net=${network:-bridge}
+  docker inspect --format="{{.NetworkSettings.Networks.$net.IPAddress}}" $(get_cid "$id")
 }
 
 function get_ip_from_cid() {
   local cid="$1"; shift
-  docker inspect --format='{{.NetworkSettings.IPAddress}}' $cid
+  local net=${network:-bridge}
+  docker inspect --format="{{.NetworkSettings.Networks.$net.IPAddress}}" $cid
 }
 
 function postgresql_cmd() {
-  docker run --rm -e PGPASSWORD="${PASS}" $IMAGE_NAME psql postgresql://$PGUSER@$CONTAINER_IP:5432/"${DB-db}" "$@"
+  docker run --rm ${docker_args:-} -e PGPASSWORD="${PASS}" $IMAGE_NAME psql postgresql://$PGUSER@$CONTAINER_IP:5432/"${DB-db}" "$@"
 }
 
 function test_connection() {
@@ -354,6 +356,8 @@ function test_value_replication() {
 }
 
 function setup_replication_cluster() {
+  test ! -z ${network-} && docker network create --driver bridge $network
+
   # Run the PostgreSQL master
   run_master $cid_suffix
   master_ip=$(get_container_ip master-$cid_suffix.cid)
@@ -372,14 +376,18 @@ function run_master_restart_test() {
   local PASS=master
 
   echo "Testing failed master restart"
-  local cluster_args="-e POSTGRESQL_ADMIN_PASSWORD=pass -e POSTGRESQL_MASTER_USER=$PGUSER -e POSTGRESQL_MASTER_PASSWORD=$PASS"
   local cid_suffix="mrestart"
+  local network=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
+  local docker_args="--net=$network"
+  local cluster_args="$docker_args -e POSTGRESQL_ADMIN_PASSWORD=pass -e POSTGRESQL_MASTER_USER=$PGUSER -e POSTGRESQL_MASTER_PASSWORD=$PASS"
   local table_name="t1"
   local master_ip=
   local slave_cids=
 
   local volume_dir=$(create_volume)
   local master_args="-v ${volume_dir}:/var/lib/pgsql/data:Z"
+
+  DOCKER_NETS="${DOCKER_NETS:-} $network"
 
   # Setup the cluster
   slave_num=2 setup_replication_cluster
@@ -396,6 +404,7 @@ function run_master_restart_test() {
 
   run_master $cid_suffix
   CONTAINER_IP=$(get_container_ip master-$cid_suffix.cid)
+  echo "New master is: $CONTAINER_IP"
 
   # Check if the new master sees existing slaves
   test_slave_visibility


omron93:
@rhcarvalho @hhorak @praiskup Thoughts?

This is necessary to fix our Jenkins RHEL testing.

hack/run_test
```diff
@@ -372,15 +376,19 @@ function run_master_restart_test() {
   local PASS=master
 
   echo "Testing failed master restart"
-  local cluster_args="-e POSTGRESQL_ADMIN_PASSWORD=pass -e POSTGRESQL_MASTER_USER=$PGUSER -e POSTGRESQL_MASTER_PASSWORD=$PASS"
   local cid_suffix="mrestart"
+  local network=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
```
rhcarvalho: ~~This is tricky, in this context we're shadowing the global `network` with some random name.~~

Oh no, I see this is the same case as `docker_args`: declared `local` but accessed as a global from functions called from `run_master_restart_test`...

Might be just me, but I find that very surprising and confusing.

Possibly we already do that same here with other "local" variables?

hack/run_test
```diff
@@ -38,6 +37,7 @@ function cleanup() {
     rm $cidfile
     echo "Done."
   done
+  test ! -z "${DOCKER_NETS-}" && docker network rm $DOCKER_NETS
```
rhcarvalho: I suggest avoiding the double negation and the clever short circuiting:

```bash
if [[ -n "${DOCKER_NETS-}" ]]; then
  docker network rm "${DOCKER_NETS}"
fi
```

hack/run_test
pkubatrh: The thing with bash local variables is that their scope extends into all functions invoked in the function block they were declared in.
So the network variable is not really global, just globally accessible inside a specific test case. That is also why I have chosen to use lower case for its name, to avoid confusion (which happened anyway). Similarly for docker_args.

hack/run_test
```diff
@@ -60,16 +60,18 @@ function get_cid() {
 
 function get_container_ip() {
   local id="$1" ; shift
-  docker inspect --format='{{.NetworkSettings.IPAddress}}' $(get_cid "$id")
+  local net=${network:-bridge}
+  docker inspect --format="{{.NetworkSettings.Networks.$net.IPAddress}}" $(get_cid "$id")
```
pkubatrh: Agreed

hack/run_test
```diff
@@ -60,16 +60,18 @@ function get_cid() {
...
 }
 
 function get_ip_from_cid() {
   local cid="$1"; shift
-  docker inspect --format='{{.NetworkSettings.IPAddress}}' $cid
+  local net=${network:-bridge}
```
pkubatrh: >Are there other calls to get_ip_from_cid in which network is undefined and becomes "bridge"?

Yes, currently the `run_replication_test` test case is not using its own network (as it does not need one) while also calling `get_ip_from_cid`.
Also once `get_container_ip` and `get_ip_from_cid` get de-duplicated more test cases will be using `network` undefined.

hack/run_test
praiskup: Yeah, there are two reasons why I dislike `local` variables. First is it confuses readers (mostly because that is rarely used feature in the wild AFAIK) and the second is code portability (but that is not our issue here).

praiskup:
I'll have a look at the code too, but does this all match reality?  Is OpenShit environment able to guarantee restarted master is going to get the same IP?

hack/run_test
praiskup: +1 for `s/! -z/-n/`, 3 lines are not needed

hack/run_test
praiskup: I am probably a bit non-familiar with docker, but for me, I would enjoy reading a few docs words in-code why such complicated coretutils work needs to be done (i.e. why we can't just use cid_suffix).

pkubatrh:
>Is OpenShift environment able to guarantee restarted master is going to get the same IP?

That is one thing I was also wondering about. I have been using IPs to identify containers in the first master-slave replication test case for sake of simplicity but it should pose no problem to switch to any other means of identification (host names would be my first guess).

hack/run_test
praiskup: Probably some '*sum' utility could help here, but IMHO it would be fine to have some `gen_unique_id` function like:
```
get_unique_id network # Set's $network variable
```

pkubatrh: >i.e. why we can't just use cid_suffix

My thoughts behind this was running the same test case for different versions of postgresql in parallel on a single machine (which I think is what the RHEL CI is doing?). So we need something more random than the test case's suffix ID to identify each private network.

Will add a bit of commentary about this to the of code.

praiskup:
@pkubatrh, unfortunately, hostnames are IMHO "hacked" by /etc/hosts within docker.  Or is there some DNS mechanism?

hack/run_test
praiskup: > running the same test case for different versions of postgresql in parallel 

Good point.  How that is possibe that the tests are run in parallel now in RHEL CI?  The makefile was not ready for that yet, when I was looking at it last time.

praiskup:
And also I have a question -- is it **guaranteed** to  to receive the same IP if we have separate network?  It is unlikely in case of regular DHCP processes (dhcp usually respects some pattern of clients and allocates IP addresses "wisely"), e.g.

hack/run_test
pkubatrh: > How that is possibe that the tests are run in parallel now in RHEL CI?

Well I would think it shouldn't be a problem to run make test VERSION=9.x for each of the versions in parallel even with the current state of the Makefile. But I don't have access to the RHEL CI so I can't check if that is really what is going on.

pkubatrh:
>unfortunately, hostnames are IMHO "hacked" by /etc/hosts within docker. Or is there some DNS mechanism? 

That is true. I was thinking about just editing /etc/hosts on the slaves when the IP of the master changes without involving any additional DNS mechanisms.

>is it guaranteed to to receive the same IP if we have separate network?

It seems to allocate the first available IP address in the address space. But I would have to look into a it a little bit more to say for certain whether or not it is guaranteed.

praiskup:
I doubt something like this is expected to be done by OpenShift too ... so I don't think it is worth doing this in test-suite.  @bparees, what happens with IP if PG master container dies and new container starts?

praiskup:
Seems like there is `--ip` parameter for docker-run, has this been already considered?

pkubatrh:
>Seems like there is --ip parameter for docker-run, has this been already considered?

Thanks for the hint, this might actually be the guarantee that was missing. This parameter still needs a private network set up for it to work.

pkubatrh:
>This parameter (--ip) still needs a private network set up for it to work.

Actually it also needs a subnet explicitly set up by the user, so we would need to specify a different subnet for each private network we will eventually use and not rely on subnets distributed by Docker. I am not really sure if that is the best way to go.

rhcarvalho:
@pkubatrh @praiskup FWIW, these tests are not run in an OpenShift environment.

There are separate tests in https://github.com/openshift/origin/tree/master/test/extended/images

rhcarvalho:
Forgive me, I don't remember how the tests are implemented.

If the problem we are trying to solve is being able to run tests in parallel, should we be aiming at removing global state?

Wouldn't be running the tests in containers with arbitrary names the way to go? 

Why do we care about specific IPs/network topology?

praiskup:
> If the problem we are trying to solve is being able to run tests in parallel, should we be aiming at removing global state?

IMO Yes.   This is something any programmer on any project should pay attention to. (though trying to avoid using global variables _in shell_ is IMO not a good way towards this).

> Wouldn't be running the tests in containers with arbitrary names the way to go?

If there are conflicts, we should make the container names unique, yes.  I'm not sure the testsuite is ready for parallel run ATM, so I am considering running parallel tests in RHEL CI a bug ATM.

> Why do we care about specific IPs/network topology?

Master/slave communication is now via IPv4 addresses.

omron93:
@pkubatrh @praiskup Any progress in this?

What about temporarily disable replication tests? (`run_replication_test` function)

bparees:
>  @bparees, what happens with IP if PG master container dies and new container starts?

@praiskup it should keep the IP.  pod IPs are static, only if the pod is deleted should it get a new IP.

praiskup:
@bparees, and are you using the `--ip` parameter?  Or how OpenShift guarantees the IP is static?
The thing is that we should test it similarly to what actually happens in OpenShift.

bparees:
@praiskup openshift(kubernetes) assigns an ip to a pod when it is created, so the containers in that pod will always have that ip until/unless the pod is deleted.  Deletion is different from stopping/moving a pod to a new node.

pkubatrh:
Unfortunately this woudl not work with podman now, since podman lacks a lot of the networking capabilities needed by this PR. So closing for now. We may revisit later once/if podman is able to work with custom configured networks.

Issue #101, opened 2016-02-27
set passwords for a set of accounts


ibotty:
It would be great to create roles and set the password for many accounts when starting the pod.

For many databases you have fine-grained permissions. It would be great to manage credentials in one place (one secret per database role), instead of having to set it once in the database, and once in the application.

I propose, that instead of hardcoding /etc/credentials/pgmaster, /etc/credentials/pguser, /etc/credentials/pgadmin, we use every directory within /etc/credentials, create the role if it does not exist yet, and set the password.

If/when a solution to downward-api for secrets (kubernetes/kubernetes#18372) lands, we could/should set role passwords automatically. That still leaves the question on what to do with additions. That can only be fixed properly when one can mount new volumes (or at least secrets) into running kubernetes pods, I suppose.

ibotty:
Well, now that secrets get updated automatically, I'd like to get a run at this issue.

My plan of action is the following. I do welcome comments on the approach.

 * change start script to also set credentials from mounted secrets, with secrets having preference over environment variables,
 * add a simple inotify-watcher script to (re)set the credentials in a separate container in the same pod.

What do you think?

bparees:
the additional container feels heavy-weight, why can't you watch in the background of the existing container?

ibotty:
Should be possible, but is harder to monitor. Fine with me either way.

praiskup:
In the beginning, this issue seemed to be about additional roles, but during start of the container.  Now it looks like we need to make some async actions for changing passwords -- which all looks like a wrong idea to me (this might be done by client tool, doesn't have to be done by container itself).

I'm not sure, can you elaborate please what is really needed and how it should be done?

ibotty:
I guess you can say that's a meta-issue. For my part, there are three things I'd like:

 * Storing credentials in secrets (benefit: no credentials in logs, one place to change credentials (postgresql server and client));
 * allowing many database users;
 * updating database credentials automatically (to keep them in sync).

I think the first two part are the most important (for my use case), but the third is a very nice thing when updating credentials. That secrets mounts get changed automatically when the secret changes is to enable that kind of thing. Is there a technical reason for not wanting it?

I also don't see what part of an application deployment should be in charge of changing credentials if not the containers:
When rerolling postgresql user credentials, are you expected to either redeploy (with downtime), or require the user to change the secret (to persist the change) _and_ run psql?

praiskup:
[...]

PR #145, opened 2016-09-26
WIP: use secrets for postgres creds

ibotty:
See issue #101.

This is only in the 9.5 directory for now.

What do you think about it? I don't really like the logic re "simple_db", etc, but I left it as is. When using secrets, it might make sense to not use magic account names (e.g. postgres user can be in any secret, no need for "user" credentials, one valid set of credentials is enough, etc.). If you have a preference on how that should be handled, I'll gladly implement it.

### DIFF
--- a/9.5/root/usr/share/container-scripts/postgresql/common.sh
+++ b/9.5/root/usr/share/container-scripts/postgresql/common.sh
@@ -43,7 +43,7 @@ You must either specify the following environment variables:
   POSTGRESQL_DATABASE (regex: '$psql_identifier_regex')
 Or the following environment variable:
   POSTGRESQL_ADMIN_PASSWORD (regex: '$psql_password_regex')
-Or both.
+Or both (or mount secrets in /run/secrets/pgusers/user or /run/secrets/pgusers/admin).
 Optional settings:
   POSTGRESQL_MAX_CONNECTIONS (default: 100)
   POSTGRESQL_MAX_PREPARED_TRANSACTIONS (default: 0)
@@ -55,8 +55,17 @@ EOF
   exit 1
 }
 
+check_cred_secret() {
+    local credpath="$1"
+    [ -f "$credpath/username" ] && \
+    [ -f "$credpath/password" ] && \
+    [[ "$(<"$credpath/username")" =~ $psql_identifier_regex ]] && \
+    [[ "$(<"$credpath/password")" =~ $psql_password_regex ]] &&
+    [ "$(wc -c < "$credpath/username")" -le 63 ]
+}
+
 function check_env_vars() {
-  if [[ -v POSTGRESQL_USER || -v POSTGRESQL_PASSWORD || -v POSTGRESQL_DATABASE ]]; then
+  if [[ -v POSTGRESQL_USER || -v POSTGRESQL_PASSWORD ]]; then
     # one var means all three must be specified
     [[ -v POSTGRESQL_USER && -v POSTGRESQL_PASSWORD && -v POSTGRESQL_DATABASE ]] || usage
     [[ "$POSTGRESQL_USER"     =~ $psql_identifier_regex ]] || usage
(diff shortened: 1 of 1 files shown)


praiskup:
[test]

9.5/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -65,13 +74,31 @@ function check_env_vars() {
     [ ${#POSTGRESQL_USER}     -le 63 ] || usage "PostgreSQL username too long (maximum 63 characters)"
     [ ${#POSTGRESQL_DATABASE} -le 63 ] || usage "Database name too long (maximum 63 characters)"
     postinitdb_actions+=",simple_db"
+  elif check_cred_secret "/run/secrets/pgusers/user" && [ -v POSTGRESQL_DATABASE ]; then
```
praiskup: This looks like we limit the container for one user, _forever_?  Sounds like we start to define API and that should be defined with respects to future..  I tried to "veto" before, but nobody listened to me (so now we can create just one db with environment variables).  And this looks like similarly limited approach.

9.5/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -172,19 +199,52 @@ function create_users() {
...
+        CREATE USER "${username}" LOGIN;
+    END IF;
+END
+\$body$
+EOF
+}
+
+function set_password() {
+    local user="$1" password="$2"
+    psql --command "ALTER USER \"${user}\" WITH ENCRYPTED PASSWORD '${password}';"
```
praiskup: There is `psql --set` option which should guard against (also not intentional) sql injection.

9.5/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -55,8 +55,17 @@ EOF
   exit 1
 }
 
+check_cred_secret() {
+    local credpath="$1"
+    [ -f "$credpath/username" ] && \
+    [ -f "$credpath/password" ] && \
+    [[ "$(<"$credpath/username")" =~ $psql_identifier_regex ]] && \
+    [[ "$(<"$credpath/password")" =~ $psql_password_regex ]] &&
+    [ "$(wc -c < "$credpath/username")" -le 63 ]
```
praiskup: I don't think that we need to bother doing the regexps, but I understand that this is sort of guideline.  Reason: That's just re-implementing of the checks which are done by PostgreSQL itself.  I bet this check exists in OpenShift gui, too.

praiskup:
As long as some async inotify machinery isn't involved, I am not _strictly_ against this.  But it is basically useless, because environment variables can be used securely, too (see issue #101).  So, to me, this doesn't help security and adds additional complexity for maintainers, reviewers, qa and everybody listening to this code-stream.

The thing is that this is still only going to be used by OpenShift, and we basically add yet another API for container <-> environment communication.  Do we plan to remove the support for env. variables?

praiskup:
Also, there is already prefered API --- kube can directly contact the server (e.g. via unix socket or IP), do such things in transaction way, and react on PostgreSQL errors properly.  Except for protecting against technical debt, I'm mostly interested saving everybody's resources..

9.5/root/usr/share/container-scripts/postgresql/common.sh
ibotty: I do agree. I feel it's kind of limited. But if having one container manage exactly one database (a good idea imo) I can understand doing an automatic createdb. Being able to set the database's superuser also feels right. I don't like the coupling though.

As indicated above, I am fine with implementing any logic that makes sense to you, as long as it's possible to use it for many database users!

ibotty:
>  But it is basically useless, because environment variables can be used securely, too (see issue #101).

For me the most important part of this patch is, that it makes it possible to keep in sync _many_ database users/credentials. That's not nicely done with env vars (and pretty verbose too).

praiskup:
@ibotty I still don't follow, what is meant by *many* database users/credentials?  Do you mean *many* users within single container, syncing one user/credentials among *many* containers?

ibotty:
Sorry, we are talking past each other. So: an example (I am running a variation of that patch in production):

I have _many_ database users connecting to a single database (running in _one_ container). To do that, I mount many secrets (`user1-pgcreds`, ..., `userN-pgcreds`) in the container (mountpath: /run/secrets/pgusers/user1, ..., /run/secrets/pgusers/userN). On container start, all database users get set the right password.

praiskup:
That makes sense to me.  Can we work with subdirectories like `.../userA/password`, etc.?

praiskup:
I mean, I see the benefit of defining new "API" which is able to handle more complicated init scenarios, I'm not sure that working with files is better than simply allowing you to bind-mount arbitrary script and sourcing/executing it.

ibotty:
Subdirectories `.../userA/password`? If you mean, that `.../userA` is a directory, and `password` is a file, that's how it works.
A secret of the form

```yaml
username: my_user
password: my_pass
```

is mapped to the filesystem when mounted as two files in /path/to/secret, username and password.

ibotty:
I do understand your argument re code complexity btw. I am totally fine with weeding out the replicated logic re `simple_db`, and only doing the `traverse users directory` part.

bparees:
it definitely seems simpler to have the image support an init script (Which i think we already have a PR for) and allow that init script to be provided via a mounted secret.  The init script can then create multiple users, or do other useful things.

This mounting of additional files in this way feels fragile.

ibotty:
I disagree. Without the possibility to set multiple users it's not easily possible to, say, have a template generate an application, consisting of a database pod, a management pod and a frontend pod (with restricted db permissions). The implementation of that feature is really easy. Most parts are only moving around functionality (which makes sense regardless of mounting creds with secrets). If the crude simpledb logic were not included, the core part of that patch is maybe 5 lines!

phracek:
As this issue is not touched for couple years, closing it. We do not have capacity to to fix it.

If it is still valid, feel free to re-open this pull request again.

PR #147, opened 2016-10-12
don't install epel-release

bparees:


bparees:
@hhorak @praiskup i don't see why we're installing this, is there a reason?

praiskup:
Seems like we started installing it in 1916d06fab2f4e04c7511beaa17c7d6480a0d4d3 ... but yeah, that's why we should keep in mind how important git-log is.  That's at least not obvious why it is installed, but I guess some epel package was needed that time.

praiskup:
Where is the nss_wrapper installed from now?

bparees:
@praiskup 
```
sh-4.2# repoquery -i nss_wrapper

Name        : nss_wrapper
Version     : 1.0.3
Release     : 1.el7
Architecture: x86_64
Size        : 61024
Packager    : CBS <cbs@centos.org>
Group       : Unspecified
URL         : http://cwrap.org/
Repository  : centos-sclo-rh
```

bparees:
@csrwng do you recall why you started installing epel-release in the commit @praiskup referenced above?

praiskup:
Please mention in commit message that we install nss_wrapper from `centos-sclo-rh` then, and LGTM.

csrwng:
@bparees yeah at the time we added it, we used it to install nss_wrapper

PR #149, opened 2016-11-02
Use private networks and statically allocated IPs for replication tests

praiskup:
Follow up for #144.  @pkubatrh can you please have a look?

### DIFF
--- a/test/run_test
+++ b/test/run_test
@@ -36,13 +36,17 @@ CIDFILE_DIR=$(mktemp --suffix=postgresql_test_cidfiles -d)
 volumes_to_clean=
 
 function cleanup() {
+  script_exit_status=$?
+
   for cidfile in $CIDFILE_DIR/* ; do
     CONTAINER=$(cat $cidfile)
 
     echo "Stopping and removing container $CONTAINER..."
     docker stop $CONTAINER
-    exit_status=$(docker inspect -f '{{.State.ExitCode}}' $CONTAINER)
-    if [ "$exit_status" != "0" ]; then
+
+    if test 0 -ne $script_exit_status -o \
+            0 -ne "$(docker inspect -f '{{.State.ExitCode}}' $CONTAINER)"
+    then
       echo "Dumping logs for $CONTAINER"
       docker logs $CONTAINER
     fi
@@ -88,7 +92,7 @@ function get_ip_from_cid() {
 }
 
 function postgresql_cmd() {
-  docker run --rm ${docker_args:-} -e PGPASSWORD="${PASS}" $IMAGE_NAME psql postgresql://$PGUSER@$CONTAINER_IP:5432/"${DB-db}" "$@"
+  docker run --rm ${psql_docker_args:-} -e PGPASSWORD="${PASS}" $IMAGE_NAME psql postgresql://$PGUSER@$CONTAINER_IP:5432/"${DB-db}" "$@"
 }
 
 function test_connection() {
@@ -310,36 +314,33 @@ function run_tests() {
   fi
 }
 
-function run_slave() {
-  local suffix="$1"; shift
-  docker run $cluster_args -e POSTGRESQL_MASTER_IP=${master_ip} \
-    -d --cidfile ${CIDFILE_DIR}/slave-${suffix}.cid $IMAGE_NAME run-postgresql-slave
-}
-
-function run_master() {
-  local suffix="$1"; shift
-  master_args=${master_args-}
-  docker run $cluster_args $master_args \
-    -d --cidfile ${CIDFILE_DIR}/master-${suffix}.cid $IMAGE_NAME run-postgresql-master >/dev/null
+# run_replication_container {master|slave} IP PRETTY_NAME [CID_SUFFIX]
+# --------------------------------------------------------------------
+# Consumes also '$network', '$cid_suffix', '$test_random_id', '$cluster_args'
+# and either $master_args or $slave_args (based on the first argument).
+run_replication_container ()
+{
+  local container_args=
+  eval "container_args+=\${${1}_args-}"
+  docker run $cluster_args ${container_args} \
+        --name $test_random_id-postgresql-$VERSION-$3 \
+        -e POSTGRESQL_MASTER_IP=$master_ip \
+        --net=$network --ip=$2 \
+        -d --cidfile ${CIDFILE_DIR}/$1-${cid_suffix}${4+-$4}.cid \
+        $IMAGE_NAME run-postgresql-$1
 }
 
 function test_slave_visibility() {
   local max_attempts=30
-
-  for slave in $slave_cids; do
-    slave_ip=$(get_ip_from_cid $slave)
+  for slave_ip in $slave_ips; do
+    : "checking master visibility from slave: $slave_ip"
     for i in $(seq $max_attempts); do
       result="$(postgresql_cmd -c "select client_addr from pg_stat_replication;" | grep "$slave_ip" || true)"
       if [[ -n "${result}" ]]; then
         echo "${slave_ip} successfully registered as SLAVE for ${master_ip}"
         break
       fi
-      if [[ "${i}" == "${max_attempts}" ]]; then
-        echo "The ${slave_ip} failed to register in MASTER"
-        echo "Dumping logs for $slave"
-        docker logs $slave
-        return 1
-      fi
+      test $i -lt $max_attempts
       sleep 1
     done
   done
@@ -354,8 +355,7 @@ function test_value_replication() {
   postgresql_cmd -c "CREATE TABLE $table_name (a integer); INSERT INTO $table_name VALUES ($value);"
 
   # Read value from slaves and check whether it is expected
-  for slave in $slave_cids; do
-    slave_ip=$(get_ip_from_cid $slave)
+  for slave_ip in $slave_ips; do
     CONTAINER_IP=$slave_ip
     for i in $(seq $max_attempts); do
       result="$(postgresql_cmd -At -c "select * from $table_name" || :)"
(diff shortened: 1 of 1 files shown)


praiskup:
Changed `s|/dev/random|/dev/urandom'` .. there's obviously not enough entropy on testing VMs, and that was typo anyway.

praiskup:
@omron93 
On my RHEL7 box, I have
```
# yum info docker
Installed Packages
Name        : docker
Arch        : x86_64
Version     : 1.10.3
....
From repo   : rhel-7-server-extras-rpms
....
```
Is that right that in jenkins we have docker 1.9?  Can we upgrade that, or is this known thing?

omron93:
@praiskup I know about this. It it not perfect...

Does transition of subscription work for you with docker 1.10?

It should work:
1. subscribe host machine
2. `docker run` `rhel7` image and you should be able to install packages inside docker container
(https://access.redhat.com/solutions/1443553)

It does not work for me in Jenkins slaves with never docker than 1.9.

Does it work for you? How did you setup this?

hack/run_test
```diff
@@ -291,36 +303,33 @@ function run_tests() {
...
-
-function run_master() {
-  local suffix="$1"; shift
-  master_args=${master_args-}
-  docker run $cluster_args $master_args \
-    -d --cidfile ${CIDFILE_DIR}/master-${suffix}.cid $IMAGE_NAME run-postgresql-master >/dev/null
+# run_replication_container {master|slave} IP PRETTY_NAME [CID_SUFFIX]
+# --------------------------------------------------------------------
+# Consumes also '$network', '$cid_suffix', '$test_random_id' and
+# '$container_args'.
```
pkubatrh: This comment should probably read cluster_args, as the container_args variable is not used outside of this function.

praiskup: Truth, thanks!  That also reads master/slave_args, I'll fix the comment.

hack/run_test
```diff
@@ -335,85 +344,136 @@ function test_value_replication() {
...
+    docker network create --driver bridge \
+            --subnet=${pattern//X/$check}/24 "$name" &>/dev/null \
+        || continue
+    found=:
+    DOCKER_NETS="${DOCKER_NETS:-} $name"
+    break
+  done
+
+  # Error when we were unable to find free subnet.
+  $found || return 1
```
praiskup: Right, the '|| return 1' part is also redundant.  Is 'false "can't find free docker network"' enough?

pkubatrh: Yes I think that would do just fine.

hack/run_test
```diff
@@ -335,85 +344,136 @@ function test_value_replication() {
...
+    slave_ips="$slave_ips $allocated_ip"
   done
-
 }
 
 function run_master_restart_test() {
   local DB=postgres
   local PGUSER=master
   local PASS=master
+  local network network_ip container_ip
```
praiskup: y

hack/run_test
```diff
@@ -424,8 +484,7 @@ function run_replication_test() {
   echo "Testing master-slave replication"
   local cluster_args="-e POSTGRESQL_ADMIN_PASSWORD=pass -e POSTGRESQL_MASTER_USER=$PGUSER -e POSTGRESQL_MASTER_PASSWORD=$PASS"
   local cid_suffix="basic"
-  local master_ip=
-  local slave_cids=
+  local master_ip= slave_ips=
```
praiskup: y

praiskup:
PTAL

praiskup:
@omron93 wrote:
> Does transition of subscription work for you with docker 1.10?
Yes.

hack/run_test
```diff
@@ -305,8 +305,8 @@ function run_tests() {
 
 # run_replication_container {master|slave} IP PRETTY_NAME [CID_SUFFIX]
 # --------------------------------------------------------------------
-# Consumes also '$network', '$cid_suffix', '$test_random_id' and
-# '$container_args'.
+# Consumes also '$network', '$cid_suffix', '$test_random_id', '$cluster_args'
+# and either $mater_args or $slave_args (based on the first argument).
```
pkubatrh: s/mater/master

hack/run_test
```diff
@@ -446,19 +447,18 @@ function run_master_restart_test() {
...
   local table_name="t1"
-  local master_ip= slave_ips=
-  local volume_dir
 
+  local volume_dir
   create_pg_volume_dir
   local master_args="-v ${volume_dir}:/var/lib/pgsql/data:Z"
 
-  # Setup the cluster
+  # Setup the cluster, keep exported vars loacl only.
```
pkubatrh: s/loacl/local

praiskup: Right, typos fixed, sorry for fat fingers.

praiskup:
We still need to wait for docker-1.9 in RHEL CI.

omron93:
Issue with docker 1.10 has been resolved in recent RHEL update. So now RHEL CI uses docker 1.10.3.

[test]

praiskup:
[test]

praiskup:
[test]

praiskup:
Merged the review fixes to trigger test.

praiskup:
Ah rebased to have 'FROM rhel7' in testsuite, too.

praiskup:
CI seems to be fine, should I fix something/can we merge?

omron93:
@hhorak Please merge.

[test]

hack/run_test
```diff
@@ -377,43 +450,30 @@ function run_master_restart_test() {
...
-  local cidfile=$CIDFILE_DIR/master-$cid_suffix.cid
-  docker kill $(cat $cidfile)
-  # Don't forget to remove its .cid file
-  rm $cidfile
-
-  run_master $cid_suffix
-  CONTAINER_IP=$(get_container_ip master-$cid_suffix.cid)
-  echo "New master is: $CONTAINER_IP"
+  : "Kill the master and create a new one"
+  docker kill "$(cat "$CIDFILE_DIR/master-$cid_suffix.cid")"
```
pkubatrh: While doing some testing of running this test case in parallel, sometimes the test would fail with an "IP already in use" error. We should add a `sleep 1` over here to wait a while for docker to properly clean up after the killed container before trying to run it again with the same IP.

praiskup: The issue with "sleep N" is that you never know whether N is enough forever :(.

praiskup:
> docker create network not being atomic

Thanks for the points.  After several weeks I'm bit desynced so looking at the code  I don't think I claim that 'docker network' should be atomic;  I rather pay attention to pick sufficiently random network name.  So can you please elaborate a bit?  I would rather fix the issues before the merge.

pkubatrh:
> So can you please elaborate a bit? I would rather fix the issues before the merge.

Current code relies on docker to make the check if the subnet of the newly created network overlaps a subnet of an existing network. While docker is indeed doing such a check, sometimes it allows the creation of two differently named networks with the same (in this case) subnet:
```
$ docker network inspect srzhAA_postgresql_9.2 | grep Subnet
                    "Subnet": "172.18.1.0/24"
$ docker network inspect kvszZS_postgresql_9.4 | grep Subnet
                    "Subnet": "172.18.1.0/24"
```
This in turn makes the test fails on container creation (in case the test try to create a container with the same IP). As I pointed out, it is only an issue when multiple tests are run at the same time. When the tests are run one after the other, this should not become an issue.

praiskup:
Argh!  Good point, I forgot how the subnet IP is allocated.  So would it make sense to do some file-system locking in "allocate network" method?

praiskup:
Anyways, have you found a bugreport against Docker?  Because that sounds like a clear
bug.

Well Initially I thought that the networking in different docker networks could be completely isolated (so we can have everywhere the same subnet IP) from others...  but it happens rather randomly.

pkubatrh:
>So would it make sense to do some file-system locking in "allocate network" method?

Yes, I hope file locking would fix the problem for us.

>Anyways, have you found a bugreport against Docker? 

I have hit this issue quite recently so haven't had the time to investigate it yet (need to try reproducing against the newest available version).
After having a quick glance at currently open issues there does not seem to be anything similar. Will get this reported as soon as possible.

praiskup:
@pkubatrh Would does the last commit help?  If yes, I should definitely edit the commit message (it should point to Docker's bugreport).

pkubatrh:
Thanks, will try this out.
Unfortunately I do not have a simple reproducer for this yet (I tried writing a script that just creates two networks with the same subnet at once but docker does not seem to have any problems with that) so this might take a while.

pkubatrh:
Managed to hit yet another issue connected to networking while testing out the flock fix provided by @praiskup 
Sometimes (every ~50 cycles on m1.medium OpenStack instance), one of the tests can find itself unable to use the network it has created for itself as the containers inside that network cannot contact each other (only themselves and the network's gateway). The issue happens pretty often compared to the other I have encountered (once ~50 test cycles).
I have opened up an upstream bug report to get more info about this: https://github.com/docker/docker/issues/30312

pkubatrh:
It seems the issue I described in the previous comment is very similar (or maybe has the same underlying issue) as the 'gateway problem' described in #152 
The reason why the docker containers sharing a single private network were unable to talk to each other was that every packet was send to the gateway instead (routing was setup ok).
Unfortunately, I now find myself unable to reproduce this anymore as I am instead running into the 'gateway problem' once again. I will add some more information into the upstream bug report for the time being and will consider these two issues as a single one.

As for this PR, I have tested the changes over ~500 sequential runs on OpenStack and have not run into any issues, so this is most likely somehow only related to running the tests in parallel. As such I think the PR is ready for merging into master. Maybe take out the last commit (critical section patch) as that is not needed for sequential runs. WDYT @praiskup

praiskup:
On Thursday, January 26, 2017 3:46:33 AM CET pkubatrh wrote:
> packet was send to the gateway instead (routing was setup ok).

So there are always two routing rules, the default route and the
'172.17.0.0 -> eth0'?

> Unfortunately, I now find myself unable to reproduce this anymore as I am
> instead running into the 'gateway problem' once again. I will add some more
> information into the upstream bug report for the time being and will consider
> these two issues as a single one.

OK.  Thanks!

> As for this PR, I have tested the changes over ~500 sequential runs on
> OpenStack and have not run into any issues, so this is most likely somehow
> only related to running the tests in parallel.

Which is a pitty, because parallel testing was the major reason to hack on this
PR.  But yes, we are sort of ready.

> As such I think the PR is ready for merging into master. Maybe take out the
> last commit (critical section patch) as that is not needed for sequential
> runs. WDYT @praiskup 

Maybe.  So should I drop that patch?

I would really be curious where's the place for race-conditions ..  maybe
because we do 'docker run --ip', which does (from our POV) at least two actions:
(a) running the container and (b) assigning ip address.  Do you think
an explicit 'docker network connect' could help?  If you hit this gateway
problem, what happens if you crate yet another network and assign the network to
containers manually?  Will the containers see each other in new network?

pkubatrh:
>So there are always two routing rules, the default route and the '172.17.0.0 -> eth0'?

Yes, that is always the case. The difference between the original gateway problem and its variant that I hit later on is that in the former case the network somehow manages to route the packets to the correct container (even though both the IP and MAC addresses are set to gateway's values). 

>Maybe. So should I drop that patch?

I feel like we should add the critical section once it is actually used by multiple processes which at this moment is not the case (afaik).

>Do you think an explicit 'docker network connect' could help? 

Not sure. Will try running the tests once again using the setup including the latest commits.

>If you hit this gateway problem, what happens if you crate yet another network and assign the network to containers manually? Will the containers see each other in new network?

The problem is limited only to the network it manifests itself on. You can create new networks and containers that connect to it work as they are supposed to. But they will have issues with networking if connected to the problematic network (even when the network gets removed and recreated with the same subnet parameter).

praiskup:
FTR, here is some testing with 'MAKEFLAGS=-j5' (== -j10 because there are two make processes).
```
# https://gist.github.com/praiskup/c787e08296a11938f4d73f66cd701292
$ time ./test_in_parallel
...snip...
real    8m6.990s
user    0m23.105s
sys     0m19.564s
```
vs. the default:
```
$ time make test SKIP_SQUASH=1
....
real    14m39.718s
user    0m7.984s
sys     0m6.602s
```
Both were successful on my box.

praiskup:
[test]

omron93:
@praiskup Update of common-scripts is needed.

praiskup:
I know, I just wanted to see that the VM for CI is allocated and terminated properly.
So, since we start VM for each CI run now, I doubt we have a motivation to continue with
this PR, right?

pkubatrh:
Well this PR was for isolating test-cases within a single run so I think the motivation is still there.

praiskup:
[test] for centos

omron93:
@praiskup @pkubatrh Also you could consider to use internal Docker DNS resolution (--name or --network-alias of a container is resolved into its IP). Instead of complicated ways to assign static IP into container...

Same way is used for mongodb replication test - https://github.com/sclorg/mongodb-container/blob/master/test/ctest_local_replication

pkubatrh:
iirc we have used IPs directly to make sure we are as similar to Openshift (using plain docker) as possible.

omron93:
> iirc we have used IPs directly to make sure we are as similar to Openshift (using plain docker) as possible.

@praiskup That is exactly where I aimed to. Openshift uses services to direct traffic among containers.

And you can emulate services by using same --network-alias for more containers. So resolving this name returns IP of all containers with this alias.

So calling of `dig` in https://github.com/sclorg/postgresql-container/blob/master/latest/root/usr/share/container-scripts/postgresql/common.sh#L104 will return IP of master. And it can change several times.

This is exactly what how it works in OpenShift IMHO. So we can match OpenShift behavior almost exactly without need for special hooks (https://github.com/sclorg/postgresql-container/blob/master/latest/root/usr/share/container-scripts/postgresql/common.sh#L106) for local tests.

pkubatrh:
>So calling of dig in https://github.com/sclorg/postgresql-container/blob/master/latest/root/usr/share/container-scripts/postgresql/common.sh#L104 will return IP of master. And it can change several times.

The PR was created with this not being true for when a pod gets restarted (according to https://github.com/sclorg/postgresql-container/pull/144#issuecomment-257619169) which is the behavior we have been trying to emulate here by forcing the same IP. But this comment is (nearly) a year old already so might not be accurate anymore...

omron93:
> The PR was created with this not being true for when a pod gets restarted (according to #144 (comment)) which is the behavior we have been trying to emulate here by forcing the same IP

OK, I haven't known it. Thanks for explanation.

pkubatrh:
Unfortunately this woudl not work with podman now, since podman lacks a lot of the networking capabilities needed by this PR. So closing for now. We may revisit later once/if podman is able to work with custom configured networks.

praiskup:
We are probably not interested in this PR anymore; because podman.

PR #150, opened 2016-11-07
Bump rhel 7 minor version

praiskup:
find -name Dockerfile.rhel7 -exec sed -i 's/^FROM rhel7.2/FROM rhel7.3/' {} +

### DIFF
--- a/9.2/Dockerfile.rhel7
+++ b/9.2/Dockerfile.rhel7
@@ -1,4 +1,4 @@
-FROM rhel7.2
+FROM rhel7
 
 # PostgreSQL image for OpenShift.
 # Volumes:
--- a/9.4/Dockerfile.rhel7
+++ b/9.4/Dockerfile.rhel7
@@ -1,4 +1,4 @@
-FROM rhel7.2
+FROM rhel7
 
 # PostgreSQL image for OpenShift.
 # Volumes:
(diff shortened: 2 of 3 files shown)


praiskup:
@anybody, could we switch back to "FROM rhel7"?

bparees:
if possible it would be my preference.  the openshift CI infrastructure has to construct a mocked up rhel7.x image, so if you change this, our infrastructure will have to change to use the rhel7.3 tag when constructing the mock image.

bparees:
[test] (to demonstrate the failure on the openshift CI)

praiskup:
@hhorak, does that sound wise?

hhorak:
Yes, I think `FROM rhel7` will be better.

praiskup:
[test]

praiskup:
PTAL

hhorak:
LGTM. @bparees does openshift CI infrastructure need some change for the `FROM rhel7` case as well? Or can we merge now?

bparees:
i've made the CI changes, i think we're ok to merge.
curious why my [test] tag got ignored though.

bparees:
(the test will still fail as the changes are working their way through, but i won't hold you guys up on that, you can merge if you're ready)

hhorak:
Merged.

PR #151, opened 2016-11-10
Allow to select more than one version to build

omron93:
This PR allows to specify more versions to build in VERSION variable. Without this it is not possible.

Also for some repositories fixes lack of simmilar PR as sclorg/mongodb-container#208.

@hhorak Please merge.

PR #152, opened 2016-11-11
Remove VERSION argument from common.mk

pkubatrh:
For the most part removes the use of VERSION argument from common.mk (only used in `make runtests` via build.sh now). Does not modify version handling in build.sh in order to still be able to build all images without calling make.
This should allow us to run multiple docker builds in parallel in the future via make.

Briefly discussed in #151

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -78,7 +78,7 @@ for dir in ${dirs}; do
 
   IMAGE_NAME="${NAMESPACE}${BASE_IMAGE_NAME}-${dir//./}-${OS}"
 
-  if [[ -v TEST_MODE ]]; then
+  if [[ "${TEST_MODE}" == "true" ]]; then
     IMAGE_NAME+="-candidate"
   fi
 
@@ -91,7 +91,7 @@ for dir in ${dirs}; do
     docker_build_with_version Dockerfile
   fi
 
-  if [[ -v TEST_MODE ]]; then
+  if [[ "${TEST_MODE}" == "true" ]]; then
     make -C ../ ${TEST_CASE:-runtests} VERSION=$dir IMAGE_NAME=${IMAGE_NAME}
     if [[ $? -eq 0 ]] && [[ "${TAG_ON_SUCCESS}" == "true" ]]; then
       echo "-> Re-tagging ${IMAGE_NAME} image to ${IMAGE_NAME%"-candidate"}"
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -78,7 +78,7 @@ for dir in ${dirs}; do
 
   IMAGE_NAME="${NAMESPACE}${BASE_IMAGE_NAME}-${dir//./}-${OS}"
 
-  if [[ "${TEST_MODE}" == "true" ]]; then
+  if [[ -v TEST_MODE ]]; then
     IMAGE_NAME+="-candidate"
   fi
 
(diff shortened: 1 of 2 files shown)


hack/build.sh
```diff
@@ -78,7 +78,7 @@ for dir in ${dirs}; do
 
   IMAGE_NAME="${NAMESPACE}${BASE_IMAGE_NAME}-${dir//./}-${OS}"
 
-  if [[ -v TEST_MODE ]]; then
+  if [[ "${TEST_MODE}" == "true" ]]; then
```
praiskup: Looks like this is not needed, I'm not sure.

hack/common.mk
```diff
@@ -11,22 +11,24 @@ endif
...
 	OPENSHIFT_NAMESPACES="$(OPENSHIFT_NAMESPACES)"
 
 .PHONY: build
-build:
-	$(script_env) $(build)
+build: $(VERSIONS)
+
+.PHONY: $(VERSIONS)
+$(VERSIONS):
+	VERSION=$@ TEST_MODE=$(TEST_MODE) $(script_env) $(build)
```
praiskup: This looks good, but are we ready for parallel run?

hack/build.sh
pkubatrh: Right, It might be better to just leave this as it is and change the Makefile rules to not define TEST_MODE when not running `make test`. Thanks

hack/common.mk
pkubatrh: Haven't run into any issues when running this (together with #149) in parallel yet. Is there something you know of that could be a problem for parallel runs?
The only thing that comes to my mind at this moment is that running Docker builds in parallel causes Docker to pull common ancestors (rhel/centos images) at the same time. Will have to check if there are no hidden issues with that.

praiskup:
On Monday, November 14, 2016 3:28:35 AM CET pkubatrh wrote:
> pkubatrh commented on this pull request.
> Haven't run into any issues when running this (together with #149) in
> parallel yet. Is there something you know of that could be a problem for
> parallel runs?

The 'make test' failed for me (on Fedora 24), I need to have a look
whether this is related to parallel make run.

> The only thing that comes to my mind at this moment is
> that running Docker builds in parallel causes Docker to pull common
> ancestors (rhel/centos images) at the same time. Will have to check if
> there are no hidden issues with that.

Hopefully docker behaves correctly here.  Still docker is client-server
tool ..

pkubatrh:
>The 'make test' failed for me (on Fedora 24), I need to have a look whether this is related to parallel make run.

Actually managed to hit a fail as well after running the test-suite a few times. Unfortunately (as I have not been using any output synchronisation) the source of the error is really hard to find. Will try running the test-suite again and investigating after the failure shows itself once more.

pkubatrh:
Managed to hit the same issue again, this time with logs and containers in tact.
The reason for the failure is that after setting up a master-slave replication scheme, for some reason both the master (IP x.x.x.2), and the slave (IP x.x.x.3) are sending replicated data to the subnet's gateway (IP x.x.x.1) and as such `pg_stat_replication` has only an entry for the gateway and not for the slave as it should have.
So far I have no idea why it is happening. Neither am I sure if this is because of a bug in Docker or on our side. The gateway's IP is does not seem to be used anywhere in the test-suite so my guess would be the former, but will have to investigate further.
For the time being I would advise against running tests in parallel.

pkubatrh:
Closing as the makefile is not located in this repository anymore.

https://github.com/sclorg/container-common-scripts/pull/10 has been created instead of this PR.

PR #155, opened 2016-12-07
Pass image ID for testing

omron93:
When running test/run use image ID instead of IMAGE_NAME. This allows parallel testing of images (for example parallel testing of PRs in Jenkins).

This is similar PR as for sclorg/mariadb-container#14. @hhorak Please take a look.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -13,6 +13,9 @@
 # OPENSHIFT_NAMESPACES - Which of available versions (subdirectories) should be
 #       put into openshift/ namespace.
 
+set -o errexit
+set -o pipefail
+
 OS=${1-$OS}
 VERSION=${2-$VERSION}
 
@@ -37,7 +40,10 @@ function docker_build_with_version {
   if [[ "${UPDATE_BASE}" == "1" ]]; then
     BUILD_OPTIONS+=" --pull=true"
   fi
-  docker build ${BUILD_OPTIONS} -t ${IMAGE_NAME} -f "${dockerfile}.version" .
+  local build_output=$(mktemp)
+  docker build ${BUILD_OPTIONS} -t ${IMAGE_NAME} -f "${dockerfile}.version" . | tee $build_output
+  IMAGE_ID="$(awk 'END {print $NF}' $build_output)"
+  rm $build_output
   if [[ "${SKIP_SQUASH}" != "1" ]]; then
     squash "${dockerfile}.version"
   fi
@@ -77,6 +83,7 @@ for dir in ${dirs}; do
   esac
 
   IMAGE_NAME="${NAMESPACE}${BASE_IMAGE_NAME}-${dir//./}-${OS}"
+  IMAGE_ID=
 
   if [[ -v TEST_MODE ]]; then
     IMAGE_NAME+="-candidate"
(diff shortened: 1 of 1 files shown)


praiskup:
@omron93  can you elaborate how this actually helps?  There should be 1:1 mapping from IMAGE_NAME to IMAGE_ID.

pkubatrh:
I guess this is for Docker to be able to run in parallel different builds of images that would otherwise be tagged under the same name?

omron93:
> I guess this is for Docker to be able to run in parallel different builds of images that would otherwise be tagged under the same name?

Exactly.

If you run for example two parallel `make test`. One is faster and is running `./test/run/` script, meantime second finishes building and is tagged as `centos/postgresql-95-centos7-candidate`. First is finishing testing and tags `centos/postgresql-95-centos7-candidate` as `centos/postgresql-95-centos7` (but it is a wrong image)

Also this "substitution" of images can theoretically happen during testing, before it,...

IMPOV best situation would be that during testing of images no "global" resources are used.

praiskup:
@pkubatrh 
> I guess this is for Docker to be able to run in parallel different builds of images that would otherwise be tagged under the same name?

Is this what we want however?  It looks like we should build only once then, and have it backed up by make target?

@omron93 
> Also this "substitution" of images can theoretically happen during testing, before it,...

I don't fully follow ... I'll try to reach you in person for quick chat :)

pkubatrh:
@praiskup 
> Is this what we want however? It looks like we should build only once then, and have it backed up by make target?

If we are talking about testing different PRs at the same time, the built image could be different for each of them. I don't think we can build only once in this case.

@omron93 
What happens after the tests are finished? AFAIK you can't tag two (or more) different images using a single `centos/postgresql-95-centos7` tag.
I would much rather see the distinction between different image builds using tags already during the build phase of said image (for example by appending commit hash to the tag).

praiskup:
After off-github discussion with @omron93, i see the problem -- the issue is that CI might clone the git image into two places (on the same machine), and runs the `make` in parallel against single docker service.

praiskup:
> I would much rather see the distinction between different image builds using tags already during the build phase of said image

Agreed.  We talked about how to identify the path to cloned directory, and generate the tag based on it... the possibilities is to put (a modified, without `/` slashes) clone path into tag name, or e.g. dev_id + inode (which is less readable).

> (for example by appending commit hash to the tag).

Git hash could be tricky for modified user's directories ... (where `git diff` is non-empty, that's not a problem in CI), though it is probably somehow possible.

Tagging under 'centos/postgresql-95-centos7' must be explicit action then.

omron93:
> CI might clone the git image into two places (on the same machine), and runs the make in parallel against single docker service.

Yes, I would like to achieve to be able to do this.

> What happens after the tests are finished? AFAIK you can't tag two (or more) different images using a single centos/postgresql-95-centos7 tag.

IMPOV this is not a problem of build scripts - only content of the directory should be built and after that tagged (`make build` or `make test` can't guarantee that it won't be retagged for some time). So it is problem of a user, if it wants to build two images in parallel (not the good idea), only one image can be tagged in the name! Build scripts can't do anything with this

pkubatrh:
>Git hash could be tricky for modified user's directories ... (where git diff is non-empty, that's not a problem in CI), though it is probably somehow possible.

We could change the Makefile to allow to specify a tag that should be used during the build (and testing). That way the CI could use a git hash for its PR testing and users could tag it as whatever they want to, as long as it is unique for their Docker daemon.

omron93:
> We could change the Makefile to allow to specify a tag that should be used during the build (and testing). That way the CI could use a git hash for its PR testing and users could tag it as whatever they want to, as long as it is unique for their Docker daemon.

Yes, it might be good idea.

However approach used in this PR is not possible for s2i based images (running `s2i build` fails - using image ID instead of tag is not supported). So am closing this PR (and similar PRs for other container images) to keep building scripts same s2i based and non-s2i based images.

As mentioned above, different approach could be used, so I will try to prepare something. But it would be helpful to have shared build scripts as mentioned in #138.

Issue #163, opened 2017-03-13
PostgreSQL containers fail to start on Fedora 25


praiskup:
```
$ docker run --rm -ti openshift/postgresql-92-centos7 
Warning: Can't detect memory limit from cgroups
Warning: Can't detect number of CPU cores from cgroups
/usr/share/container-scripts/postgresql/common.sh: line 9: MEMORY_LIMIT_IN_BYTES: unbound variable
```

praiskup:
Probably some security feature kills this hack.  My original complaint was that this is not something to be done by container itself (some ugly heuristics) -- environment should set those options.

As a band-aid, we should use `${MEMORY_LIMIT_IN_BYTES-}`, and probably remove the hack completely?

pkubatrh:
This seems to have happened because of a change in container-selinux policies, but is already fixed in the latest version of the package (container-selinux-2.10-1) so there is no need to hotfix this unless to make sure it does not happen in future.

praiskup:
Because we have `set -e`, it still is an issue.  Could you cite the ticket related to container-selinux issue?

pkubatrh:
Have not found a ticket related to this anywhere. I have just updated to the most recent version of the package and that fixed the issue for me.
My knowledge of selinux and its policies is quite limited but I think this is the upstream commit that modifies the access to cgroups files:
https://github.com/projectatomic/container-selinux/commit/c9829f9ea1542d876bc4641720c1a373db47ab20

praiskup:
Ok, thanks for the info!  Seems like the recent update fixed my box too.

PR #164, opened 2017-03-13
Don't fail if $MEMORY_LIMIT_IN_BYTES is unset

praiskup:
This is result of failing root/usr/bin/cgroup-limits script;
which throws warnings in case of failure.

Hot-fixes: #163


hhorak:
Makes sense, merging.

PR #165, opened 2017-03-13
migration: add basic construct to help with migrations

praiskup:
The dump & restore way.

### DIFF
--- a/9.5/root/usr/share/container-scripts/postgresql/README.md
+++ b/9.5/root/usr/share/container-scripts/postgresql/README.md
@@ -18,6 +18,14 @@ initialization by passing `-e VAR=VALUE` to the Docker run command.
 |  `POSTGRESQL_DATABASE`       | Database name                                  |
 |  `POSTGRESQL_ADMIN_PASSWORD` | Password for the `postgres` admin account (optional)     |
 
+Alternatively, the following options are related to migration scenario:
+
+|    Variable name                       |    Description                         |
+| :------------------------------------- | -------------------------------------- |
+|  `POSTGRESQL_MIGRATION_REMOTE_HOST`    | Hostname/IP to migrate from            |
+|  `POSTGRESQL_MIGRATION_ADMIN_PASSWORD` | Password for the remote 'postgres' admin user |
+|  `POSTGRESQL_MIGRATION_IGNORE_ERRORS`  | Set to 'yes' to ignore sql import errors (optional, default 'no') |
+
 The following environment variables influence the PostgreSQL configuration file. They are all optional.
 
 |    Variable name              |    Description                                                          |    Default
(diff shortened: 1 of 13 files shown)


praiskup:
@hhorak fyi ... untested tbh.

praiskup:
@hhorak PTAL

9.5/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -187,6 +200,23 @@ function set_passwords() {
...
+    # Migration path.
+    (
+        if [ ${POSTGRESQL_MIGRATION_IGNORE_ERRORS-no} = no ]; then
+            echo '\set ON_ERROR_STOP on'
+        fi
+        # initdb automatically creates 'postgres' role;  creating it again would
+        # fail the whole migration so we drop it here
+        PGPASSWORD="$POSTGRESQL_MIGRATION_ADMIN_PASSWORD" \
+        pg_dumpall -h "$POSTGRESQL_MIGRATION_REMOTE_HOST" \
+            | grep -v '^CREATE ROLE postgres'
```
pkubatrh: This will drop every role beginning with 'postgres' instead of just dropping initdb's 'postgres' role.

9.5/root/usr/share/container-scripts/postgresql/README.md
```diff
@@ -61,6 +69,55 @@ or if it was already present, [`postgres`](http://www.postgresql.org/docs/9.2/st
...
+this process to not waste additional storage space.  On the other hand, any
+failing SQL command on the pipe causes immediate failure of whole migration
+process (this typically happens if the remote cluster uses some server-side
+module which is not available/enabled in the actual container).
+
+We fail by default because we want to ensure the **all** or **nothing**
+result of scripted, unattended migration;  while in general, successful
+migration is expected (not guaranteed!) anyway when you migrate from previous
+versions of PostgreSQL server in `sclorg` container (e.g. migration from
+`openshift/postgresql-92-centos7` to `centos/postgresql-95-centos7`).
```
hhorak: This paragraph sounds a bit artificially, what about this:
```
If some SQL commands fail during applying, the default behavior
of the migration script is to fail as well to ensure the **all** or **nothing**
result of scripted, unattended migration. In most common cases, successful
migration is expected (but not guaranteed!), given you migrate from
a previous version of PostgreSQL server container, that is created using
the same principles as this one (e.g. migration from
`openshift/postgresql-92-centos7` to `centos/postgresql-95-centos7`).
Migration from a different kind of PostgreSQL container can likely fail.
```

9.5/root/usr/share/container-scripts/postgresql/common.sh
praiskup: Good point, thanks.

praiskup:
Thanks for comments, PoC PR fixed.

praiskup:
The major benefit of #172 compared to this PR is that it is easily deployable - but how big problem would be to hack on openshift so that redeploying from `centos/postgresql-94-centos7-candidate` to  `centos/postgresql-95-centos7-candidate` requires some automatic steps?  Because having such functionality would be ultimate win here.

9.5/root/usr/share/container-scripts/postgresql/README.md
```diff
@@ -61,6 +69,55 @@ or if it was already present, [`postgres`](http://www.postgresql.org/docs/9.2/st
...
+
+If this **all** or **nothing** principle is inadequate for you, and you know
+what you are doing, there's optional `POSTGRESQL_MIGRATION_IGNORE_ERRORS` option
+which does **best effort** migration (some data might be lost, it is up to user
+to review the standard error output and fix the issues manually in
+post-migration time).
+
+So before you do database migration, be prepared to perform manual steps to get
+all your data migrated.  We provide the helpers for users' convenience, but it
+is in general impossible to guarantee fully automatic migration.
```
hhorak: Please, mind, that the container image provides help for users' convenience, but fully automatic migration is not guaranteed. Thus, before you start proceeding with the database migration, get prepared to perform manual steps in order to get all your data migrated.

hhorak:
I haven't spotted any obvious issue and it looks clear to me. As for nitpicking, I've added a proposal for restructuralize one paragraph a bit..

pkubatrh:
Agree with Honza's suggestion for the doc edit. Otherwise lgtm.

Also needs to be added to other postgresql versions before merging.

praiskup:
FWIW, waiting for #175 is done, to avoid unnecessary rebases; let me know if I should work on this first.

praiskup:
Please take a look.  I've added the migration support into v9.4+ and added a pagila testcase.

praiskup:
[test]

praiskup:
[test]

praiskup:
[test]

pkubatrh:
lgtm, thanks! Merging

PR #167, opened 2017-03-17
Speedup 'docker build' from caches

praiskup:
Installing of RPM takes fair amount of time so we need to have it
done before 'ADD root /' statement;  so if something is changed in
'root' directory, 'docker build' doesn't have to re-download and
re-install all the packages.

There's, however, still the 'aufs' docker issue we touched in
issue #116 (docker/docker#24660) so we need to have
`fix-permissions` installed before package installation is done.


praiskup:
Ping.

omron93:
LGTM. Thanks.

PR #171, opened 2017-05-15
tests: Only test help.1

pkubatrh:
Also pulls in newer sclorg/container-common-scripts to be able to build help.1

same PR as https://github.com/sclorg/mariadb-container/pull/35


praiskup:
Looks to be OK.

pkubatrh:
Modified the PR to leave the `for` loop in place in case we need to test multiple files in future, as pointed out in https://github.com/sclorg/ror-container/pull/17

PR #172, opened 2017-05-16
pg_upgrade approach

praiskup:
This is "data directory upgrade" PoC 2.
Complements PR #165.

**Pros:**
- should be faster compared to #165, though this needs to be benchmarked once we have PoC 1 implemented by fully automatic way
- easily deploy-able, by setting `$POSTGRESQL_UPGRADE` and
  ```diff
  - centos/postgresql-95-centos7-candidate
  + centos/postgresql-95-centos7-candidate
  ```

**Cons:**
- cumbersome for N -> N+2 migration paths (if we provided version A, B and C, migration from A to C means going through B), this beats the migration a lot
- requires doubled amount of free space within physical volume, and there's no easy way how to detect that
- non-systematic POV -> two versions of the same software need to be installed into one container; that's why containers are there (let's install each SW version into separate container...)
- we heavily depend on `scls` now ... so far we were trying to hide the fact the PostgreSQL server and the tooling requires the `scl enable ...` things...  Accepting this PR basically means that we won't be able to avoid using `scl` tooling.
- downgrading is not possible (we can only do N to N+1)
- downtime is unavoidable, in the #165 PoC we can keep the old container running at least read-only (but maybe more) till the migration is finished

**TODOs/Ideas**
- somehow detect there's not enough space in /var/lib/pgsql/data (or quota)
- backup internaly? (but that would mean 300% instead of 200% space needed)

Let me know if I forgot about something.


praiskup:
@bazulay, @ccoleman, @hhorak, @pkubatrh, @rthallisey, please take a look.

rthallisey:
@praiskup Excellent write up.

9.5/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -218,3 +223,149 @@ function wait_for_postgresql_master() {
...
+  test -f "$versionfile"
+  version=$(cat "$versionfile")
+
+  # If we don't support pg_upgrade, skip.
+  test -z "${POSTGRESQL_PREV_VERSION-}" && return 0
+
+  # No need to call pg_upgrade, skip.
+  if test "$POSTGRESQL_VERSION" = "$version"; then
+      # Make _sure_ we have this safety-belt here, otherwise our users would
+      # just specify '-e POSTGRESQL_UPGRADE=hardlink' permanently, even for
```
bazulay: I would not give the option to POSTGRESQL_UPGRADE=hardlink

9.5/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -218,3 +223,149 @@ function wait_for_postgresql_master() {
...
+      echo >&2 "made some mistake, keeping the variable set you might"
+      echo >&2 "risk a data loss in future!"
+      echo >&2 "==============="
+      echo >&2
+
+      # Exit here, but allow _really explicit_ foot-shot.
+      ${POSTGRESQL_UPGRADE_FORCE-false}
+      return 0
+  fi
+
```
bazulay: Why even consider running upgrade when the versions match ?

9.5/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -218,3 +223,149 @@ function wait_for_postgresql_master() {
...
+  # the upgrade
+  info_msg "Starting the pg_upgrade process."
+  "${upgrade_cmd[@]}" || { rm -rf "$PGDATA_new" && false ; }
+
+  # Move the important configuration and remove old data.  This is highly
+  # careless, but we can't do more for this over-automatized process.
+  info_msg "Swap the old and new PGDATA and cleanup."
+  mv "$PGDATA"/*.conf "$PGDATA_new"
+  rm -rf "$PGDATA"
```
bazulay: why not rename it as $PGDATA_backup_$(date +%s) .... and delete it only next upgrade (before calculating the free space) ?

bazulay:
@fbladilo PTAL

9.5/root/usr/share/container-scripts/postgresql/README.md
```diff
@@ -96,3 +96,38 @@ values stored in the variables and the actual passwords. Whenever a database
...
+container tooling under /var/lib/pgsql/data, which is usually external
+bind-mountpoint.  The `pg_upgrade` execution is then similar to dump&restore
+approach -- it starts both old and new PostgreSQL servers (within container) and
+"dumps" the old datadir while and at the same time it "restores" it into new
+datadir.  This operation requires a lot of data files copying, so you can decide
+what type of upgrade you'll do by setting `$POSTGRESQL_UPGRADE` appropriately:
+
+|    Variable value  |    Description                                 |
+| :----------------- | ---------------------------------------------- |
+|  `copy`            | The data files are copied from old datadir to new datadir.  This option has low risk of data loss in case of some upgrade failure. |
```
praiskup: Me too, that's by default anyway..  Or what's your suggestion here?

9.5/root/usr/share/container-scripts/postgresql/common.sh
praiskup: Because there's not much motivation to waste additional space (if I got your question right).  This is pattern from current RPM packaging:  two predictable path names ... we automatically remove the new **broken** directory what that makes no sense to keep it.

9.5/root/usr/share/container-scripts/postgresql/README.md
bazulay: Not to have a choice ... always copy

9.5/root/usr/share/container-scripts/postgresql/common.sh
praiskup: I picked `link` because that's what also `pg_upgrade` uses (`--link` option).

praiskup:
[test]

9.5/root/usr/share/container-scripts/postgresql/README.md
praiskup: If we remove that option, we remove the only significant benefit compared to #165 -> "faster" upgrade.

9.5/root/usr/share/container-scripts/postgresql/common.sh
praiskup: That's to mimic what is happening in `mariadb` container or on Debian's PostgreSQL (as I was told at least).  Some people want to automatically upgrade when it is possible "transparently" -> I don't agree that's good option, but my thought was to allow this when user really knows he want's to do that.

praiskup: I mean, they might want to set  `-e POSTGRESQL_UPGRADE=copy POSTGRESQL_UPGRADE_FORCE=yes` which is simply `no-op` in case of no pg_upgrade is needed (container just starts), but it does the upgrade if it is possible.

9.5/root/usr/share/container-scripts/postgresql/common.sh
bazulay: The motivation for leaving the old backup could be that in the future one might want to rollback from the application perspective.  
For example a stateful application constructed out of frontend, backend & db ... usually the orchestration of upgrade for such application means to bring down the frontend, backuend , upgrade db , upgrade be, upgrade fe. so in case the upgrade of rhe DB succeeded while the upgrade of the application failed and the admin wants to roll back (even an auto rollback is possible)

9.5/root/usr/share/container-scripts/postgresql/common.sh
bazulay: what fits in a regular OS & its file system does not necessarily fits k8s,
for data safety I would would always use copy , 
But we can have it as a parameter just for testing ... I would not expose it.

praiskup: @bazulay, users are warned properly though. They have to backup anyways right before they do `POSTGRESQL_UPGRADE=copy`.  There is no point to remove the only benefit of doing `pg_upgrade` ...

9.5/root/usr/share/container-scripts/postgresql/common.sh
praiskup: Well, I simply disagree. We start the DB right after pg_upgrade run, and there's a high risk of `ENOSPC` just because we would keep about 100% of datadir size unnecessarily.  I agree that having an additional option explicitly **asking** for not deleting the datadir might be useful ...

My point about all of this binary datadir upgrade is:  Don't do upgrades unless you really know what you are doing ... It is not place to play games;  also, I'm sure wise admins won't do upgrades unless that's really necessary. Which in turn means that they would downvote this `pg_upgrade` N->(N+1) wofkflow.

praiskup:
I pushed one bugfix commit for the logic, revealed by testsuite.

praiskup:
Typical test case would be:
```bash
mkdir data
setfacl -m u:postgres:rwx data
chcon  -R -t svirt_sandbox_file_t data
docker run --rm -ti -e POSTGRESQL_ADMIN_PASSWORD=admin_pass \
-v `pwd`/data:/var/lib/pgsql/data centos/postgresql-94-centos7
... this initializes the data ...
... and in separate shell add some data into the db ...
^C
sudo tar -czf data.tar.gz --selinux --acls --xattrs data # backup
sudo rm -rf data
sudo tar -xf data.tar.gz --selinux --xattrs --acls

# Keep the same options as for re-deployment, but add POSTGRESQL_UPGRADE
# environment varialbe + change the image name:
docker run  -e POSTGRESQL_ADMIN_PASSWORD=admin_pass \
 -v `pwd`/data:/var/lib/pgsql/data \
 -e POSTGRESQL_UPGRADE=hardlink \
 centos/postgresql-95-centos7
```

9.5/root/usr/share/container-scripts/postgresql/common.sh
bazulay: If one chooses to use environment variables to indicate an upgrade, it can not be used on openshift deployment config, because there is no way to give a replication controller specific env variables on a reschedule. this method can only  be used with a deployer pod.

carbonin: @bazulay I think that's the case @praiskup is accounting for here.

If the upgrade env vars are set, but no upgrade is necessary we don't run the upgrade. So to get the deployment config type of behavior (auto-upgrade) you would just always deploy with `POSTGRESQL_UPGRADE=copy` but not the force option.

This way you would get an upgrade if you deploy the container from the next version, but would have to endure a warning in your log every other reschedule.

carbonin:
I'm 👍 on this approach for a single container solution.
This will allow opt-in upgrades for deployments and is very similar to the upgrade procedures we automated for the ManageIQ [PG9.4 to PG9.5 upgrade](https://github.com/ManageIQ/manageiq-appliance/blob/master/LINK/usr/bin/miq_postgres_upgrade.sh)

I'm not sure if it was considered, but could we utilize something like [this](https://github.com/tianon/docker-postgres-upgrade) in an [init-container](https://docs.openshift.org/latest/architecture/core_concepts/pods_and_services.html#pods-services-init-containers)? We could install *all* the currently supported versions of PG and write the container execution code to recognize the previous version and provide a target version using an ENVVAR. I feel like this would eliminate both the need to install multiple versions in the same image as well as the limitation of only upgrading from version N to N+1.

This container should be runnable as a standalone `docker run` (as the linked one is), but could also be configured as an init-container when running on openshift or kubernetes.

Thoughts?

praiskup:
@carbonin, sorry for late reply - we considered the "upgrade-only" container... The "init-container" is new for me TBH, and it is interesting automation idea -- it's implementation looks like it depends on the #165 pull request.  So we could experiment with this later.

praiskup:
JFTR, installing another version of PostgreSQL ATM means ~15MB more data in the squashed (CentOS, as I tested the build on my Fedora box) image:  285 -> 299 MB.

praiskup:
Ugh, this is rather fat PR.  I decided to implement the upgrade path into all 3 supported image versions (to allow migration from 9.2, too); and I added test-case for the `pg_upgrade` case.  For that purpose I added pagila [1] database into `hack/pagila` directory.  This means ~3MB of data...  which is pretty ugly, tbh.

Pagila is pretty good testing database, which I'd really like to have in the test-cases...  so do you think it would be OK to make the tests depend on `pagila` package from host machine (it would mean that we need to install `pagila` our CI hosts). @omron93 ?

[1] http://pgfoundry.org/projects/dbsamples

praiskup:
Any remark is welcome!

hhorak:
Indeed this PR is fat :) but I must say good work!! I went through it rather quicker, but haven't spotted any issues... Just as for the pagila, I think downloading during the test execution should be possible as well. IMO we can expect git/wget/etc available on the testing machine.

pkubatrh:
The changes lgtm, havent spotted any issues either.

Needs to be rebased against latest master to get rid of recent rhel7 Dockerfile conflicts.

praiskup:
PTAL

9.4/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -218,3 +223,165 @@ function wait_for_postgresql_master() {
...
+  info_msg "Starting the pg_upgrade process."
+
+  # Once we stop support for PostgreSQL 9.4, we don't need
+  # REDHAT_PGUPGRADE_FROM_RHEL hack as we don't upgrade from 9.2 -- that means
+  # that we don't need to fiddle with unix_socket_director{y,ies} option.
+  REDHAT_PGUPGRADE_FROM_RHEL=1 \
+  "${upgrade_cmd[@]}" || { rm -rf "$PGDATA_new" && false ; }
+
+  # Move the important configuration and remove old data.  This is highly
+  # careless, but we can't do more for this over-automatized process.
```
carbonin: I've also had problems with this.

The real issue comes when a new version breaks compatibility with parameters, this happened from 9.4 to 9.5  (specifically for the checkpoint_segments config variable moving to max/min_wal_size). In this case if a user specified those settings in their conf the server would not start after upgrade.

I think this (replacing the config) is still the most simple solution for now, but as a follow-up/enhancement could we utilize the concept of an [include directory](https://www.postgresql.org/docs/9.4/static/config-setting.html#CONFIG-INCLUDES) for postgresql configs?

This would separate user changes from the default config shipped with a particular version and might mitigate the issues encountered here. Additionally this would map nicely to using OpenShift/Kubernetes [ConfigMaps](https://docs.openshift.org/latest/dev_guide/configmaps.html) for user PG config overrides.

praiskup: @carbonin wrote:
> I think this (replacing the config) is still the most simple solution for now, but as a follow-up/enhancement could we utilize the concept of an include directory for postgresql configs?
Try:
```
cat data/userdata/postgresql.conf | grep ^include
include '/var/lib/pgsql/openshift-custom-postgresql.conf'
```

So in postgresql.conf should only be the default options + the include statement.

praiskup: But yeah, I really think that the `pg_upgrade` won't work well for everybody;  and that's why we make the documentation saying that users should make themselves comfortable with manual rollback in case of any failure.

praiskup:
@pkubatrh, @hhorak -> I'm curious whether I can yet update this PR somehow?

hhorak:
@praiskup The current rebuild round is over, so IMO yes, it's still time to change this PR.

praiskup:
I meant :) is some change needed, or can we move forward?

praiskup:
Ping, can we merge?

pkubatrh:
As no one is against this I will be merging it in. Thanks for the changes Pavel!

PR #173, opened 2017-05-22
Dockerfiles update

omron93:
Use COPY instruction instead of ADD
(COPY is preffered for simple usage - see https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy)

Provide same labels for all base image variants (CentOS, RHEL, Fedora)
 - follow recomended labels from Project Atomic Container best practices
 - add summary and description labels (Fedora requirements) - use ENV for simple using of same value in more descriptive labels
 - remove architecture label for RHEL based images (this label is set in RHEL image - no need to define it again)

@pkubatrh @praiskup What should be values of `SUMMARY` and `DESCRIPTION`? (my suggestion is mix of descriptions from rpms, probably missing something important)

@hhorak @praiskup @pkubatrh Please take a look and merge.

### DIFF
--- a/9.4/Dockerfile
+++ b/9.4/Dockerfile
@@ -16,10 +16,23 @@ ENV POSTGRESQL_VERSION=9.4 \
     HOME=/var/lib/pgsql \
     PGUSER=postgres
 
-LABEL io.k8s.description="PostgreSQL is an advanced Object-Relational database management system" \
+ENV SUMMARY="The programs needed to create and run a PostgreSQL server $POSTGRESQL_VERSION" \
+    DESCRIPTION="PostgreSQL is an advanced Object-Relational database management \
+system (DBMS). This image contains the programs needed to create \
+and run a PostgreSQL server, the client \
+programs that you'll need to access a PostgreSQL DBMS server, as \
+well as HTML documentation for the whole system."
+
+LABEL summary=$SUMMARY \
+      description="$DESCRIPTION" \
+      io.k8s.description="$SUMMARY" \
       io.k8s.display-name="PostgreSQL 9.4" \
       io.openshift.expose-services="5432:postgresql" \
-      io.openshift.tags="database,postgresql,postgresql94,rh-postgresql94"
+      io.openshift.tags="database,postgresql,postgresql94,rh-postgresql94" \
+      name="centos/postgresql-94-centos7" \
+      com.redhat.component="rh-postgresql94-docker" \
+      version="9.4" \
+      release="1"
 
 EXPOSE 5432
 
--- a/9.4/Dockerfile
+++ b/9.4/Dockerfile
@@ -16,12 +16,10 @@ ENV POSTGRESQL_VERSION=9.4 \
     HOME=/var/lib/pgsql \
     PGUSER=postgres
 
-ENV SUMMARY="The programs needed to create and run a PostgreSQL server $POSTGRESQL_VERSION" \
-    DESCRIPTION="PostgreSQL is an advanced Object-Relational database management \
-system (DBMS). This image contains the programs needed to create \
-and run a PostgreSQL server, the client \
-programs that you'll need to access a PostgreSQL DBMS server, as \
-well as HTML documentation for the whole system."
+ENV SUMMARY="Advanced Object-Relational database management system" \
+    DESCRIPTION="PostgreSQL is an advanced Object-Relational database management system (DBMS). \
+The image contains the client and server programs that you'll need to \
+create, run, maintain and access a PostgreSQL DBMS server."
 
 LABEL summary=$SUMMARY \
       description="$DESCRIPTION" \
--- a/9.4/Dockerfile
+++ b/9.4/Dockerfile
@@ -16,7 +16,7 @@ ENV POSTGRESQL_VERSION=9.4 \
     HOME=/var/lib/pgsql \
     PGUSER=postgres
 
-ENV SUMMARY="Advanced Object-Relational database management system" \
+ENV SUMMARY="PostgreSQL is an advanced Object-Relational database management system" \
     DESCRIPTION="PostgreSQL is an advanced Object-Relational database management system (DBMS). \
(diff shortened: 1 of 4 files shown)


praiskup:
Please consider documenting in "commit message" first, and then C&P the commit to github.  Git log history is what will persist forever.


> Add summary and description labels + use same labels in CentOS images as in RHEL images. @pkubatrh @praiskup What should be values of SUMMARY and DESCRIPTION? (my suggestion is mix of descriptions from rpms, probably missing something important)

I would suggest copying what we have in fedora pkgdb, https://admin.fedoraproject.org/pkgdb/package/container/postgresql/ or if we conclude that there's some better wording, let's sync that wording into Fedora pkgdb, too.

> Remove architecture label - it is set in RHEL base image so not need to redefine it.

Is it set in fedora and centos base images too?

omron93:
> Please consider documenting in "commit message" first, and then C&P the commit to github. Git log history is what will persist forever.

You are right. Commit messages updated.

> I would suggest copying what we have in fedora pkgdb, https://admin.fedoraproject.org/pkgdb/package/container/postgresql/ or if we conclude that there's some better wording, let's sync that wording into Fedora pkgdb, too.

Updated. Thanks.

> Is it set in fedora and centos base images too?

No, it isn't.

Fedora guidelines contains:
"architecture -- Architecture the software in the image should target (Optional: if omitted, it will be built for all supported Fedora Architectures)"

"architecture: usually "x86_64", unless the container image supports other/all architectures. "

So not sure what is the purpose of this label. For example in Fedora packages are often available for all Fedora primary architectures.

Any suggestions how to handle this label?

hhorak:
As for the architecture label, we probably need to follow the guidelines, but I've just submitted a request for setting this label in the base image (like it is done in case of RHEL image): https://pagure.io/atomic-wg/issue/276 (until it is done, I think, we need to keep maintaining this label in the Fedora dockerfile though).

omron93:
@hhorak And what about CentOS? Base image doesn't have `architecture` label and also no guidelines :-)

hhorak:
@omron93 My understanding of this is that it is not needed at this point and in case of RHEL/Fedora images, it's purely "preparing for the future". So, once it will be necessary in case of base CentOS image, we'll either find the label in the base or we'll realize what needs to be done. Long story short, I wouldn't get stressed at this point and simply not putting the label into centos images :)

9.4/Dockerfile
```diff
@@ -16,14 +16,25 @@ ENV POSTGRESQL_VERSION=9.4 \
     HOME=/var/lib/pgsql \
     PGUSER=postgres
 
-LABEL io.k8s.description="PostgreSQL is an advanced Object-Relational database management system" \
+ENV SUMMARY="Advanced Object-Relational database management system" \
```
hhorak: We should keep PostgreSQL keyword in the summary, for cases it is the only field available.. So, I'd rather use something like `PostgreSQL object-relational database server`

praiskup: +1, I'll try to change this in pkgdb

praiskup: Is the previous wording `PostgreSQL is an advanced Object-Relational database management system` ok, then?

hhorak: Such a sentence with a verb is not used in other containers, but in this case it sounds good to me and I have no better alternative..

omron93: +1, updated.

omron93:
Added change to remove use of `MAINTAINER` instruction.

9.5/Dockerfile.rhel7
```diff
@@ -14,21 +14,25 @@ ENV POSTGRESQL_VERSION=9.5 \
...
+    DESCRIPTION="PostgreSQL is an advanced Object-Relational database management system (DBMS). \
+The image contains the client and server programs that you'll need to \
+create, run, maintain and access a PostgreSQL DBMS server."
 
-# Labels consumed by Red Hat build service
-LABEL name="rhscl/postgresql-95-rhel7" \
+LABEL summary=$SUMMARY \
+      description="$DESCRIPTION" \
+      io.k8s.description="$SUMMARY" \
+      io.k8s.display-name="PostgreSQL 9.4" \
```
praiskup: s/9.4/9.5/ is expected

9.5/Dockerfile
```diff
@@ -10,20 +10,30 @@ FROM centos:centos7
...
+
+LABEL summary=$SUMMARY \
+      description="$DESCRIPTION" \
+      io.k8s.description="$SUMMARY" \
+      io.k8s.display-name="PostgreSQL 9.4" \
       io.openshift.expose-services="5432:postgresql" \
-      io.openshift.tags="database,postgresql,postgresql95,rh-postgresql95"
+      io.openshift.tags="database,postgresql,postgresql94,rh-postgresql94" \
+      name="centos/postgresql-94-centos7" \
+      com.redhat.component="rh-postgresql94-docker" \
```
praiskup: is `com.redhat.component` label what we need for centos imae?

hhorak: I was thinking about this as well, and came to something like this: If people have problems with centos images, we should ideally know about it and there is probably no better way than reporting to RH bugzilla (or github if they find it).. This is why `com.redhat.component` makes sense to me.. But I accept that this is not a strong reason..

9.5/Dockerfile.rhel7
praiskup: ping @omron93, this needs to be fixed

omron93: Sorry. It is fixed now.

praiskup:
I guess the missing 'maintainer' LABEL in RHEL's dockerfiles is intentional. Then LGTM, thanks!

omron93:
> I guess the missing 'maintainer' LABEL in RHEL's dockerfiles is intentional.

Yes, it is intentional. Honza (@hhorak) should know why don't want it...

PR #176, opened 2017-07-13
prototype of new README.md for version 9.5 according new rules

ficap:
needs deeper review. some lines are new and also in original version there were bad version numbers in links,.. too. I tried to fix that.

### DIFF
--- a/9.4/root/usr/share/container-scripts/postgresql/README.md
+++ b/9.4/root/usr/share/container-scripts/postgresql/README.md
@@ -1,12 +1,56 @@
-PostgreSQL Docker image
-=======================
+% POSTGRESQL-94(1) Container Image Pages
+% SoftwareCollections.org
+% July 19, 2017
 
-This repository contains Dockerfiles for PostgreSQL images for general usage and OpenShift.
+PostgreSQL 9.4 SQL Database Server Docker image
+===============================================
+
+This container image includes PostgreSQL 9.4 SQL database server for OpenShift and general usage.
 Users can choose between RHEL and CentOS based images.
+The RHEL image is available in the [Red Hat Container Catalog](https://access.redhat.com/containers/#/registry.access.redhat.com/rhscl/postgresql-94-rhel7)
+as registry.access.redhat.com/rhscl/postgresql-94-rhel7.
+The CentOS image is then available on [Docker Hub](https://hub.docker.com/r/centos/postgresql-94-centos7/)
+as centos/postgresql-94-centos7.
+
+
+Description
+-----------
+
+This container image provides a containerized packaging of the PostgreSQL postgres daemon
+and client application. The postgres server daemon accepts connections from clients
+and provides access to content from PostgreSQL databases on behalf of the clients.
+You can find more information on the PostgreSQL project from the project Web site
+(https://www.postgresql.org/).
+
+
+Usage
+-----
+
+For this, we will assume that you are using the `rhscl/postgresql-94-rhel7` image.
+If you want to set only the mandatory environment variables and not store the database
+in a host directory, execute the following command:
+
+```
+$ docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 rhscl/postgresql-94-rhel7
+```
+
+This will create a container named `postgresql_database` running PostgreSQL with
+database `db` and user with credentials `user:pass`. Port 5432 will be exposed
+and mapped to the host. If you want your database to be persistent across container
+executions, also add a `-v /host/db/path:/var/lib/pgsql/data` argument. This will be
+the PostgreSQL database cluster directory.
+
+If the database cluster directory is not initialized, the entrypoint script will
+first run [`initdb`](http://www.postgresql.org/docs/9.4/static/app-initdb.html)
+and setup necessary database users and passwords. After the database is initialized,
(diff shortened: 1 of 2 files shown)


ficap:
also original version used as example postgres openshift/postgresql-92-centos7 i replaced it with centos/postgresql-95-centos7 is it correct?

ficap:
Does postgres really start with PID 1? Because my docker top postgresql_database shows PID 1972.

pkubatrh:
>openshift/postgresql-92-centos7 i replaced it with centos/postgresql-95-centos7 is it correct?

Probably a leftover from when we copied the `README.md` when 9.5 got added. It's not really a major issue to be using 9.2 in examples but thanks for correcting this.

>Does postgres really start with PID 1? Because my docker top postgresql_database shows PID 1972.

It will not be PID 1 on the host (that is why `docker top` shows you a different number) but inside the running container the `postgres` process will be running as PID 1 due to the container having its own process namespace. Try `docker exec postgresql_database ps ax`

9.5/root/usr/share/container-scripts/postgresql/README.md
```diff
@@ -96,3 +113,19 @@ values stored in the variables and the actual passwords. Whenever a database
 container starts it will reset the passwords to the values stored in the
 environment variables.
 
+
+Troubleshooting
+---------------
+At first the postgres deamon in the container logs to the standard output, so the log is available in the container log. The log can be examined by running:
```
pkubatrh: Seems to be missing a verb over here. Possibly something along the lines of "At first the postgres daemon writes its logs to the standard output ..."?

pkubatrh: Also typo in deamon -> daemon

pkubatrh:
It seems like the first heading of `README.md` gets used as the title line for man/groff headers/footers, which does look a bit weird:

```
PostgreSQL(9.5)                     Server                     PostgreSQL(9.5)

       This  container  image  includes PostgreSQL 9.5 SQL database server for
       OpenShift and general usage.  Users can choose between RHEL and  CentOS
       based  images.   The  RHEL  image  is available in the Red Hat Registry
       <https://access.redhat.com/containers>     as      registry.access.red-
       hat.com/rhscl/postgresql-95-rhel7.   The CentOS image is then available
       on Docker Hub  <https://hub.docker.com/r/centos/postgresql-95-centos7/>
       as centos/postgresql-95-centos7.
                                      ...
Database                              SQL                      PostgreSQL(9.5)
```

Maybe we should add at least some metadata after all to make this look a bit more appealing? WDYT @hhorak @ficap

hhorak:
@pkubatrh yeah, the beginning of the man page doesn't look very pretty, but is there something we can do about it in the README.md file? I think the place this should be fixed is either atomic-reactor, or go-md2man, which might either receive slightly adjusted input or some more metadata.

hhorak:
One another suggestion -- I think we should use RHEL-based image and also include link to the container catalog (it's possible to use direct links to the particular image), same as it is done in this MySQL's pull-request: https://github.com/sclorg/mysql-container/pull/188

pkubatrh:
`go-md2man` is able to parse some metadata from the `README.md`. If I understood the code correctly, `go-md2man` parses any lines at the start of the file beginning with a `%` symbol and uses their contents for title line of the `help.1` file.

So when you add something like this to the beginning of the `README.md` file

```
% IMAGE_NAME (1) Container Image Pages
% MAINTAINER
% DATE
```

`help.1` will look like this:
```
IMAGE_NAME (1)                       DATE                       IMAGE_NAME (1)

PostgreSQL 9.5 SQL Database Server Docker image
       This container image includes PostgreSQL 9.5 SQL database server for
       OpenShift and general usage.  Users can choose between RHEL and CentOS
       based images.  The RHEL image is available in the Red Hat Registry
       <https://access.redhat.com/containers> as
       registry.access.redhat.com/rhscl/postgresql-95-rhel7.  The CentOS image
       is then available on Docker Hub
       <https://hub.docker.com/r/centos/postgresql-95-centos7/> as
       centos/postgresql-95-centos7.
                     
MAINTAINER                   Container Image Pages              IMAGE_NAME (1)
```

However (afaict) the lines beginning with the `%` sign would still show up in the README on github...

>One another suggestion -- I think we should use RHEL-based image and also include link to the container catalog

+1, makes sense.

9.5/root/usr/share/container-scripts/postgresql/README.md
ficap: You're right. I'll fix it asap.

ficap:
So what's the result... 
use 
``
% IMAGE_NAME (1) Container Image Pages``
``% MAINTAINER``
``% DATE``
 
in README.md or not?

pkubatrh:
I think it is worth it to put it in. It would look better on github if the metadata were placed inside a comment, but that would need changing `go-md2man` (might be a good feature request though).

pkubatrh:
@hhorak @praiskup Are you guys ok with including the manpage meta-data into the README? You can take a look on how the proposed changes look over here:
https://github.com/ficap/postgresql-container/blob/new-readme-format/9.4/root/usr/share/container-scripts/postgresql/README.md

@ficap Just a tip for future PRs: When you are moving stuff around and editing it, it is easier on the reviewer if you do those two steps in different commits since github can then highlight the specific changes made. The commits can then be easily squashed before merging.

pkubatrh:
Lets go with the version that has the header included.

@ficap please rebase to latest master so that we can merge this in.

pkubatrh:
lgtm, merging

PR #182, opened 2017-08-10
Add postgresql version 9.6

pkubatrh:
Uses https://github.com/sclorg/container-common-scripts/pull/20 . Should not be merged until this PR is merged.

Packages for 9.6 not yet available so the build has been disabled for this version for now.


pkubatrh:
Pulled in newest container-common-scripts.

praiskup:
We can't build 9.6 image at all, which is expected probably.  Otherwise it looks good to me.

praiskup:
Ah, `git grep 95 latest` and `git grep 9.5 latest` don't give me empty output yet.

praiskup:
What about to have `TARGET=f26_copr` for `Dockerfile.f26_copr`; and install packages from https://copr.fedorainfracloud.org/coprs/g/db-sig/sclo-postgresql96 ?

pkubatrh:
>Ah, git grep 95 latest and git grep 9.5 latest don't give me empty output yet.

Right, since the README was changed after the PR was created. Will make sure it uses correct versions.

pkubatrh:
> edit: maybe rather TARGET=fedora_copr to allow us to have only one fedora dockerfile till we have some templating system on place ...

Might be better to just use the regular fedora target for this (for now), since we do not have a fedora Dockerfile for 9.6 yet anyway. And this way we would not need to make any additional changes to the common scripts until we know what kind of templating system to use in future.

praiskup:
Per @hhorak 's note, the rh-postgresql96 packages already live here:
https://cbs.centos.org/repos/sclo7-rh-postgresql96-rh-candidate/x86_64/os/

Perhaps we could enable the repository temporarily to have at least one container built.  @pkubatrh WDYT?

praiskup:
Without paying too much attention to the patch, this did the trick for me:
```patch
 RUN yum install -y centos-release-scl-rh && \
+    yum-config-manager --add-repo https://cbs.centos.org/repos/sclo7-rh-postgresql96-rh-candidate/x86_64/os/ && \
+    echo gpgcheck=0 >> /etc/yum.repos.d/cbs.centos.org_repos_sclo7-rh-postgresql96-rh-candidate_x86_64_os_.repo && \
     INSTALL_PKGS="rsync tar gettext bind-utils rh-postgresql96 rh-postgresql96-postgresql-contrib nss_wrapper" && \
```

praiskup:
Otherwise looks good now, thanks.

pkubatrh:
Good idea, thanks. I have enabled the centos7 build with testing rpms

pkubatrh:
[test]

praiskup:
LGTM

pkubatrh:
Tests are passing now, merging. Thanks for review

PR #186, opened 2017-09-26
Make sure the tests can be run even when isolated

pkubatrh:
Due to some recent changes in the test suite it can no longer be run when the versioned directories are isolated.
This PR tackles this issue by first renaming `hack` to `test` since `hack` at this point only contains scripts/libraries used for testing anyway and second by making some modifications to the existing symbolic links from versioned directories to provide all the scripts/libraries needed for isolated execution.


test/pagila.sh
```diff
@@ -5,7 +5,7 @@ set -e
 die() { echo "$*" >&2 ; exit 1; }
 
 test -z "$CID" && die "Please specify \$CID variable"
-test -d common || die "Please run me from git root directory"
+# test -d common || die "Please run me from git root directory"
```
pkubatrh: Is this needed just so that we do not download pagila multiple times?

pkubatrh:
@praiskup @hhorak PTAL

test/pagila.sh
praiskup: yes

test/test_lib
```diff
@@ -13,7 +13,8 @@ get_image_id ()
     set -- $1
     case $2 in
         local)
-            echo "$(cat "$1"/.image-id)"
+            # Default to $IMAGE_NAME if it is set since .image-id might not exist
```
praiskup: I miss something here :).  How this can happen?

praiskup:
LGTM, thanks!

test/test_lib
praiskup: per irc chat, this is all about dist-git

PR #187, opened 2017-10-04
rhel7: Build 9.6 with beta packages

pkubatrh:
set version label to 1, remove release label completely

@hhorak is the above safe to do in CentOS Dockerfiles as well? So that we can have as similar Dockerfiles as possible.

### DIFF
--- a/latest/Dockerfile.rhel7
+++ b/latest/Dockerfile.rhel7
@@ -28,8 +28,7 @@ LABEL summary=$SUMMARY \
       io.openshift.tags="database,postgresql,postgresql96,rh-postgresql96" \
       name="rhscl/postgresql-96-rhel7" \
       com.redhat.component="rh-postgresql96-docker" \
-      version="9.6" \
-      release="1"
+      version="1"
 
 EXPOSE 5432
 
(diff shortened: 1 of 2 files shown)


praiskup:
LGTM if testsuite passes.

praiskup:
I suppose the versioning change deserves some deeper background story, but no stress here ...

pkubatrh:
Ah right, migration test for 9.6 fails as it tries to pull the 9.6 image remotely (to test remote:9.6 -> local:9.6 migration). I guess the correct solution would be to fallback to rhscl-beta images if we do not find a GA image.

praiskup:
`remote:9.6 -> local:9.6` testing for RHEL7 is not that necessary, so I'm not even against disabling this part for RHEL7

pkubatrh:
We might need to do that anyway (at least for now) as the upgrade/migration code is not in the beta image yet...

pkubatrh:
Also for future CentOS images this would fail since they are not built and available in the registry until the PR is merged in afaik. Given this I have removed the test for migration from $VERSION:remote to $VERSION:local completely.

praiskup:
> I have removed the test for migration from $VERSION:remote to $VERSION:local completely

+1, thanks, LGTM

pkubatrh:
Thanks for review. Tests are green now so merging this in.

PR #192, opened 2017-10-10
Add templates

praiskup:
[test-only]


praiskup:
Killed, my guess is enomem, but dunno.

PR #193, opened 2017-10-18
Use distgen for generating Dockerfiles

dhodovsk:
In this PR it is shown how [distgen] (https://github.com/devexp-db/distgen) can be used to generate Dockerfiles for multiple versions and distributions. We tried to minimize the differences between actual and generated Dockerfiles.

WIP: We still need to figure out how to prevent changes in generated Dockerfiles - e.g warn user when they try to push commits on generated Dockerfiles without specs modified.

### DIFF
--- a/9.4/Dockerfile.rhel7
+++ b/9.4/Dockerfile.rhel7
@@ -45,7 +45,7 @@ RUN yum install -y yum-utils gettext && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
-    INSTALL_PKGS="rsync tar gettext bind-utils rh-postgresql94 rh-postgresql94-postgresql-contrib nss_wrapper postgresql92-postgresql-server rh-postgresql94-postgresql-upgrade" && \
+    INSTALL_PKGS="rsync tar gettext bind-utils nss_wrapper rh-postgresql94 rh-postgresql94-postgresql-contrib postgresql92-postgresql-server rh-postgresql94-postgresql-upgrade" && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all && \
--- a/9.5/Dockerfile.rhel7
+++ b/9.5/Dockerfile.rhel7
@@ -45,7 +45,7 @@ RUN yum install -y yum-utils gettext && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
-    INSTALL_PKGS="rsync tar gettext bind-utils rh-postgresql95 rh-postgresql95-postgresql-contrib nss_wrapper rh-postgresql94-postgresql-server" && \
+    INSTALL_PKGS="rsync tar gettext bind-utils nss_wrapper rh-postgresql95 rh-postgresql95-postgresql-contrib rh-postgresql94-postgresql-server" && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all && \
--- a/Makefile
+++ b/Makefile
@@ -12,3 +12,12 @@ include common/common.mk
 # Additional upgrade tests.  Not hooked into CI ATM.
 upgrade-tests: $(VERSIONS)
 	test/run_upgrade_test 9.2:remote 9.4:local 9.5:local 9.6:local
+
+DG = /usr/bin/dg
+DG_EXEC = ${DG} --max-passes 25 --template Dockerfile.template --multispec specs/multispec.yml
+
+generate_dockerfiles:
+	for version in ${VERSIONS}; do \
+		${DG_EXEC} --distro centos-7-x86_64.yaml --multispec-selector version=$${version} --output $${version}/Dockerfile; \
+		${DG_EXEC} --distro rhel-7-x86_64.yaml --multispec-selector version=$${version} --output $${version}/Dockerfile.rhel7; \
+	done
--- a/latest/Dockerfile
+++ b/latest/Dockerfile
@@ -42,7 +42,7 @@ COPY root/usr/libexec/fix-permissions /usr/libexec/fix-permissions
 RUN yum install -y centos-release-scl-rh && \
     yum-config-manager --add-repo https://cbs.centos.org/repos/sclo7-rh-postgresql96-rh-candidate/x86_64/os/ && \
     echo gpgcheck=0 >> /etc/yum.repos.d/cbs.centos.org_repos_sclo7-rh-postgresql96-rh-candidate_x86_64_os_.repo && \
-    INSTALL_PKGS="rsync tar gettext bind-utils rh-postgresql96 rh-postgresql96-postgresql-contrib nss_wrapper rh-postgresql95-postgresql-server" && \
+    INSTALL_PKGS="rsync tar gettext bind-utils nss_wrapper rh-postgresql96 rh-postgresql96-postgresql-contrib rh-postgresql95-postgresql-server" && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all && \
--- a/latest/Dockerfile.rhel7
+++ b/latest/Dockerfile.rhel7
@@ -44,8 +44,9 @@ RUN yum install -y yum-utils gettext && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
-    INSTALL_PKGS="rsync tar gettext bind-utils rh-postgresql96 rh-postgresql96-postgresql-contrib nss_wrapper rh-postgresql95-postgresql-server" && \
-    yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
+
+    INSTALL_PKGS="rsync tar gettext bind-utils nss_wrapper rh-postgresql96 rh-postgresql96-postgresql-contrib rh-postgresql95-postgresql-server" && \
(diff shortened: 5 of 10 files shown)


bkabrda:
This looks reasonable to me; it's a small enough change that it's easy to understand and doesn't break anything and allows further improvements in future. My advice would be to completely delete the generated Dockerfiles from the repo, but that's really up to the repo maintainers.

9.4/Dockerfile.rhel7
```diff
@@ -29,7 +29,8 @@ LABEL summary=$SUMMARY \
       name="rhscl/postgresql-94-rhel7" \
       com.redhat.component="rh-postgresql94-docker" \
       version="9.4" \
-      release="1"
+      release="1" \
+      maintainer="SoftwareCollections.org <sclorg@redhat.com>"
```
praiskup: This is not truth for RHEL images.

9.5/Dockerfile.rhel7
```diff
@@ -1,4 +1,4 @@
-FROM rhel7
+FROM registry.access.redhat.com/rhel7
```
praiskup: I'm not sure whether we can do this, @hhorak/@pkubat?

praiskup:
Seems to be reasonable first step --> my workflow would be to maintain the template(s) and hit some maintainers's make target once I'm done. I don't think `make dg` is nice make target spelling, but that's nit.

praiskup:
Note that you need to have `distgen v0.17` from fc26 updates-testing.

9.5/Dockerfile.rhel7
pkubatrh: Might cause some issues as this will always try to pull the image from the registry (or expect something tagged like this locally), which likely is not what is going on in OSBS.

For postgresql this is ok, since it will get replaced by the specific base image but I would rather gave the same format in every image.

pkubatrh:
Missing postgresql 9.6

Makefile
```diff
@@ -12,3 +12,12 @@ include common/common.mk
...
 	test/run_upgrade_test 9.2:remote 9.4:local 9.5:local 9.6:local
+
+DG = /usr/bin/dg
+DG_EXEC = ${DG} --max-passes 25 --template Dockerfile.template --multispec specs/multispec.yml
+
+dg:
+	${DG_EXEC} --distro centos-7-x86_64.yaml --multispec-selector version="9.4" --output 9.4/Dockerfile
+	${DG_EXEC} --distro centos-7-x86_64.yaml --multispec-selector version="9.5" --output 9.5/Dockerfile
+	${DG_EXEC} --distro rhel-7.3-x86_64.yaml --multispec-selector version="9.4" --output 9.4/Dockerfile.rhel7
+	${DG_EXEC} --distro rhel-7.3-x86_64.yaml --multispec-selector version="9.5" --output 9.5/Dockerfile.rhel7
```
praiskup: Shouldn't we have 'rhel-7-x86_64.yaml' config, or some other magic doing "The Right Thing"?

latest/Dockerfile.rhel7
```diff
@@ -46,11 +47,11 @@ RUN yum repolist > /dev/null && \
     yum install -y yum-utils gettext && \
     yum-config-manager --disable \* &> /dev/null && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
-    yum-config-manager --enable rhel-server-rhscl-7-beta-rpms && \
     yum-config-manager --enable rhel-7-server-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
-    INSTALL_PKGS="rsync tar gettext bind-utils rh-postgresql96 rh-postgresql96-postgresql-contrib nss_wrapper rh-postgresql95-postgresql-server" && \
-    yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
+    yum-config-manager --enable rhel-server-rhscl-7-beta-rpms && \
```
dhodovsk: I didn't notice this error, will try to figure out a better way.

Makefile
dhodovsk: You mean one of [these](https://github.com/devexp-db/distgen/tree/master/distconf) or would you like to see a different approach for obtaining distribution specific information?

praiskup:
[test]

latest/Dockerfile.rhel7
```diff
@@ -28,7 +28,8 @@ LABEL summary=$SUMMARY \
       io.openshift.tags="database,postgresql,postgresql96,rh-postgresql96" \
       name="rhscl/postgresql-96-rhel7" \
       com.redhat.component="rh-postgresql96-docker" \
-      version="1"
+      version="9.6" \
+      release="1"
```
pkubatrh: We are not setting release in newer images anymore as it is updated by the build system.

latest/Dockerfile
```diff
@@ -39,10 +39,10 @@ COPY root/usr/libexec/fix-permissions /usr/libexec/fix-permissions
 # This image must forever use UID 26 for postgres user so our volumes are
 # safe in the future. This should *never* change, the last test is there
 # to make sure of that.
-RUN yum install -y centos-release-scl-rh && \
+RUN yum install -y centos-release-scl && \
```
pkubatrh: What are the differences between these two repositories? @hhorak

latest/Dockerfile.rhel7
pkubatrh: Also using version="1" for all new images, since the version of postgresql is already incorporated into the image name

dhodovsk: Could you please specify which images are we discussing? Are they new release of old images? Could it be safe to change it in all postgresql version images or is this only related to this(postgresql 9.6 rhel) image?

pkubatrh: I meant new images introduced in RHSCL 3.0 (and later) - so for this repository that would be only postgresql 9.6.

Changing the version and release for older images would likely not introduce any major issues but it would be confusing to see the version drop to 1 so we decided to leave the older images as is.

latest/Dockerfile
omron93: `centos-release-scl` allows also community developed collections and packages - `sclo-ror42`, `sclo-subversion19`, `sclo-vagrant1`,...

Makefile
```diff
@@ -12,3 +12,14 @@ include common/common.mk
...
+DG = /usr/bin/dg
+DG_EXEC = ${DG} --max-passes 25 --template Dockerfile.template --multispec specs/multispec.yml
+
+generate_dockerfiles:
+	${DG_EXEC} --distro centos-7-x86_64.yaml --multispec-selector version="9.4" --output 9.4/Dockerfile
+	${DG_EXEC} --distro centos-7-x86_64.yaml --multispec-selector version="9.5" --output 9.5/Dockerfile
+	${DG_EXEC} --distro centos-7-x86_64.yaml --multispec-selector version="9.6" --output 9.6/Dockerfile
+	${DG_EXEC} --distro rhel-7-x86_64.yaml --multispec-selector version="9.4" --output 9.4/Dockerfile.rhel7
+	${DG_EXEC} --distro rhel-7-x86_64.yaml --multispec-selector version="9.5" --output 9.5/Dockerfile.rhel7
+	${DG_EXEC} --distro rhel-7-x86_64.yaml --multispec-selector version="9.6" --output 9.6/Dockerfile.rhel7
```
praiskup: @dhodovsk wrote (dunno where the comment disappeared):
> You mean one of [these](https://github.com/devexp-db/distgen/tree/master/distconf) or would you like to see a different approach for obtaining distribution specific information?

You switched from rhel-7.3-x86_64 to rhel-7-x86_64, so we don't have to bump the version all the time, looks better.

Though here I would go with `for version in $(VERSIONS); do .... done` loop.

latest/Dockerfile.rhel7
```diff
@@ -46,11 +46,11 @@ RUN yum repolist > /dev/null && \
     yum install -y yum-utils gettext && \
     yum-config-manager --disable \* &> /dev/null && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
-    yum-config-manager --enable rhel-server-rhscl-7-beta-rpms && \
     yum-config-manager --enable rhel-7-server-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
-    INSTALL_PKGS="rsync tar gettext bind-utils rh-postgresql96 rh-postgresql96-postgresql-contrib nss_wrapper rh-postgresql95-postgresql-server" && \
-    yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
+    yum-config-manager --enable rhel-server-rhscl-7-beta-rpms && \
```
praiskup: I'm afraid this is not going to work (missing backslash).

praiskup: Sorry, false alarm -> I mis-interpreted the github's diff format.

praiskup:
[test]

praiskup:
Btw., the rhel7 failure is caused by inconsistent rhscl beta channel;  so it is not blocking this, IMO.

latest/Dockerfile
hhorak: right, and I think we should stay with `centos-release-scl-rh`, to be sure we don't install anything from the other set..

praiskup:
[test]

omron93:
[test]

@dhodovsk To be able to trigger CI tests, you have to be public member of github sclorg/ organization. I've sent you an invitation.

dhodovsk:
[test]

pkubatrh:
[test]

pkubatrh:
@dhodovsk You still need to set your sclorg membership to public in order for the centos CI to pickup your test command.

praiskup:
The test failure needs rebase against #202.

dhodovsk:
[test]

Makefile
praiskup: Ping on this.  I often do 'make VERSIONS=9.6' only, for this target it would be useful, too.

praiskup:
As I said, I'm basically +1 for this PR as a first step, and I'm not even against removing the generated stuff.  @dhodovsk you need to convince some committer, though, too.

pkubatrh:
>I'm not even against removing the generated stuff

If we are aiming to remove the generated Dockerfiles/READMEs from github we should also add a dependency to `make build` so that all the files are generated before trying to run a build.

praiskup:
Yeah, I just stated my opinion;  I don't claim that this PR in particular aims to solve that, nor that it should.

pkubatrh:
Merged in the current state. We can always build on top of it if something turns out to not be the correct way to do this.

PR #185, opened 2017-09-09
Do not use tables in README.md

hhorak:
The reasons to not use tables in man page are:
* tables are hard to keep readable in plaintext
* tables don't look nice when rendered and long variables or complicated default value are used
* go-md2man does not convert tables into man page format in versions that are mostly used

@praiskup @pkubatrh FYI, WDYT?

PR #203, opened 2017-11-02
More distgen - directory generation

dhodovsk:
In this PR we added ability to generate directories 9.4, 9.5 and 9.6 using distgen.

Directories are generated using ./src directory and with command `make generate`


9.4/root/usr/share/container-scripts/postgresql/README.md
```diff
@@ -51,49 +51,49 @@ Environment variables and volumes
 The image recognizes the following environment variables that you can set during
 initialization by passing `-e VAR=VALUE` to the Docker run command.
 
-**`POSTGRESQL_USER`**
```
pkubatrh: These trailing spaces are actually left there on purpose.

See: https://github.com/sclorg/postgresql-container/pull/185#issuecomment-329566358

gen-dirs.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+set -e
+cd templates
+FILES=$(find -follow)
+VERSIONS="9.4 9.5 9.6"
```
pkubatrh: VERSIONS should get passed to the script from the Makefile, so that the list can be maintained in one place.

pkubatrh:
Since we agreed to remove the generated files in future we should make sure that we do not lose all of the git history by moving latest into template in a separate commit in this PR.

pkubatrh:
(not sure if we even need `latest` after moving it to `template`)

9.4/root/usr/share/container-scripts/postgresql/README.md
praiskup: Not only that, white-space changes just cause rush in pull-requests.  Please consider fixing your editor to not remove trailing white-spaces automatically.

pkubatrh:
>moving latest into template in a separate commit in this PR

To have an idea what I mean by this please check how `latest/` was introduced: https://github.com/sclorg/postgresql-container/pull/170

dhodovsk:
We might also want to preserve Dockerfile history therefore there are now two more commits to make that happen.

Makefile
```diff
@@ -2,6 +2,7 @@
 BASE_IMAGE_NAME = postgresql
 VERSIONS = 9.4 9.5 9.6
 OPENSHIFT_NAMESPACES = 9.2
+export VERSIONS
```
pkubatrh: I am a bit scared exporting VERSIONS might break something in the existing scripts. It would be better if you just kept the versions variable local for the `gen-dirs.sh` script. So something like:

```
VERSIONS="$(VERSIONS)" ./gen-dirs.sh
```

pkubatrh:
>We might also want to preserve Dockerfile history therefore there are now two more commits to make that happen.

That is true, thanks for picking up on this!

Makefile
```diff
@@ -13,11 +13,5 @@ include common/common.mk
...
-DG = /usr/bin/dg
-DG_EXEC = ${DG} --max-passes 25 --template Dockerfile.template --multispec specs/multispec.yml
-
-generate_dockerfiles:
-	for version in ${VERSIONS}; do \
-		${DG_EXEC} --distro centos-7-x86_64.yaml --multispec-selector version=$${version} --output $${version}/Dockerfile; \
-		${DG_EXEC} --distro rhel-7-x86_64.yaml --multispec-selector version=$${version} --output $${version}/Dockerfile.rhel7; \
-	done
+distgen:
+	VERSIONS="$(VERSIONS)" ./gen-dirs.sh
```
praiskup: Please use make targets instead of this script, and please develop them in `container-common-scripts` project ...

Running this script takes ~12s on my box, because distgen is pretty slow TBH:

    $ make distgen
    real    0m11.999s
    user    0m11.124s
    sys     0m0.827s

(at least at this point).  `make` should help with building it in parallel.

gen-dirs.sh
```diff
@@ -0,0 +1,80 @@
...
+    mkdir ../$1
+    echo "Generating Dockerfile: $1/Dockerfile"
+    /usr/bin/dg --max-passes 25 --template ${PWD}/../Dockerfile.template \
+        --multispec ../specs/multispec.yml --distro centos-7-x86_64.yaml \
+        --multispec-selector version=$1 --output ../$1/Dockerfile
+    echo "Generating Dockerfile: $1/Dockerfile.rhel7"
+    /usr/bin/dg --max-passes 25 --template ${PWD}/../Dockerfile.template \
+        --multispec ../specs/multispec.yml --distro rhel-7-x86_64.yaml \
+        --multispec-selector version=$1 --output ../$1/Dockerfile.rhel7
+
```
praiskup: I'm not sure I follow right, but I think the goal is to remove the generated products from this repo, right?

I would really prefer to not list centos/rhel dockerfiles here, and anywhere else.  We'll have fedora dockerfiles soon, and maybe scls+fedora? Who knows..  So it would be nice if, when generating Dockerfile, the $OS variable was respected.

gen-dirs.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+set -e
+cd templates
+FILES=$(find -follow)
```
praiskup: Hm, I dislike this, tbh.  Especially that we filter everything through the build script, while simple copy would often be enough.

Maybe we could separate the real templates into ./templates, and normal files under ./files?

Other approach could be using some "manifest file", I implemented something like this in https://github.com/devexp-db/cont-postgresql/blob/master/cl-manifest
That file declars what should be copied, what should be "instantiated" (and by what tool) and everything explicitly (not copying some accidentally occurred files). This should stay executed by postgresql "in parallel" if implemented.

gen-dirs.sh
```diff
@@ -0,0 +1,80 @@
...
+
+generate_file() {
+    if [ -L $1 ]; then
+        echo "Skipping symbolic link: $1"
+        return
+    elif [[ -d $1 && ! -d ../${version}/$1 ]]; then
+        mkdir ../${version}/$1
+    else
+        echo "Generating: ${version}/$1"
+        /usr/bin/dg --max-passes 1 --multispec ../specs/multispec.yml \
```
praiskup: Once this is moved to Makefile (common.mk probably), please ensure that instead of /usr/bin/dg you are using `$(DG)` variable; to allow anybody to replace the distgen version from local installation (virtualenv).

Makefile
pkubatrh: I agree. Adding the rules to `container-common-scrtips` will make the distgen change much easier to adopt in other images if the maintainers choose to do so.

dhodovsk:
Blocked by https://github.com/sclorg/container-common-scripts/pull/38

manifest.sh
```diff
@@ -0,0 +1,66 @@
+# Manifest for image directories creation
+# every dest and link_name path will be prefixed by $DEST_DIR/$version
+ 
+DEST_DIR='' # optional, defaults to $PWD
```
praiskup: s/DEST_DIR/DESTDIR/ -> that's far more standard variable

manifest.sh
```diff
@@ -0,0 +1,66 @@
...
+DISTGEN_RULES="
+    src=templates/Dockerfile
+    dest=Dockerfile;
+
+    src=templates/root/usr/share/container-scripts/postgresql/README.md
+    dest=root/usr/share/container-scripts/postgresql/README.md;
+
+    src=templates/root/usr/share/container-scripts/postgresql/common.sh
+    dest=root/usr/share/container-scripts/postgresql/common.sh
+    mode=0755
```
praiskup: This doesn't have to be executable.

manifest.sh
```diff
@@ -0,0 +1,66 @@
...
+"
+
+# Symbolic links
+SYMLINK_RULES="
+    link_target=root/usr/share/container-scripts/postgresql/README.md
+    link_name=README.md;
+
+    link_target=../test
+    link_name=test
+"
```
praiskup: This seems to create absolute symlinks on the host system during build.  Such simlinks - if copied into image - will lead to non-existing paths.

manifest.sh
```diff
@@ -0,0 +1,66 @@
...
+    src=templates/root/usr/share/container-scripts/postgresql/openshift-custom-recovery.conf.template
+    dest=root/usr/share/container-scripts/postgresql/openshift-custom-recovery.conf.template;
+
+    src=templates/root/usr/bin/cgroup-limits
+    dest=root/usr/bin/cgroup-limits
+    mode=0755;
+
+    src=templates/root/usr/share/container-scripts/postgresql/scl_enable
+    dest=root/usr/share/container-scripts/postgresql/scl_enable
+    mode=0755;
```
praiskup: I don't think this needs to be executable.

praiskup:
- I really like the speed of the generated output
- I love the fact that we can adjust permissions of particular files! (the current approach doesn't ensure the proper permissions, but the following point should solve it...)
- I miss dealing with Dockerfile.rhel7 - it is not re-created if I drop it (probably we should drop everything generated from the source directory)
- ccpp.yml is not geneerated
- I would not use the `templates/` directory, but rather `source/` or `src/` (not everything under that is a jinja/distgen template)

For me, I'm leaning towards +1, but not without another positive votes.

phracek:
The PR looks good to me. Good job @dhodovsk .
I am fine with merging. I would prefer to ACK it and create another PR for rest issues.

WDYT?

TomasTomecek:
Wow, I'm really impressed how far distgen got. The manifest is some really good work (I would personally prefer json/yaml + python, but if the proposed implementation suits your needs, that's just fine).

I also agree with Pavel the time will tell. Hence I'm +1 on merging and addressing issues as they come.

dhodovsk:
The requested changes should be now applied.

I'd like to discuss the possibility to drop directories that can be generated. I think we should do it to prevent the case someone commits changes in generated files (and not src). But it seems like a big step and we have to make sure it is safe.

praiskup:
It is actually not yet ready for the merge (see my comment above for the remaining stuff), but the agreement needs to be done by @pkubatrh @hhorak (the actual committers), and maybe others. I planned this PR for discussion for tomorrow container meeting (@TomasTomecek @dhodovsk @phracek feel free to come).

pkubatrh:
>I miss dealing with Dockerfile.rhel7 - it is not re-created if I drop it (probably we should drop everything generated from the source directory)

I think this is the only issue that is yet unsolved and I would like to have `Dockerfile.rhel7` as well. It would also be necessary in case we decide to have the generated files available somewhere (in a separate branch?) since now it just defaults to Dockerfile (without the extension).

Also if I generate the centos version of the Dockerfile and then try to generate a rhel7 version (via `make TARGET=rhel7 generate`) the Dockerfile does not get overwritten.

pkubatrh:
I believe Fedora should also be in the multispec, since we have a Dockerfile available for 9.6

praiskup:
The `make generate-all` generates:
```patch
diff --git a/9.4/Dockerfile.rhel7 b/9.4/Dockerfile.rhel7
index dd28dc8..be5558a 100644
--- a/9.4/Dockerfile.rhel7
+++ b/9.4/Dockerfile.rhel7
@@ -30,7 +30,7 @@ LABEL summary=$SUMMARY \
       name="rhscl/postgresql-94-rhel7" \
       com.redhat.component="rh-postgresql94-docker" \
       version="9.4" \
-      usage="docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 rhscl/postgresql-94-rhel7" \
+      usage="docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 centos/postgresql-96-centos7" \
       maintainer="SoftwareCollections.org <sclorg@redhat.com>"
 
 EXPOSE 5432
diff --git a/9.5/Dockerfile.rhel7 b/9.5/Dockerfile.rhel7
index d3119c3..0f32887 100644
--- a/9.5/Dockerfile.rhel7
+++ b/9.5/Dockerfile.rhel7
@@ -30,7 +30,7 @@ LABEL summary=$SUMMARY \
       name="rhscl/postgresql-95-rhel7" \
       com.redhat.component="rh-postgresql95-docker" \
       version="9.5" \
-      usage="docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 rhscl/postgresql-95-rhel7" \
+      usage="docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 centos/postgresql-96-centos7" \
       maintainer="SoftwareCollections.org <sclorg@redhat.com>"
 
 EXPOSE 5432
diff --git a/latest/Dockerfile.rhel7 b/latest/Dockerfile.rhel7
index 0934b6d..3e7d4f9 100644
--- a/latest/Dockerfile.rhel7
+++ b/latest/Dockerfile.rhel7
@@ -30,7 +30,7 @@ LABEL summary=$SUMMARY \
       name="rhscl/postgresql-96-rhel7" \
       com.redhat.component="rh-postgresql96-docker" \
       version="1" \
-      usage="docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 rhscl/postgresql-96-rhel7" \
+      usage="docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 centos/postgresql-96-centos7" \
       maintainer="SoftwareCollections.org <sclorg@redhat.com>"
 
 EXPOSE 5432

```

praiskup:
Ok, I probably missed the point of `generate-all`, dunno.

Btw. before merging this, I'd really love to see the CI set-up that the generator does
it's job correctly.  E.g. could we add check that `make generate` generates no changes?

pkubatrh:
Why is there a `Dockerfile.rhel7` in `src/`? I see why the fedora one would be present, as we are not yet generating it (should we?) but having the rhel7 version as well seems superfluous.

src/Dockerfile
```diff
@@ -34,7 +34,7 @@ LABEL summary=$SUMMARY \
 {%- else %}
       version="{{ spec.version }}" \
 {%- endif %}
-      usage="docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 {{ spec.org }}/postgresql-{{ spec.short }}-{{ spec.prod }}" \
+      usage="docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 centos/postgresql-96-centos7" \
```
pkubatrh: Why this change?

dhodovsk: This was a mistake I created while rebasing, it should be OK now.

dhodovsk:
You are right, there is no reason having Dockerfile.rhel in src. That file appeared there when saving git history for files in src. Now it is removed. Thank you for catching the issues, @pkubatrh @praiskup !

praiskup:
>  E.g. could we add check that make generate generates no changes?

This ^^^ seems to be really needed if we meant this PR real :-) or (maybe) drop all the pre-generated stuff just now?

pkubatrh:
>or (maybe) drop all the pre-generated stuff just now?

+1, if we drop them now we will also get to keep the git history in `src/` without any additional hassle

dhodovsk:
What about version 9.2? It isn't generated using distgen, since it has been deprecated. Can it be also dropped?

pkubatrh:
Common scripts need to get update to be able to use the new distgen functions and the CI needs to be updated to run said functions to make the tests green.

Otherwise lgtm.

pkubatrh:
@praiskup I meant the common script submodule inside the PR needs to get updated to the latest master. I dont think there is a need for additional changes to the common scripts.

pkubatrh:
Now we only need to make sure the CI are able to generate the old layout properly.

@praiskup @omron93 What would be the best way to achieve this? I dont really want to make `generate` a dependency of `build` rule since that would force distgen on everyone. How about testing (in the CI) for the existence of the manifest and running `make generate` if it is present?

praiskup:
> How about testing (in the CI) for the existence of the manifest and running make generate if it is present?

CI can simply run the `make generate` and it can by no-op by default.

pkubatrh:
>CI can simply run the make generate and it can by no-op by default.

Currently it is a failed op as the generate rule cannot satisfy its dependency on the manifest file.

praiskup:
I see, thanks.  So we probably need the following snippet in postgresql's Makefile, to trigger the dependancy:

```patch
+
+ build: generate
```

praiskup:
Ah, seems like there's race ... because with `build: generate` we'll run `generate` at the same time with `build-serial`... which will fail (usually).  So I would propose workaround with `build-serial: generate`.

omron93:
`/usr/bin/dg: No such file or directory` - Should this be installed in CI? What is best way for RHEL/CentOS?

pkubatrh:
>What is best way for RHEL/CentOS?

I believe it should be available in epel

praiskup:
per @hhorak note, cccp.yml should be fixed first before we drop the generated paths from git repo.

omron93:
[test][test-openshift]

omron93:
> I believe it should be available in epel

Distgen is being installed from EPEL.

pkubatrh:
>cccp.yml should be fixed first before we drop the generated paths from git repo.

So leave the existing code in for the time being and remove it in the future once the container pipeline is ready for it?

praiskup:
Probably, @TomasTomecek, how far the distgen support in cccp is?

TomasTomecek:
@praiskup this is what we use to build container images in cccp using distgen:

https://github.com/container-images/tools/blob/master/hooks/pre_build_centos
https://github.com/CentOS/container-index/blob/master/index.d/user-cont.yaml#L13

praiskup:
So it just works, right?  Only the cccp.yml needs to be fixed, cool!

pkubatrh:
In that case all should be ready to have this merged once the PR is rebased and cccp.yaml modified accordingly right?

@dhodovsk ping

praiskup:
That's how I understand that, yes.

dhodovsk:
I am now rebasing and adding following changes:
 - add https://github.com/container-images/tools/blob/master/hooks/pre_build_centos script to new directory: ./hooks in root of this repository
 - append line `prebuild-script: hooks/pre_build_centos` to `src/cccp.yaml`

Is that enough? @TomasTomecek @pkubatrh

TomasTomecek:
It should be. I don't know if the cccp.yaml file present in git repo overrides the one defined in cccp index git repo.

pkubatrh:
Is there a way to test things out with the current configuration before merging the changes?

pkubatrh:
Now that I look at it more closely. Will the prebuild script be even run if the yaml file in the container-index point to versioned directories?

If I understand this correctly cccp expects to find the cccp over there and will not look for it in `src/`, or anywhere else in the matter. And then cccp wont know it should first run the hook to generate the files.

hooks/pre_build_centos
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -ex
+yum install epel-release -y
+# yum update -y
+# our deps
+yum install distgen make -y
+
+mkdir -p root/
+make VARIANT=centos DISTRO=centos-7-x86_64 source
```
pkubatrh: Probably needs to be `make TARGET=centos7 generate`? We also do not need the `root/` directory creation.

dhodovsk:
The commit `Add prebuild-script to cccp.yml` is now reverted. Please, let me know if there is anything else I could do.

pkubatrh:
>The commit Add prebuild-script to cccp.yml is now reverted. Please, let me know if there is anything else I could do.

I think we need at least the hook part of the commit (the actual script I commented on), since the versioned directories will still need to be generated.

Its just that the configuration option to run it will need to be added elsewhere - similarly to https://github.com/container-images/tools

dhodovsk:
I am sorry, but I have no insight in functionality and importance of changes related to cccp, so I am not able to decide the right workflow, could you please provide more info of what changes exactly are needed so the requirements are met? Or maybe provide a commit with those changes?

pkubatrh:
I have added a commit with what I believe is needed for the cccp to work once changes are made in https://github.com/CentOS/container-index

Now the question is, do we merge this in first and make the changes in index later? The other option is to merge this PR in without the removal of versioned directories so that cccp works both with the old and with new configuration option (`prebuild-script`).

@hhorak @praiskup @omron93 WDYT?

omron93:
I would merge this PR and next fix the index.

pkubatrh:
[test-openshift]

praiskup:
@hhorak would it be OK to have the cccp outage temporarily?

pkubatrh:
Since Honza will not be able to comment on this for a while and everyone else is in favour of merging the PR and fixing the issues (and the tests are green) let us merge this in.

TomasTomecek:
Woooooooooh!

PR #205, opened 2017-11-08
Adding Dockerfile.fedora

ficap:


omron93:
@ficap Thanks.

LGTM. It also comply with Fedora guidelines - https://fedoraproject.org/wiki/Container:Guidelines
I see no reason why not to merge.

Issue #195, opened 2017-10-23
Templates have a buggy liveness probe 


pkubatrh:
Description copied from original bug report (https://bugzilla.redhat.com/show_bug.cgi?id=1474683)

Templates have the following livenessProbe included:

```
          livenessProbe:
            initialDelaySeconds: 30
            tcpSocket:
              port: 5432
            timeoutSeconds: 1
```

which lead to the following in the postgress log within the container every 30 sec:

```
LOG:  incomplete startup packet
LOG:  incomplete startup packet
LOG:  incomplete startup packet
```

Steps to Reproduce:

```
oc process -f /usr/share/ansible/openshift-ansible/roles/openshift_examples/files/examples/v1.5/db-templates/postgresql-ephemeral-template.json  -v POSTGRESQL_USER=jim -v POSTGRESQL_PASSWORD=redhat -v POSTGRESQL_DATABASE=testdb | oc create -f -
secret "postgresql" created
service "postgresql" created
deploymentconfig "postgresql" created
[root@inf150 ~]# oc get all
NAME            REVISION   DESIRED   CURRENT   TRIGGERED BY
dc/postgresql   1          1         1         config,image(postgresql:9.5)

NAME              DESIRED   CURRENT   READY     AGE
rc/postgresql-1   1         1         1         1m

NAME             CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
svc/postgresql   172.30.83.86   <none>        5432/TCP   1m

NAME                    READY     STATUS    RESTARTS   AGE
po/postgresql-1-2j6sl   1/1       Running   0          57s
[root@inf150 ~]# 
[root@inf150 ~]# oc rsh postgresql-1-2j6sl
sh-4.2$ tail -f /var/lib/pgsql/data/userdata/pg_log/postgresql-Tue.log 
LOG:  autovacuum launcher started
LOG:  incomplete startup packet
LOG:  incomplete startup packet
LOG:  incomplete startup packet
LOG:  incomplete startup packet
LOG:  incomplete startup packet
LOG:  incomplete startup packet
LOG:  incomplete startup packet
LOG:  incomplete startup packet
```

Actual results:
"LOG:  incomplete startup packet" every 30 seconds in /var/lib/pgsql/data/userdata/pg_log/postgresql-Tue.log 

Expected results:
Our template checks should not cause more logs

Additional info:
Having a look at 
[...]

PR #206, opened 2017-11-16
templates: use pg_isready as liveness probe

pkubatrh:
Resolves: #195

### DIFF
--- a/examples/postgresql-ephemeral-template.json
+++ b/examples/postgresql-ephemeral-template.json
@@ -5,7 +5,7 @@
     "name": "postgresql-ephemeral",
     "annotations": {
       "openshift.io/display-name": "PostgreSQL (Ephemeral)",
-      "description": "PostgreSQL database service, without persistent storage. For more information about using this template, including OpenShift considerations, see https://github.com/sclorg/postgresql-container/blob/master/9.5.\n\nWARNING: Any data stored will be lost upon pod destruction. Only use this template for testing",
+      "description": "PostgreSQL database service, without persistent storage. For more information about using this template, including OpenShift considerations, see https://github.com/sclorg/postgresql-container/.\n\nWARNING: Any data stored will be lost upon pod destruction. Only use this template for testing",
       "iconClass": "icon-postgresql",
       "tags": "database,postgresql",
       "template.openshift.io/long-description": "This template provides a standalone PostgreSQL server with a database created.  The database is not stored on persistent storage, so any restart of the service will result in all data being lost.  The database name, username, and password are chosen via parameters when provisioning this service.",
@@ -14,7 +14,7 @@
       "template.openshift.io/support-url": "https://access.redhat.com"
     }
   },
-  "message": "The following service(s) have been created in your project: ${DATABASE_SERVICE_NAME}.\n\n       Username: ${POSTGRESQL_USER}\n       Password: ${POSTGRESQL_PASSWORD}\n  Database Name: ${POSTGRESQL_DATABASE}\n Connection URL: postgresql://${DATABASE_SERVICE_NAME}:5432/\n\nFor more information about using this template, including OpenShift considerations, see https://github.com/sclorg/postgresql-container/blob/master/9.5.",
+  "message": "The following service(s) have been created in your project: ${DATABASE_SERVICE_NAME}.\n\n       Username: ${POSTGRESQL_USER}\n       Password: ${POSTGRESQL_PASSWORD}\n  Database Name: ${POSTGRESQL_DATABASE}\n Connection URL: postgresql://${DATABASE_SERVICE_NAME}:5432/\n\nFor more information about using this template, including OpenShift considerations, see https://github.com/sclorg/postgresql-container/.",
   "labels": {
     "template": "postgresql-ephemeral-template"
   },
@@ -245,8 +245,8 @@
     {
       "name": "POSTGRESQL_VERSION",
       "displayName": "Version of PostgreSQL Image",
-      "description": "Version of PostgreSQL image to be used (9.2, 9.4, 9.5 or latest).",
-      "value": "9.5",
+      "description": "Version of PostgreSQL image to be used (9.4, 9.5, 9.6 or latest).",
+      "value": "9.6",
       "required": true
     }
   ]
--- a/examples/postgresql-ephemeral-template.json
+++ b/examples/postgresql-ephemeral-template.json
@@ -127,11 +127,11 @@
                   }
                 },
                 "livenessProbe": {
+                  "exec": {
+                    "command": [ "/bin/sh", "-i", "-c", "pg_isready -h 127.0.0.1 -p 5432" ]
+                  },
                   "timeoutSeconds": 1,
-                  "initialDelaySeconds": 30,
-                  "tcpSocket": {
-                    "port": 5432
-                  }
+                  "initialDelaySeconds": 30
                 },
                 "env": [
                   {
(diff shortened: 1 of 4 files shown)


pkubatrh:
@praiskup Can you take a quick look at this?

@pvalena @bparees Is there anything more we need to do to get the change into the Openshift templates?

pvalena:
AFAIK you primarily need to add the new version `9.6`

- as a default for deployment; and
- in all texts, such as ` 9.2, 9.4, 9.5 or latest `

Thanks!

pkubatrh:
Makes sense, thanks.

examples/postgresql-ephemeral-template.json
```diff
@@ -245,8 +245,8 @@
     {
       "name": "POSTGRESQL_VERSION",
       "displayName": "Version of PostgreSQL Image",
-      "description": "Version of PostgreSQL image to be used (9.2, 9.4, 9.5 or latest).",
-      "value": "9.5",
+      "description": "Version of PostgreSQL image to be used (9.4, 9.5, 9.6 or latest).",
```
pvalena: Is 9.2 not relevant anymore?

pkubatrh: Has been deprecated for a long while already.

examples/postgresql-ephemeral-template.json
```diff
@@ -5,7 +5,7 @@
     "name": "postgresql-ephemeral",
     "annotations": {
       "openshift.io/display-name": "PostgreSQL (Ephemeral)",
-      "description": "PostgreSQL database service, without persistent storage. For more information about using this template, including OpenShift considerations, see https://github.com/sclorg/postgresql-container/blob/master/9.5.\n\nWARNING: Any data stored will be lost upon pod destruction. Only use this template for testing",
+      "description": "PostgreSQL database service, without persistent storage. For more information about using this template, including OpenShift considerations, see https://github.com/sclorg/postgresql-container/.\n\nWARNING: Any data stored will be lost upon pod destruction. Only use this template for testing",
```
pvalena: AFAIK this was meant to give container-specific instructions (real simplistic), thus reading `README.md` - usage - for direct usage of OpenShift; regardless of version(most ppl just use `latest`). IOW: I'd rather keep the 9.6.

pvalena: f.e. https://github.com/sclorg/postgresql-container/tree/master/9.5#environment-variables-and-volumes

examples/postgresql-ephemeral-template.json
pvalena: Ok. We'll need to keep an eye for that.

examples/postgresql-ephemeral-template.json
pkubatrh: I have modified this in preparation for the last stages of image content generation (see #203 ) - eventually there will not be a `9.6` to directly link to and so the current link would result in a 404.

pkubatrh: Additionally, it is currently impossible to link to the 9.6 README anyway, without going through `latest/` which might change into a different version in the future. So it makes sense to me to just point users to the root.

pvalena:
-- I hope it's not a problem for OpenShift, as they might link to it somewhere AFAIK.

@bparees, is it fine if `9.6` folder is just a symlink? Can we link root, or possibly `latest` directory?

pvalena:
FTR: Subsequently I'll create image-streams and PR for openshift/origin.

pvalena:
Also: *before* this PR gets merged I'd like to have the original template tested it by CI in OpenShift, so the changes can be actually validated.

See https://github.com/sclorg/postgresql-container/issues/204.

examples/postgresql-ephemeral-template.json
praiskup: pg_isready is not available for 9.2 anyway

examples/postgresql-ephemeral-template.json
```diff
@@ -127,11 +127,11 @@
                   }
                 },
                 "livenessProbe": {
+                  "exec": {
+                    "command": [ "/bin/sh", "-i", "-c", "pg_isready -h 127.0.0.1 -p 5432" ]
```
praiskup: I guess the `/bin/sh` is necessary, but I would rather avoid using the `-i` option, if possible.

pkubatrh: I have left it in only because it was present in other /bin/sh invocations inside the template

praiskup:
> @bparees, is it fine if 9.6 folder is just a symlink? Can we link root, or possibly latest directory?

This is long time discussed thing;  IMO linking directly to gitrepo - to concrete paths - is a bad practice.  There's a card on trello where we could discuss/implement a better place for documentation.

Regarding the link, it is a temporary issue ... since we could move the "latest" to PostgreSQL 10 soon, at least for testing purposes.

praiskup:
This looks good to me, but having this tested is a must :-)

examples/replica/postgresql_replica.json
```diff
@@ -71,7 +71,7 @@
       "name": "IMAGESTREAMTAG",
       "displayName": "ImageStreamTag",
       "description": "The OpenShift ImageStreamTag to use for PostgreSQL.",
-      "value": "postgresql:9.5"
+      "value": "postgresql:9.6"
```
bparees: there's no postgres 9.6 imagestream tag in openshift yet:
https://github.com/openshift/origin/blob/master/examples/image-streams/image-streams-centos7.json#L724-L795

so you need to deliver that before you start making templates default to looking for it.

bparees:
@praiskup can you lay out the plan/order of operations here?  is it to:

1) deliver this here
2) deliver the 9.6 imagestream to origin/image-streams
3) deliver this template to origin/db-templates

?

examples/replica/postgresql_replica.json
pvalena: We'll update at once - in our repos. The template will be actually tested with freshly build and tagged image into custom namespace.

praiskup:
According to the discussion here, I think so :-), per @pvalena comment:

>  FTR: Subsequently I'll create image-streams and PR for openshift/origin.

pvalena:
@bparees - The order is a bit parallelized, but basically:

 1. openshift tests
 2. updated template (container folder already done)
 3. generated [image-streams](https://github.com/sclorg/image-streams/tree/staging), although currently using [manual comparison](https://github.com/sclorg/image-streams/tree/origin)
 4. Pull Requests to openshift/origin

Where (1+2) and 3 are being worked on in parallel (I'm working on the image-streams; and together with @hhorak will provide help with the steps above).

pkubatrh:
I will look into getting the tests up and running so we can test this out properly.

pkubatrh:
[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
@praiskup Seems like resalloc does not want to cooperate again.

praiskup:
[test-openshift]

pkubatrh:
Thanks! Tests are green now

pkubatrh:
Fixed one more issue with pure image tests [test-openshift]

pkubatrh:
[test-openshift]

praiskup:
lgtm

pkubatrh:
Thanks for review! I will make one more rebase against container common scripts to pick up https://github.com/sclorg/container-common-scripts/pull/45

It is not needed for postgresql's openshift tests right now but I would like to have it fixed for eventual s2i (or other cases accessing the internet) before the problem is forgotten again.

pkubatrh:
[test-openshift]

pkubatrh:
Tests are green. Merging this in. Thanks everyone!

PR #207, opened 2017-11-16
Do not base Fedora image onto s2i-base

pkubatrh:


omron93:
Makes sense - RHEL/CentOS Dockerfiles aren't using s2i-base.

So merging.

PR #208, opened 2017-11-16
Source to image

dhodovsk:


latest/root/usr/bin/run-postgresql
```diff
@@ -22,12 +22,17 @@ else
   try_pgupgrade
 fi
 
+process_extending_files ${APP_DATA}/src/pre-init ${CONTAINER_SCRIPTS_PATH}/pre-init
```
pkubatrh: This call should also be added to `run-postgresql-slave`

latest/root/usr/share/container-scripts/postgresql/README.md
```diff
@@ -235,6 +235,54 @@ enough space for the copy;  upgrade failure because of not enough space might
...
+```
+$ s2i build --assemble-user root ~/image-configuration/ postgresql new-postgresql
+```
+
+The directory passed to `s2i build` should contain one or more of the
+following directories:
+
+##### `postgresql-config/`
+
+when running `run-service` command contained
```
pkubatrh: I would restate a bit differently since right now it seems a bit confusing. Something along the lines of:

The `postgresql.conf` configuration file contained in this folder is used instead of the default configuration when `postgresql` is started

pkubatrh:
@praiskup Please take a look as well

pkubatrh:
I wonder... Would this work for the Fedora version out of the box as well?

latest/root/usr/share/container-scripts/postgresql/README.md
```diff
@@ -235,6 +235,54 @@ enough space for the copy;  upgrade failure because of not enough space might
...
+
+This image can be extended using
+[source-to-image](https://github.com/openshift/source-to-image).
+
+For example to build customized image `new-postgresql`
+with configuration in `~/image-configuration/` run:
+
+
+```
+$ s2i build --assemble-user root ~/image-configuration/ postgresql new-postgresql
```
praiskup: Is `--asemble-user root` realistically useable in openshift?  Shouldn't we rather avoid root user at all?

latest/root/usr/share/container-scripts/postgresql/README.md
praiskup: This sounds bad to me. Shouldn't we again rather only "include" (additional?) configuration file into the default postgresql.conf?

examples/extending-image/postgresql-config/postgresql.conf
```diff
@@ -0,0 +1,528 @@
+# -----------------------------
+# PostgreSQL configuration file
+# -----------------------------
+#
```
praiskup: This example configuration file should be few-lines long only (this file is actually generated at postgresql build-time, and as such it would get staled very soon) and do something real..  and be included into the maint `postgresql.conf` file, or possibly transitively included into the configuration generated by our scripts...

latest/root/usr/share/container-scripts/postgresql/README.md
dhodovsk: User root is in this solution is used only for building the image using s2i. When the built container is run, it is under postgre user who is only in group postgre.

praiskup: So is this really commonly enabled feature in OpenShift, so we can make it a "general" requirement?  It seems to be pretty dangerous to get root access, even during the build..

dhodovsk:
@pkubatrh yes, it should also work with fedora, but I ran into a problem with python2 package:

Steps to reproduce:
```
$ docker run -it registry.fedoraproject.org/f25/s2i-base:latest bash
bash-4.3# dnf install -y python2
Fedora 25 - x86_64                                                        80 MB/s |  50 MB     00:00    
Fedora 25 - x86_64 - Updates                                              73 MB/s |  24 MB     00:00    
Last metadata expiration check: 0:00:06 ago on Fri Nov 24 08:51:08 2017.
Package python-2.7.13-2.fc25.x86_64 is already installed, skipping.
Dependencies resolved.
Nothing to do.
Complete!
bash-4.3# rpm -V python2
package python2 is not installed
```

Do you know something about this? Could you, please, help me resolve this so I can rebase and apply s2i also to fedora image?

pkubatrh:
@dhodovsk What exactly is the issue? Python 2.7 is installed, only the package is not named python2, just python (old style naming).

In any case the postgresql image should not be based on s2i-base at all. It is an oversight that is already fixed in master. It should be based on s2i-core, which is unfortunately not yet built. (https://github.com/sclorg/s2i-base-container/issues/144)

latest/root/usr/share/container-scripts/postgresql/README.md
dhodovsk: Other possibility is to add postgres user to root group (as is done e.g in mysql https://github.com/sclorg/mysql-container/blob/master/root-common/usr/libexec/container-setup#L58) but I can't help, it doesn't feel right. I think this issue deserves more discussion. @praiskup @hhorak what do you think?

dhodovsk:
@pkubatrh the issue occurred while verifying INSTALED_PKGS - https://github.com/sclorg/postgresql-container/blob/master/latest/Dockerfile.fedora#L51. It works with registry.fedoraproject.org/fedora:26 but not with registry.fedoraproject.org/f25/s2i-base:latest bash. I don't know the python background in fedora but it seemed interesting. Anyway I built fedora s2i-core locally and used it in fedora image with s2i here and it went ok.

latest/root/usr/share/container-scripts/postgresql/README.md
praiskup: To better understand, what exactly is the 'root' needed for?

dhodovsk: Actually, that is really good question! The problem with missing write permission (that caused failed s2i builds) on /opt/app-root/src is now managed in Dockerfiles with https://github.com/sclorg/postgresql-container/pull/208/files#diff-4040b6e79b7069db62e62fee9e5c3532R72. fix-permissions might fail, but the build finishes successfully.  Thank you for making this comment and standing your ground! :)

latest/s2i/bin/run
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+/usr/bin/run-postgresql
```
praiskup: Can this be symlink to `/usr/bin/run-postgresql`?

dhodovsk:
@praiskup please review discussed changes, there is another example of handling permissions for ${APP_DATA} here: https://github.com/sclorg/varnish-container/blob/master/4/Dockerfile#L46

examples/extending-image/start-hook/set_passwords.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+# postgresql image encrypts user passwords at service start
+# the functionality can be disabled by providing this file (start-hook/set_passwords.sh) in s2i build
```
praiskup: Missing newline.

latest/Dockerfile
```diff
@@ -67,6 +69,8 @@ ENV BASH_ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
 
 VOLUME ["/var/lib/pgsql/data"]
 
+# {APP_DATA} needs to be accessed by postgres user while s2i assembling
+RUN chown -R 26:0 ${APP_DATA}
```
praiskup: `chmod -R og+rwx` should be probably used too

latest/Dockerfile.rhel7
```diff
@@ -1,4 +1,4 @@
-FROM rhel7
+FROM rhscl/s2i-core-rhel7:1
```
praiskup: typo `:1` suffix

latest/root/usr/bin/run-postgresql-slave
```diff
@@ -28,9 +28,13 @@ check_env_vars
 generate_passwd_file
 generate_postgresql_config
 
+process_extending_files ${APP_DATA}/src/pre-init ${CONTAINER_SCRIPTS_PATH}/pre-init
+
 wait_for_postgresql_master
 export MASTER_FQDN=$(postgresql_master_addr)
 initialize_replica
 
+process_extending_files ${APP_DATA}/src/init ${CONTAINER_SCRIPTS_PATH}/init
```
praiskup: Should slave call these hooks?

latest/root/usr/share/container-scripts/postgresql/start-hook/set_passwords.sh
```diff
@@ -0,0 +1,14 @@
...
+fi
+
+if [ -v POSTGRESQL_MASTER_USER ]; then
+psql --command "ALTER USER \"${POSTGRESQL_MASTER_USER}\" WITH REPLICATION;"
+psql --command "ALTER USER \"${POSTGRESQL_MASTER_USER}\" WITH ENCRYPTED PASSWORD '${POSTGRESQL_MASTER_PASSWORD}';"
+fi
+
+if [ -v POSTGRESQL_ADMIN_PASSWORD ]; then
+psql --command "ALTER USER \"postgres\" WITH ENCRYPTED PASSWORD '${POSTGRESQL_ADMIN_PASSWORD}';"
+fi
```
praiskup: Probably editor mis-config (missing newline).

latest/root/usr/share/container-scripts/postgresql/init-hook/set-config.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+cat >> "$PGDATA/postgresql.conf" <<EOF
+
+# Custom extending configuration
+EOF
+
+extending_cfg_dir="${APP_DATA}/src/postgresql-config"
+
+for conf in $(ls ${extending_cfg_dir}); do
```
praiskup: `for conf in "$extending_cfg_dir"/*.conf; do`

latest/root/usr/share/container-scripts/postgresql/init-hook/set-config.sh
```diff
@@ -0,0 +1,14 @@
...
+    cat >> "$PGDATA/postgresql.conf" <<EOF
+include '$extending_cfg_dir/${conf}'
+EOF
```
praiskup: Nit, but echo would be also OK and a bit more readable.

praiskup:
Thanks, it is close now :-) could we please add s2i test-cases?  Mariadb can be used for inspiration.

latest/root/usr/bin/run-postgresql-slave
pkubatrh: I asked for it to be added because I thought even slaves want to have their configuration files extendible. But looking at pg_basebackup documentation it seems that the slaves will get the master's configurations files along with everything else, so it might not be needed after all.

praiskup: More it seems to me that the password change method is newly called for slave, too.

latest/root/usr/bin/run-postgresql-slave
```diff
@@ -28,9 +28,13 @@ check_env_vars
 generate_passwd_file
 generate_postgresql_config
 
+process_extending_files ${APP_DATA}/src/pre-init ${CONTAINER_SCRIPTS_PATH}/pre-init
```
omron93: This file needs to be updated to use `init-hook` and `start-hook` directories.

dhodovsk:
The requested changes has been applied; now also to Dockerfile.template and versions 9.5 and 9.4.

test/run_test
```diff
@@ -132,7 +135,7 @@ function create_container() {
   local cargs=${DOCKER_ARGS:-}
   cidfile="$CIDFILE_DIR/$name"
   # create container with a cidfile in a directory for cleanup
-  docker run $cargs --cidfile $cidfile -d $IMAGE_NAME "$@"
+  docker run $cargs --cidfile $cidfile -d "$@" $IMAGE_NAME
```
praiskup: This doesn't seem right.  `$@` brings the actual command to be run inside the container.

test/run_test
```diff
@@ -558,6 +561,60 @@ run_doc_test() {
...
+
+  # need to set these because `postgresql_cmd` relies on global variables
+  PGUSER=user
+  PASS=password
+
+  # need this to wait for the container to start up
+  CONTAINER_IP=$(get_container_ip ${container_name})
+  test_connection "$container_name"
+
+  configuration="$(docker exec -t "$(get_cid $container_name)" bash -c 'set +f; shopt -s nullglob; egrep -hv "^(#|\!|\[|$)" /etc/my.cnf /etc/my.cnf.d/* /opt/rh/mysql*/root/etc/my.cnf /opt/rh/mysql*/root/etc/my.cnf.d/*' | sed 's,\(^[[:space:]]\+\|[[:space:]]\+$\),,' | sort -u)"
```
praiskup: I'm bit lost here :)

test/run_test
```diff
@@ -34,6 +35,8 @@ test -n "${OS-}" || false 'make sure $OS is defined'
 CIDFILE_DIR=$(mktemp --suffix=postgresql_test_cidfiles -d)
 
 volumes_to_clean=
+s2i_args="--pull-policy=never "
```
praiskup: Used only once, shouldn't be used everywhere?  Could this be moved to `run_s2i_test`?

test/run_test
```diff
@@ -34,6 +35,8 @@ test -n "${OS-}" || false 'make sure $OS is defined'
...
+test_dir="$(readlink -zf $(dirname "${BASH_SOURCE[0]}"))"
```
praiskup: `test_dir=$(readlink -f "$(dirname "$0")")` could work too, preferably should be moved to `run_s2i_test` too as a local variable.

test/run_test
```diff
@@ -558,6 +561,60 @@ run_doc_test() {
...
+  _s2i_test_image "s2i_config_build" ""
+
+  # return back original value for IMAGE_NAME
+  export IMAGE_NAME=${image_name_backup}
+
+  echo "  Testing s2i mount"
+  test_app_dir=$(mktemp -d)
+  cp -Lr ${test_dir}/test-app ${test_app_dir}/
+  _s2i_test_image "_s2i_test_mount" "-v ${test_app_dir}/test-app:/opt/app-root/src/:z"
+  rm -rf ${test_app_dir}
```
praiskup: This is not executed if `cp` or `_s2i_test_image` fails.

test/run_test
```diff
@@ -558,6 +561,60 @@ run_doc_test() {
...
+
+    DOCKER_ARGS="
+-e POSTGRESQL_DATABASE=db
+-e POSTGRESQL_USER=user
+-e POSTGRESQL_PASSWORD=password
+-e POSTGRESQL_ADMIN_PASSWORD=password
+-e POSTGRESQL_BACKUP_USER=backuser
+-e POSTGRESQL_BACKUP_PASSWORD=pass
+"
+  create_container "$container_name" ${mount_opts}
```
praiskup: Should be probably: 
```
DOCKER_ARGS="
...
" create_container ...
```
... to not set the global.

9.4/Dockerfile.rhel7
```diff
@@ -70,6 +72,9 @@ ENV BASH_ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
 
 VOLUME ["/var/lib/pgsql/data"]
 
+# {APP_DATA} needs to be accessed by postgres user while s2i assembling
+RUN chown -R 26:0 ${APP_DATA} && \
+    chmod -R og+rwx ${APP_DATA}
```
praiskup: Should others have `+rwx` too?

dhodovsk: I don't think I understand. Do you suggest in rhel image different approach from the one in centos image? https://github.com/sclorg/postgresql-container/pull/208#discussion_r153995545

praiskup: I always mistakenly substitute s/owner/o/ instead of s/user/u/.  Sorry for previous bad advice..

latest/Dockerfile
praiskup: I meant `chmod -R ug+rwx`.

9.4/Dockerfile.rhel7
praiskup: So the answer actually :-) no -> rhel/centos should do the same thing from this POV, I'm just asking whether "others" need to have the `rwx` permissions (IMO it shouldn't be needed).  This is security question...

test/run_test
```diff
@@ -53,6 +56,8 @@ function cleanup() {
   rmdir $CIDFILE_DIR
 
   ct_path_foreach "$volumes_to_clean" cleanup_volume_dir
+
+  test -d $s2i_test_app_dir && rm -rf ${s2i_test_app_dir}
```
praiskup: Please double quote (especially with `rm -rf`).

praiskup:
It would be awesome to have also openshift s2i test-case :-), non blocker though, just saying...

pkubatrh:
@dhodovsk Could this also be rebased to latest master so we can try this out in Openshift as well? Having an s2i openshift test case would be great too

pkubatrh:
Actually because of the limitation that the example application needs to be already stored inside a remote repository the openshift s2i tests would not work anyway unless we hacked around it...

dhodovsk:
It is now rebased and issue for trancking openshift s2i test is opened here https://github.com/sclorg/postgresql-container/issues/214

pkubatrh:
Thanks! [test-openshift]

hhorak:
I didn't look deeply but it looks like a great work, thanks all involved! One thing I wanted to check -- is it possible to prepare an example for using TLS for postgresql connections using this new feature by extending the image? it's requested in https://bugzilla.redhat.com/show_bug.cgi?id=1483380 and s2i/flexibility should allow that IMO.

praiskup:
I'm still not sure about permissions in openshift-case; and since we are not yet testing that, are we sure that it actually works? The `docker run --rm -ti  -u <randomid>  centos/postgresql-96-centos7 id` say that the group id is `root`, which means that the fix-permissions shouldn't complain in assemble script .. but still, mongo/mariadb containers add explicitly the user to `root` group -- are those containers just unnecessarily benevolent regarding permissions, or isn't the openshift case not yet working?

9.4/Dockerfile.rhel7
```diff
@@ -70,6 +72,9 @@ ENV BASH_ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
 
 VOLUME ["/var/lib/pgsql/data"]
 
+# {APP_DATA} needs to be accessed by postgres user while s2i assembling
+RUN chown -R 26:0 ${APP_DATA} && \
+    chmod -R ug+rwx ${APP_DATA}
```
praiskup: Thinking again, shouldn't we actually use the `fix-permissions` script few lines above while we are fixing the permissions of datadir?

9.5/root/usr/share/container-scripts/postgresql/README.md
```diff
@@ -235,6 +235,45 @@ enough space for the copy;  upgrade failure because of not enough space might
...
+```
+$ s2i build ~/image-configuration/ postgresql new-postgresql
+```
+
+The directory passed to `s2i build` should contain one or more of the
+following directories:
+
+##### `postgresql-config/`
+
+contained files will be included at the end of image postgresql.conf file during database initialization
```
praiskup: Only `*.conf` files will be included. Slightly clearer would be `postgresql-init-config` (since this is only used during initialization).

praiskup: But, OTOH, why not to drop this hook at all, until it is explicitly requested?  It is pretty brittle, and easily implementable through `init/` by simply appending to the config file in the shell file.  Note that we are not even testing that the include statement actually works...

9.5/root/usr/share/container-scripts/postgresql/README.md
```diff
@@ -235,6 +235,45 @@ enough space for the copy;  upgrade failure because of not enough space might
...
+
+
+##### `init/`
+
+contained shell scripts (`*.sh`) are sourced once, when database is initialized
```
praiskup: I'm not sure, but is this explicitly named `init/`, instead of `postgresql-init/`? Do you count with multi-service container use-case (that more daemons will source the shell files)?

pkubatrh:
>The docker run --rm -ti -u <randomid> centos/postgresql-96-centos7 id say that the group id is root, which means that the fix-permissions shouldn't complain in assemble script

I tried it out quickly inside an Openshift s2i build and actually the situation is different in there:

```
---> Installing application source ...
+ id
uid=26(postgres) gid=26(postgres) groups=26(postgres)
+ /usr/libexec/fix-permissions ./
```

followed by a lot of `Operation not permitted` warnings, as expected. So I guess adding the postgres user to the root group is necessary after all.

praiskup:
So yeah, seems like this ugly `usermod` is needed, we should carefully document why is it so (so we don't have to reobserve).  Just a question -- do we really need to run `fix-permission` during assemble run?  What files does that really affect?

praiskup:
> What files does that really affect?

Answering myself, many of them.  New question :-) do we need write permissions actually?

pkubatrh:
>do we need write permissions actually?

For the source files under `/opt/app-root/src`? One use case that comes to mind is easier debugging of the built image without the need to have it rebuilt with new sources

omron93:
> do we really need to run fix-permission during assemble run? 

We need to add/ensure g+r for `root` group. To be able to run this image under random UID (~group is 0)... So I think yes.

dhodovsk:
So based on comments above,  fix-permission wasn't changed in latest rebase and is called during assemble. Postgres user is added to root group in Dockerfile to be able to perform the operations.

9.4/Dockerfile.rhel7
```diff
@@ -70,6 +72,11 @@ ENV BASH_ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
 
 VOLUME ["/var/lib/pgsql/data"]
 
+# {APP_DATA} needs to be accessed by postgres user while s2i assembling
+# postgres user changes permissiond of files in APP_DATA during assemblng
+RUN chown -R 26:0 ${APP_DATA} && \
+    useradd -G root 26
```
praiskup: This doesn't work on my system.  You probably want `usermod -a -G root postgres`.
I don't think removal of `chmod` was correct, because default umask should be 0022 -> which means that group get's only `r-x` permissions.

pkubatrh:
[test-openshift]

9.4/root/usr/share/container-scripts/postgresql/README.md
```diff
@@ -235,6 +235,45 @@ enough space for the copy;  upgrade failure because of not enough space might
...
+##### `postgresql-config/`
+
+contained files will be included at the end of image postgresql.conf file during database initialization
+
+
+##### `init/`
+
+contained shell scripts (`*.sh`) are sourced once, when database is initialized
+
+##### `start-hook/`
```
praiskup: Still no comment about `s/start-hook/postgresql-start-hook/`.  Mariadb has `mysql-cfg`, `mysql-init` and `mysql-pre-init`.  Shouldn't we have the namespaced hook directory, too?

9.5/Dockerfile.rhel7
```diff
@@ -70,6 +72,12 @@ ENV BASH_ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
 
 VOLUME ["/var/lib/pgsql/data"]
 
+# {APP_DATA} needs to be accessed by postgres user while s2i assembling
+# postgres user changes permissiond of files in APP_DATA during assemblng
```
praiskup: What about:
Being also in supplemental group `0`, postgres user can by default change the group
ownership of files in $APP_DATA to `postgres:root` (by fix-permissions script in assemble
script).  This group ownership change is needed because the s2i-built image is run under
arbitrary user uid (but gid=0).

9.5/Dockerfile.rhel7
```diff
@@ -70,6 +72,12 @@ ENV BASH_ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
...
+RUN chown -R postgres:0 ${APP_DATA} && \
+    chmod -R ug+rwx ${APP_DATA} && \
+    usermod -a -G root postgres
```
praiskup: I still have the two questions:
- is the write permissions needed in case of postgresql?
- could we just drop chown/chmod and use `fix-permissions` above (where the data ownership is changed)?

9.5/Dockerfile.rhel7
hhorak: two typos: "permissiond" and "assemblng"

9.5/Dockerfile.rhel7
hhorak: I think `fix-permissions` should be enough, at least it looks like it's enough for mysql: https://github.com/sclorg/mysql-container/blob/master/root-common/usr/libexec/container-setup#L57

hhorak:
@dhodovsk @praiskup It's hard to see whether there is still anything to be fixed. From a quick look, it seems things like 's/init/postgresql-init/` and `s/start-hook/postgresql-start-hook/` are still to be addressed...

praiskup:
I would prefer to have that hooks namespaced, yes.

dhodovsk:
@praiskup Is it now ok?

9.5/root/usr/share/container-scripts/postgresql/README.md
praiskup: Comments are still valid, but if not dropping this (smaller number of hooks, the less work with maintenance) - at least please document that *.conf files are included.

praiskup:
Seems to be, small nit with the docs.

latest/root/usr/share/container-scripts/postgresql/README.md
```diff
@@ -235,6 +235,45 @@ enough space for the copy;  upgrade failure because of not enough space might
...
+```
+$ s2i build ~/image-configuration/ postgresql new-postgresql
+```
+
+The directory passed to `s2i build` should contain one or more of the
+following directories:
+
+##### `postgresql-config/`
+
+contained files will be included at the end of image postgresql.conf file during database initialization
```
praiskup: Meh, github fight's against me.  This should be either dropped, or at least you should mention that only *.conf files are sourced.

praiskup:
[test-openshift]

praiskup:
lgtm

pkubatrh:
Merging, thanks!

PR #209, opened 2017-11-22
tests: add run-openshift template tests

pkubatrh:
Resolves: #204

### DIFF
--- a/test/run-openshift
+++ b/test/run-openshift
@@ -8,24 +8,23 @@
 
 THISDIR=$(dirname ${BASH_SOURCE[0]})
 
-source ${THISDIR}/test-lib-openshift.sh
+source "$THISDIR"/test-lib-openshift.sh
 
 set -exo nounset
 
 function check_postgresql_os_service_connection() {
-  local util_image_name="${1}" ; shift
-  local service_name="${1}" ; shift
-  local user="${1}" ; shift
-  local pass="${1}" ; shift
-  local database="${1}" ; shift
-  local timeout="${1:-60}" ; shift || :
+  local util_image_name=$1 ; shift
+  local service_name=$1 ; shift
+  local user=$1 ; shift
+  local pass=$1 ; shift
+  local database=$1 ; shift
+  local timeout=${1:-60} ; shift || :
   local pod_ip=$(ct_os_get_service_ip ${service_name})
 
   : "  Service ${service_name} check ..."
 
-  local cmd="echo 'SELECT 42 as testval' | PGPASSWORD=${pass} PGCONNECT_TIMEOUT=15 \
-             psql -t -h ${pod_ip} -U ${user} -d ${database}"
-  local expected_value='42'
+  local cmd="PGPASSWORD=${pass} pg_isready -t 15 -h ${pod_ip} -U ${user} -d ${database}"
+  local expected_value='accepting connections'
   local output
   local ret
   SECONDS=0
@@ -47,7 +46,7 @@ function check_postgresql_os_service_connection() {
 }
 
 function test_postgresql_pure_image() {
-  local image_name=${1:-centos/postgresql-96-centos7}
+  local image_name=$1
   local image_name_no_namespace=${image_name##*/}
   local service_name=${image_name_no_namespace}
 
--- a/test/run-openshift
+++ b/test/run-openshift
@@ -1,0 +1,90 @@
+#!/bin/bash
+#
+# Test the Postgresql image in OpenShift.
+#
+# IMAGE_NAME specifies a name of the candidate image used for testing.
+# The image has to be available before this script is executed.
+#
+
+THISDIR=$(dirname ${BASH_SOURCE[0]})
+
+source ${THISDIR}/test-lib-openshift.sh
+
+set -exo nounset
+
+function check_postgresql_os_service_connection() {
+  local util_image_name="${1}" ; shift
+  local service_name="${1}" ; shift
+  local user="${1}" ; shift
+  local pass="${1}" ; shift
+  local database="${1}" ; shift
+  local timeout="${1:-60}" ; shift || :
+  local pod_ip=$(ct_os_get_service_ip ${service_name})
+
+  : "  Service ${service_name} check ..."
+
+  local cmd="echo 'SELECT 42 as testval' | PGPASSWORD=${pass} PGCONNECT_TIMEOUT=15 \
+             psql -t -h ${pod_ip} -U ${user} -d ${database}"
+  local expected_value='42'
+  local output
+  local ret
+  SECONDS=0
+
+  echo -n "Waiting for ${service_name} service becoming ready ..."
+  while true ; do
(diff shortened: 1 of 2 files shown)


pkubatrh:
@omron93 Are the tests run when they are present or do we have to trigger them manually in some way?

test/run-openshift
```diff
@@ -0,0 +1,90 @@
...
+#
+
+THISDIR=$(dirname ${BASH_SOURCE[0]})
+
+source ${THISDIR}/test-lib-openshift.sh
+
+set -exo nounset
+
+function check_postgresql_os_service_connection() {
+  local util_image_name="${1}" ; shift
```
praiskup: Nit:please use just "foo=$baz" syntax.  Looks much nicer, the quotes and brackets are not needed in this case.

test/run-openshift
```diff
@@ -0,0 +1,90 @@
...
+#
+# Test the Postgresql image in OpenShift.
+#
+# IMAGE_NAME specifies a name of the candidate image used for testing.
+# The image has to be available before this script is executed.
+#
+
+THISDIR=$(dirname ${BASH_SOURCE[0]})
+
+source ${THISDIR}/test-lib-openshift.sh
```
praiskup: I would rather quote $THISDIR.

test/run-openshift
```diff
@@ -0,0 +1,90 @@
...
+  local user="${1}" ; shift
+  local pass="${1}" ; shift
+  local database="${1}" ; shift
+  local timeout="${1:-60}" ; shift || :
+  local pod_ip=$(ct_os_get_service_ip ${service_name})
+
+  : "  Service ${service_name} check ..."
+
+  local cmd="echo 'SELECT 42 as testval' | PGPASSWORD=${pass} PGCONNECT_TIMEOUT=15 \
+             psql -t -h ${pod_ip} -U ${user} -d ${database}"
```
praiskup: Could we use `pg_isready`, since that will be the live check in future, too?

test/run-openshift
```diff
@@ -0,0 +1,90 @@
...
+    echo -n "."
+    [ ${SECONDS} -gt ${timeout} ] && break
+    sleep 3
+  done
+  echo " FAIL"
+  return 1
+}
+
+function test_postgresql_pure_image() {
+  local image_name=${1:-centos/postgresql-96-centos7}
```
praiskup: The default value seems to be unused, so `image_name=$1` would look a bit safer.

test/run-openshift
```diff
@@ -0,0 +1,90 @@
...
+  ct_os_deploy_template_image ${THISDIR}/postgresql-ephemeral-template.json \
+                              DATABASE_SERVICE_NAME="${service_name}" \
+                              POSTGRESQL_USER=testu \
+                              POSTGRESQL_PASSWORD=testp \
+                              POSTGRESQL_DATABASE=testdb
+
+  ct_os_wait_pod_ready "${service_name}" 60
+  check_postgresql_os_service_connection "${image_name}" "${service_name}" testu testp testdb
+
+  ct_os_delete_project
```
praiskup: Nice, this seems to be promising!  I'm looking forward to see the library code.

praiskup: Ah, it's already there :-) my bad.

praiskup:
Seems like the 'test-openshift' target used to be in https://github.com/sclorg/rhscl-container-ci/blob/master/yaml/jobs/collections/mongodb-rh.yaml so the openshift tests used to be triggered automatically for mongo.  Better approach could be to invoke the `tests-openshift` target for every image, and that target could succeed with "not-implemented" message if the image is not yet CI-tested for openshift.

pkubatrh:
@praiskup Made some changes according to your suggestions, thanks!

pkubatrh:
Seems like the code works on both rhel and centos-based Openshift installations. Only the library does not yet provide any functions to install openshift on a rhel7 system.

pkubatrh:
@omron93 Since `make test-openshift` should be a no-op if the openshift tests are not present can we enable the make target to be triggered by default on every image in CI?

omron93:
> @omron93 Since make test-openshift should be a no-op if the openshift tests are not present can we enable the make target to be triggered by default on every image in CI?

OK. I've enabled it for all images. Test **has to be explicitly triggered** by [test-openshift]

> Seems like the code works on both rhel and centos-based Openshift installations. Only the library does not yet provide any functions to install openshift on a rhel7 system.

I would let installing oc tool to the user (and to the CI), same we do with `docker-squash`. It simplify our tests and avoid problems with different environments.

pkubatrh:
>RHEL CI is failing because you try to install centos package (CI image already contain oc).

Looking at the logs the centos image does not seem to have openshift installed by default. Are we going to update it so it does or should I just make the install function call specific to centos?

pkubatrh:
>In CentOS there is problem with starting cluster. I'll investigate it tomorrow.

The same issue is in rhel, it happens only the first time the `cluster up` action is called. No idea how to work around it other than just not failing the first time and trying again...

omron93:
> The same issue is in rhel, it happens only the first time the cluster up action is called. No idea how to work around it other than just not failing the first time and trying again...

iptables has to be flushed. Please take a look at https://github.com/sclorg/container-common-scripts/pull/40

With this it should work. No `ct_os_install_in_centos` is needed (I would let this to user to setup testing machine - number of OS version,... is too big). So only `ct_os_cluster_up` is enought to start OpenShift in CI.

pkubatrh:
[test-openshift]

pkubatrh:
Seems like the openshift test fails now on `docker login`

pkubatrh:
[test-openshift]

omron93:
> Seems like the openshift test fails now on docker login

It seems that it takes some time. In last centos openshift tests it succeeded in fifth attempt.

I'll try to find out what is causing this.

pkubatrh:
[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
@praiskup @omron93 For some reason I do not see rhel7 CI starting.

omron93:
> For some reason I do not see rhel7 CI starting.

There was Jenkins upgrade by Central CI admins. But problems with plugins occurred so we have to wait.

pkubatrh:
All checks passing now

praiskup:
Looks fine.

pkubatrh:
Thanks for review! Merging

test/run-openshift
```diff
@@ -0,0 +1,88 @@
...
+                          --env POSTGRESQL_ADMIN_PASSWORD=test
+
+  ct_os_wait_pod_ready "${service_name}" 60
+  check_postgresql_os_service_connection "${image_name}" "${service_name}" postgres test postgres
+
+  ct_os_delete_project
+}
+
+function test_postgresql_template() {
+  local image_name=${1:-centos/postgresql-96-centos7}
```
pvalena: Why is there 'backup' image_name?
IMHO this should be tested specifically with the CI-built image.

PR #212, opened 2017-11-24
tests: do not set default $OS

pkubatrh:
We check whether `$OS`is set in the run-test script. If we fallback to a default before that check the test does not make any sense.


praiskup:
I don't see any problem on the first sight.  Thanks!

omron93:
Thanks. Merging.

PR #215, opened 2017-12-13
Adding server started message to run scripts

coreydaley:
Adding server started message to run scripts
since the last message displayed is always
server stopped which can be confusing for users

### DIFF
--- a/9.2/root/usr/bin/run-postgresql
+++ b/9.2/root/usr/bin/run-postgresql
@@ -25,4 +25,5 @@ set_passwords
 pg_ctl stop
 
 unset_env_vars
+echo "Starting server..."
 exec postgres "$@"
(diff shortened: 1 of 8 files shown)


coreydaley:
@bparees ptal

pkubatrh:
This message will never get printed out as the shell execs into `postgres` right before it ...

coreydaley:
@pkubatrh Whoops, I'll move the message before the exec and repush.  thanks

coreydaley:
@pkubatrh patch updated, ptal

bparees:
I probably would have gone w/ "Starting server..." (and i'm curious why we get no stdout from the server when it starts?) but this is fine.

coreydaley:
@bparees updated

pkubatrh:
Yes, "starting server" makes more sense, thanks! lgtm [test]

pkubatrh:
[test]

pkubatrh:
Merging, thanks!

Issue #222, opened 2018-01-10
generated 9.5/Dockerfile.rhel7 differs from latest/Dockerfile.rhel7


praiskup:
@dhodovsk can you have a look?
```patch
     yum-config-manager --enable rhel-7-server-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
-    INSTALL_PKGS="rsync tar gettext bind-utils nss_wrapper rh-postgresql95 rh-postgresql95-postgresql-contrib rh-postgresql94-postgresql-server" && \
+
+    INSTALL_PKGS="rsync tar gettext bind-utils nss_wrapper rh-postgresql96 rh-postgresql96-postgresql-contrib rh-postgresql95-postgresql-server" && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
```

dhodovsk:
This is expected, latest version is 9.6.

pkubatrh:
>This is expected, latest version is 9.6.

Pavel was poining out that the generated version of `latest/Dockerfile.rhel7` has an empty line in the middle of a RUN command (which would likely make the build fail)

dhodovsk:
Fix is provided in referenced PR.

PR #224, opened 2018-01-10
Update Dockerfile.template

dhodovsk:
Fixes https://github.com/sclorg/postgresql-container/issues/222

This PR also adds recent Dockerfile changes to Dockerfile.template

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -30,14 +30,12 @@ LABEL summary=$SUMMARY \
       name="{{ spec.org }}/postgresql-{{ spec.short }}-{{ spec.prod }}" \
       com.redhat.component="{{ spec.redhat_component }}" \
 {%- if spec.version in spec.rhscl3_introduced %}
-      version="1"
+      version="1" \
 {%- else %}
       version="{{ spec.version }}" \
-      release="1"
 {%- endif %}
-{%- if spec.maintainer %} \
-      maintainer="{{ spec.maintainer }}"
-{%- endif %}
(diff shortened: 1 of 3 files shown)


Dockerfile.template
```diff
@@ -30,14 +30,12 @@ LABEL summary=$SUMMARY \
...
+      version="1" \
 {%- else %}
       version="{{ spec.version }}" \
-      release="1"
-{%- endif %}
-{%- if spec.maintainer %} \
-      maintainer="{{ spec.maintainer }}"
 {%- endif %}
+      usage="docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 {{ spec.org }}/postgresql-{{ spec.short }}-{{ spec.prod }}" \
+      maintainer="SoftwareCollections.org <sclorg@redhat.com>"
```
praiskup: While we are touching this, shouldn't maintainer= be set only for Fedora dockerfiles?

dhodovsk: @hhorak what do you think? Maintainer has been added to every Dockerfile in this commit: https://github.com/sclorg/postgresql-container/commit/b66f12fdc924c49b3f9b2f657546365f15d9bf93

pkubatrh: Last time we had this conversation we got to the conclusion that it is ok to leave this in if it simplifies things. But this should really get documented somewhere in our guidelines since are coming back to it...

hhorak: Given we set the maintainer label in s2i-core, it should be inherited to other containers, so omitting it is probably fine, but I don't see any problem with re-defining it, so if it makes the template simpler, I'm fine with that.

pkubatrh:
lgtm, thanks! Merging

PR #231, opened 2018-02-07
run_test: introduce $TESTS variable

praiskup:
If the $TESTS variable is set to space-separated list of test-case
names, only those will be executed, e.g.:

$ make test VERSIONS=9.5 TESTS=run_s2i_test


omron93:
Why not to use TESTS variable in https://github.com/sclorg/container-common-scripts/blob/master/test.sh#L16 ? - pass content of TESTS variable to `test/run` script.

To still be able to pass tests on command line as arguments when running test script directly... Or I don't know: Which way of selecting tests is "better" in bash world?

praiskup:
Thanks for having a look!  Take this PR as "hotfix" :-) to ease our development rather than real fix.  I think both ways would be fine for now, but if we implemented this in commons we would have to enforce that all the `./run` scripts to have the same semantics..  And before we do, we should move the `run` script into commons too, and give it a proper "commandline interface" imo.

omron93:
Makes sense. LGTM.

PR #232, opened 2018-02-07
iron the s2i hooks

praiskup:
- postgresql-config renamed to postgresql-cfg to have consistent
  naming across sclorg images
- the postgresql-cfg is now handled in "pre-init" phase, at the
  same time when the environment volumes are taken into account
- postgresql-start-hook renamed to postgresql-start to avoid
  non-standard -hook suffix
- better documentation to ensure reader when/under which condition
  the hooks are evaluated


praiskup:
@ficap, @hhorak, @omron93, @pkubatrh followup for today's discussion

hhorak:
LGTM, except that the examples are not changed to match this new behaviour.

praiskup:
[test-openshift]

praiskup:
Seems like some random issue? https://coreservices-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhscl-images-postgresql-rh-openshift/33/ [test-openshift]

test/run_test
```diff
@@ -597,19 +598,19 @@ run_s2i_test() {
...
+  IMAGE_NAME=$s2i_image_name _s2i_test_image s2i_config_build ""
 
   echo "  Testing s2i mount"
-  cp -Lr "${test_dir}/test-app" "${s2i_test_app_dir}/"
-  _s2i_test_image "_s2i_test_mount" "-v ${s2i_test_app_dir}/test-app:/opt/app-root/src/:z"
+  create_volume_dir
+  cp -Lr "$test_dir/test-app/postgresql-init" "$volume_dir"
+  setfacl -R -m u:26:r-x "$volume_dir" # Execute bit is needed only for dirs
+
+  IMAGE_NAME=$s2i_image_name _s2i_test_image _s2i_test_mount "-v ${volume_dir}:/opt/app-root/src/:z"
```
pkubatrh: Shouldn't this run with the original (un-extended) image?

praiskup: Oh, right, good catch (not a new bug, though?)

pkubatrh: From what I can see the old version ran it that way:
```
local image_name_backup="${IMAGE_NAME}
export IMAGE_NAME="${IMAGE_NAME}-testapp"
...
export IMAGE_NAME="${image_name_backup}"
echo "  Testing s2i mount"
cp -Lr "${test_dir}/test-app" "${s2i_test_app_dir}/"
_s2i_test_image "_s2i_test_mount" "-v ${s2i_test_app_dir}/test-app:/opt/app-root/src/:z"
```

omron93:
LGTM, thanks.

pkubatrh:
[test-openshift]

praiskup:
[test-openshift]

praiskup:
[test] seems like we have oom issues in rhel CI

ficap:
how about moving https://github.com/sclorg/postgresql-container/pull/232/files#diff-7246cf97d2ed60c5d8e8d585c1b7050fR10 these functionality to /usr/share/container-scripts/postgresql/pre-init as it is done with postgresql-start to allow user to override this functionality as was proposed in new concept?

ficap:
also to have consistent naming I'd recommend renaming `/usr/share/container-scripts/postgresql/postgresql-start` to ` `.../start`. or what do you think? name-prefixed dirs are meant to be used with user defined s2i configs like `${APP_ROOT}/src/postgresql-start`. what's your opinion? also you can check out our diagram available at https://github.com/ficap/sclorg-s2i-images-schemes and provide any feedback.

praiskup:
> move stuff to /usr/share/container-scripts/postgresql/pre-init

I don't see big problem in this, but new hook deserves new PR I think

> renaming /usr/share/container-scripts/postgresql/postgresql-start to .../start

I don't know whether user can not have two containers configured through
one s2i project (say apache config and postgres config in one directory),
if that was possible, the prefixed paths are slightly better.

ficap:
I agree but these are our scripts and user is not able standard way to manipulate them. User uses `${APP_ROOT}/src/postgresql-start`. and priority handling is done using `process_extend_files`

praiskup:
If you mean `$CONTAINER_SCRIPTS_PATH`, it is IMO just an implementation detail.
Users shouldn't care what path that variable points to..?

praiskup:
Fixed (after f2f discussion), @ficap PTAL.

praiskup:
[test-openshift]

ficap:
LGTM. Try to move stuff to /usr/share/container-scripts/postgresql/pre-init in other PR. But it's not critical yet. :)

ficap:
Other PR can be move `root/usr/share/container-scripts/postgresql/*conf.template` to`.../cnf/*conf.template`

praiskup:
Please consider filling separate issues to not forget about your thoughts.  Config 
files in `src/postgresql-cfg` are not supposed to contain templates.

PR #236, opened 2018-02-13
Let the username/password validation on PostgreSQL server

praiskup:
Psql has a feature guarding against sql injection (psql --set),
so use this and don't try to add some artificial limits on our
own.  This has a benefits like:

 - we don't have to care about sql injections
 - our passwords can contain spaces, and other special
   characters
 - simplified "usage" output

The downside is that the rejection of wrongly specified
passwords/username is a delayed.  But because PostgreSQL itself
doesn't create any artificial requirements on those values, it is
pretty unlikely.

### DIFF
--- a/src/root/usr/share/container-scripts/postgresql/common.sh
+++ b/src/root/usr/share/container-scripts/postgresql/common.sh
@@ -23,9 +23,6 @@ export POSTGRESQL_CONFIG_FILE=$HOME/openshift-custom-postgresql.conf
 
 postinitdb_actions=
 
-psql_identifier_regex='^[a-zA-Z_][a-zA-Z0-9_]*$'
-psql_password_regex='^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$'
-
 # match . files when moving userdata below
 shopt -s dotglob
 # extglob enables the !(userdata) glob pattern below.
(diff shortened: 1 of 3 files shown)


praiskup:
Take this as "suggestion" only or topic for discussion, please.

hhorak:
This makes pretty good sense to me, I only wish there was something similar for mysql/mariadb :(

praiskup:
Well there still must be a way even in mariadb/mysql, at least through some
library (say python library, I've checked that python runtime is awailable
in mariadb container).

hhorak:
Yes, that would help, although I don't think it's worth installing more packages to the container. I'd rather talk to upstream and make such a feature available in `mysqladmin` tool.

praiskup:
[test-openshift]

praiskup:
[test-openshift]

praiskup:
Anybody thinks that this change is not wise idea?  I'm just thinking whether I should rebase this, or just close.

praiskup:
The test failure should be fixed by #263.

pkubatrh:
[test]

praiskup:
Rebased, with some additional testsuite changes.

praiskup:
[test] Resubmit for:
```
Running: yum-config-manager --disable rhel-\*
FATAL: command execution failed
hudson.remoting.ProxyException: java.util.concurrent.TimeoutException: Ping started at 1522325945692 hasn't completed by 1522326185693
Caused: Command close created at
	at hudson.remoting.Command.<init>(Command.java:6
```

pkubatrh:
[test-openshift]

pkubatrh:
The centos CI is failing because it expects the unreleased v10 image to not be available in a registry (seems like a wrong expectation to make)

rhel CI failure seems like #255 again ...

praiskup:
[test-openshift]

praiskup:
[test-openshift]

praiskup:
[test][test-openshift]

praiskup:
[test-openshift]

praiskup:
[test-openshift]

praiskup:
[test][test-openshift]

praiskup:
[test] [test-openshift]

praiskup:
What about this one?

pkubatrh:
Makes sense. Tested locally with podman just in case.

LGTM, merging

PR #242, opened 2018-03-03
Build PostgreSQL 10

praiskup:


praiskup:
[test-openshift]

praiskup:
[test-openshift]

praiskup:
[test-openshift]

praiskup:
[test-openshift]

praiskup:
Please take a look.

praiskup:
[test-openshift]
```patch
-       test/run_upgrade_test 9.2:remote 9.4:local 9.5:local 9.6:local
+       OS=$(OS) test/run_upgrade_test 9.2:remote 9.4:local 9.5:local 9.6:local 10:local
```

praiskup:
[test-openshift]

Dockerfile.template
```diff
@@ -48,8 +48,18 @@ COPY root/usr/libexec/fix-permissions /usr/libexec/fix-permissions
...
-{%- if not spec.stable and spec.staging_repo_setup %}
-{{ spec.staging_repo_setup }}
+{%- if not spec.stable and spec.has_devel_repo %}
+  {%- if config.os.id == 'centos' and spec.has_devel_repo.centos %}
+    yum-config-manager --add-repo https://cbs.centos.org/repos/sclo7-rh-postgresql{{ spec.version }}-rh-candidate/x86_64/os/ && \
+    echo gpgcheck=0 >> /etc/yum.repos.d/cbs.centos.org_repos_sclo7-rh-postgresql{{ spec.version }}-rh-candidate_x86_64_os_.repo && \
+  {%- elif config.os.id == 'rhel' %}
+    {%- if spec.has_devel_repo.rhel == 'brew' %}
+    yum-config-manager --add-repo http://download.devel.redhat.com/brewroot/repos/rhscl-{{ spec.rhscl_version.development }}-rh-postgresql{{ spec.version }}-rhel-7-build/latest/x86_64 && \
+    echo gpgcheck=0 >> /etc/yum.repos.d/download.devel.redhat.com_brewroot_repos_rhscl-{{ spec.rhscl_version.development }}-rh-postgresql{{ spec.version }}-rhel-7-build_latest_x86_64.repo && \
```
praiskup: Hm, I'm not sure about this.  I tried to fiddle enough ... but most probably this
won't work for dist-git release process, right?  Anything could be done better
@pkubat, without giving up RHEL CI entirely?

pkubatrh: I think its ok to have this in upstream as long as its only the unreleased images that are using it. And we change it to beta repositories as soon as they are available.

praiskup:
Ping.

latest/Dockerfile.rhel7
```diff
@@ -24,13 +24,13 @@ create, run, maintain and access a PostgreSQL DBMS server."
...
       io.openshift.expose-services="5432:postgresql" \
-      io.openshift.tags="database,postgresql,postgresql96,rh-postgresql96" \
-      name="rhscl/postgresql-96-rhel7" \
-      com.redhat.component="rh-postgresql96-docker" \
-      version="1" \
-      usage="docker run -d --name postgresql_database -e POSTGRESQL_USER=user -e POSTGRESQL_PASSWORD=pass -e POSTGRESQL_DATABASE=db -p 5432:5432 rhscl/postgresql-96-rhel7" \
+      io.openshift.tags="database,postgresql,postgresql10,rh-postgresql10" \
+      name="rhscl/postgresql-10-rhel7" \
+      com.redhat.component="rh-postgresql10-docker" \
+      version="10" \
```
pkubatrh: version label should start at `1` for new images. Same for centos version.

pkubatrh:
Otherwise lgtm.

pkubatrh:
Could we merge this after getting full distgen support in #203? I think it would save @dhodovsk a lot of additional work rebasing the PR.

praiskup:
If the deadline allows, no problem with me.

praiskup:
[test][test-openshift]

praiskup:
[test-openshift]

praiskup:
I played a bit with https://github.com/sclorg/postgresql-container-generated and it seems
like PostgreSQL 9.6 based container now has "newly" `version=1`;  but IMO it should have
been that way even before...  is there a bug ATM in the git repo?

pkubatrh:
>it seems like PostgreSQL 9.6 based container now has "newly" version=1; but IMO it should have
been that way even before... is there a bug ATM in the git repo?

I remember this being discussed somewhere in the past and I think we got to the conclusion that it is not worth syncing with the downstream image (that uses `version=1`). But the current state should be the correct one if we compare to RHEL images.

The change could cause some confusion for the users since the version dropped from 9.6 to 1. But this might be a good situation to get more information about whether it is actually an issue for anyone. :)

praiskup:
[test-openshift]

praiskup:
The CI failure is not related to this PR, ptal.

praiskup:
> I remember this being discussed somewhere in the past and I think we got to the conclusion that it is not worth syncing with the downstream image (that uses version=1). But the current state should be the correct one if we compare to RHEL images.

The "downstream" term is overloaded but I got it now, I'll revert to version=9.6 for centos, and I'll set postgresql-10-centos7 to version=1 to not cause changes in postgresql-96-*.

pkubatrh:
>The "downstream" term is overloaded but I got it now, I'll revert to version=9.6 for centos, and I'll set postgresql-10-centos7 to version=1 to not cause changes in postgresql-96-*.

Seems ok. Will try to use a different term form "downstream" next time to not cause more confusion.

@hhorak are we using the rhscl-beta namepace for centos images?

praiskup:
updated:
```
postgresql-10-centos7: version=1
postgresql-10-rhel7: version=1
postgresql-96-centos7: version=9.6
postgresql-96-rhel7: version=1
everything else (older versions): according to spec.version
```

praiskup:
ping

pkubatrh:
[test-openshift]

pkubatrh:
Test fail looks like #255 again. Ran the tests on my rhel vm and no failure popped up.

LGTM, merging. Thanks!

praiskup:
Thanks, https://github.com/sclorg/postgresql-container-generated/commit/c382d83b97b12296af7136b5a696ab9c7bbfde69

PR #247, opened 2018-03-06
run_test: use ct_random_string function

praiskup:
.. which is new library function which should work fine for
environments with blocked SIGPIPE.  See discussion in:
https://github.com/sclorg/postgresql-container/commit/49a01540bb31


praiskup:
[test-openshift]

PR #251, opened 2018-03-10
Add 'postgresql-pre-start' hook

praiskup:
- also cleanup some stderr output caused by missing '../init' and
  '../boot' directory
- add missing double-quotes

### DIFF
--- a/examples/s2i-dump-data/.s2i/bin/assemble
+++ b/examples/s2i-dump-data/.s2i/bin/assemble
@@ -1,0 +1,36 @@
+#! /usr/bin/bash -x
+
+# fail early
+set -e
+
+# source the convenience tooling
+source "${CONTAINER_SCRIPTS_PATH}/common.sh"
+
+# set $PGDATA variable
+set_pgdata
+
+# assert uninitialized data
+test ! -f "$PGDATA/postgresql.conf"
+
+# empty config file is needed after 'initialize_database' call
+touch "$POSTGRESQL_CONFIG_FILE"
+
+initialize_database
+
+# start local PostgreSQL server (wait with '-w')
+pg_ctl -w start -o "-h ''"
+
+# load all sql files
+shopt -s nullglob
+for file in /tmp/src/init/*.sql; do
+    psql -f "$file"
+done
+
+pg_ctl stop
+
+# dump the data into $PWD (in-image storage)
+tar caf data.tar.xz -C "$PGDATA" .
+rm -rf "$PGDATA"
+
+# install pre-start hook
+cp -r /tmp/src/postgresql-pre-start .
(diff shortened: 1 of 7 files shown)


praiskup:
[test]

praiskup:
[test]

praiskup:
[test]

praiskup:
See the (optional and WIP) second commit, which benefits from the new hook.
With `s2i` build, users can build initial state of database into the image, and then
almost instantly start the database from the s2i image.

praiskup:
I know, the downside of `-pre-init` is (at least for postgresql case) that it evokes
that it is executed before database initialization, but it is executed every time
regardless the init is or is not executed...

pkubatrh:
I dont like pre-init either, for the same reason Pavel has already mentioned. If anything I would choose pre-start.

omron93:
I see your point.

+1 for -pre-start

praiskup:
I'll switch to `*-pre-start` then.  Btw., I'm not in love with the pattern though.

I wish we could use only one hook, say called `postgresql-start`;  and implemented
everything by "hooks"..., so we could have:
```
postgresql-start                                                                                                      
    0100_set_pgdata.sh                                                                                                
    0200_env_check.sh
    0300_generate_config.sh                                                                                           
    0400_initdb.sh
    0500_local_start.sh
    0600_update.sh
    0700_local_stop.sh
    0900_start.sh
```

If user re-defined something, or inserted something in between one "checkpoint",
it would be 100% clear what is happening.  And we would have less work in defining
semantics of each startup period..  and consistent feeling across all the images.

omron93:
> I wish we could use only one hook, say called postgresql-start; and implemented
everything by "hooks"..., so we could have:

That would definitely allow the highest degree of customization.
On the other hand new user to our images would have to know all our default files to know how to add what he wants. Now it's more clear: we say put the file here and there is high chance that it'll work.

praiskup:
> On the other hand new user to our images would have to know all our default files to know how to add what he wants.

Since all the sourced files are first printed to stdout, it is sort of trivial to find the right name
for it.  I mean, I'm OK with the somewhat artificial naming `*-pre-start` (even though one
can say "how something can happen before the container even starts).  I'll fix tomorrow.

omron93:
> how something can happen before the container even starts

It can be interpreted with database instead of container.  So -pre-start executes before database start makes sense.

praiskup:
[test-openshift]

praiskup:
[test-openshift]

praiskup:
ping

omron93:
LGTM, thanks.

pkubatrh:
lgtm as well. Fail seems unrelated so merging.

PR #258, opened 2018-03-21
pre-build: init the submodule before calling make

pkubatrh:


pkubatrh:
Needed for cccp as it does not seem to be cloning the git repositories incl. submodules by default.

praiskup:
seems fine, dunno how to test

pkubatrh:
>seems fine, dunno how to test

I do not think there is a way for us to test this other than merging the PR. Let us hope this is the last issue that needs fixing before we have the pipeline working again.

PR #263, opened 2018-03-29
tests: fix run_s2i_bake_data_test image cleanup

praiskup:
(dunno why it worked before in CI?)


pkubatrh:
lgtm. Maybe we did not see a fail for this because of the docker tag issues?

PR #268, opened 2018-04-04
Initialize generated sources branch

pkubatrh:
Including the changes made to the generated files from when distgen got introduced (in two commits - one for distgen introduction, one for the rest).

The most interesting here is probably the `update` script as that will be tool to be used to work with the branch.

@praiskup @omron93 PTAL


update
```diff
@@ -0,0 +1,58 @@
...
+    cd source
+    make generate-all
+    for i in $VERSIONS; do
+        cp -r "$i" ../
+    done
+)
+rm -rf source
+
+# Add both existing and deleted files to the index
+git add ./*
```
praiskup: This shouldn't have an effect for deleted files, the list of files is defined by shell path expansion.

update
```diff
@@ -0,0 +1,58 @@
...
+# copy the branch content into source/
+mkdir source
+for i in $COPY_PATHS; do
+    cp -r "$i" source
+done
+VERSIONS=$(sed -n 's/^VERSIONS[[:space:]]*=//p' source/Makefile)
+
+git checkout generated
+# Sometimes git does not remove common/ on branch change
+[ -d common ] && rm -rf common
```
praiskup: This wouldn't be needed if we started from scratch, by dangling commit, has this been considered?  Plus, it would be somewhat desirable in future to move the update script to common scripts.

update
```diff
@@ -0,0 +1,58 @@
...
+
+for i in $GENERATED_PATHS; do
+    case "$i" in
+    # Do not remove the update script itself
+    update)
+        continue
+        ;;
+    *)
+        rm -rf "$i"
```
praiskup: I would pay a lot more attention to this, `$GENERATED_PATHS` is an output from `ls -1`, which might well contain some spaces in dirnames, and those could lead to accidental removes, maybe that's not huge security issue...
Maybe IFS='\n' is needed.

praiskup:
We should probably somehow disable the CI for non-master branches?

update
pkubatrh: Yes, this adds all the files currently in the root to the index. The command to add deleted ones is below this line.

However, I guess it would be better to just add the content of $VERSIONS here, since that is all we care about.

update
pkubatrh: True, thanks for catching this. Will fix in next version.

update
pkubatrh: I did not see much added value to starting from scratch other than to work around the issue. But you have a point that it would make sense to add update into common scripts instead so I will do that now (since at least python will need it in near future).

omron93:
Looks good. Thanks. It would be great to have script in common.

> We should probably somehow disable the CI for non-master branches?

CI should be updated to test only PRs into master branch.

pkubatrh:
>It would be great to have script in common.

Yes, will add the update script to commons and then modify this PR accordingly.

praiskup:
> CI should be updated to test only PRs into master branch.

So testing what happens [test].

pkubatrh:
Reopened as it seems like github automagically closes PRs when there is nothing to merge.

pkubatrh:
Rebased to latest common scripts that include the update script via https://github.com/sclorg/container-common-scripts/pull/78

Commit structure is the same as in previous version of this PR

praiskup:
Looks fine, no deep study though (it's not needed IMO).  Note that merging this is probably needed to move the push-automation forward... (otherwise I would vote for waiting for having the automation in first).

pkubatrh:
Thanks. Will merge this now then to make sure the update script is available for future automation.

PR #272, opened 2018-04-16
tests: Fix centos failures due to postgresql10 being unreleased on rhel

pkubatrh:
Fixed by removing one of the if branches. Was there a reason to have it fail when an older image is available?

### DIFF
--- a/test/run-openshift
+++ b/test/run-openshift
@@ -170,7 +170,9 @@ function test_postgresql_update() {
   done
 
   if docker pull "$old_image" 2>/dev/null; then
-    $released
+    # Check if we do not have a stale unreleased versions list
+    # Fail only on rhel, on centos the image is likely already released
+    $released || [ "$OS" = "centos7" ]
   else
     # We can not test against released image.
     ! $released


pkubatrh:
[test-openshift]

test/run-openshift
```diff
@@ -169,9 +169,7 @@ function test_postgresql_update() {
     esac
   done
 
-  if docker pull "$old_image" 2>/dev/null; then
-    $released
```
praiskup: Hm, if there's no other option ... but I added the $released check on purpose, to check that released variable is set to valid value (aka that NOT_RELEASED_VERSIONS is not staled)

pkubatrh: I see, makes sense, thanks!
In that case I can add something like `$released || [ "$OS" = "centos7" ]` to make the exception centos specific (as we release centos images directly after a PR is merged)

pkubatrh:
[test-openshift]

praiskup:
lgtm

praiskup:
one nit: "Fix centos failures due to psql10" seems to be a bit confusing, under psql I see /usr/bin/psql and it makes me think what's with psql v10.0 :-)

pkubatrh:
Ah that's right, will fix the message before merging then.

PR #280, opened 2018-05-14
Deprecate version 9.4

pkubatrh:
Mention version 10 in the README

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # Variables are documented in common/build.sh.
 BASE_IMAGE_NAME = postgresql
-VERSIONS = 9.4 9.5 9.6 10
+VERSIONS = 9.5 9.6 10
 OPENSHIFT_NAMESPACES = 9.2
 NOT_RELEASED_VERSIONS =
 
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ For more information about concepts used in these container images, see the
 Versions
 ---------------
 PostgreSQL versions currently provided are:
-* [postgresql-9.4](https://github.com/sclorg/postgresql-container/tree/generated/9.4)
 * [postgresql-9.5](https://github.com/sclorg/postgresql-container/tree/generated/9.5)
 * [postgresql-9.6](https://github.com/sclorg/postgresql-container/tree/generated/9.6)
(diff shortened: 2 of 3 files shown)


praiskup:
Just check, should the generated content for 9.4 be removed or not?

pkubatrh:
Iirc the `generate` script was written as to not keep archives of deprecated versions, same as before the `postgresql:generated` branch introduction. So that should be the expected behaviour.

pkubatrh:
Seems like the 9.4 Dockerfiles were still generated even when the 9.4 version was not used in Makefile so I have removed the version from multispec as well.

praiskup:
I think that's because the 9.4 is not auto-removed, because VERSIONS= is generated against "master" branch, not against "generated" branch (as it was in https://github.com/sclorg/postgresql-container-generated/blob/c382d83b97b12296af7136b5a696ab9c7bbfde69/update)

praiskup:
Yeah, I was wrong ... it's really removed by the `rm -rf !(common|source|update)` instruction.  The problem seems to be dumped into the error output:
```
* 'git add --ignore-removal <pathspec>', which is the current default,
  ignores paths you removed from your working tree.

* 'git add --all <pathspec>' will let you also record the removals.
```

So the removals are probably not committed.

pkubatrh:
Yeah, seems like that might be the issue. I will add the `--all` option to the update script then and rebase.

pkubatrh:
Rebased to latest common scripts. Lets see if that makes the difference.

pkubatrh:
One more rebase to get in https://github.com/sclorg/container-common-scripts/pull/89

praiskup:
[test]

praiskup:
[test]
@omron93, I tried to hot-fix RHEL's CI (let's see how it works), but since I guess CentOS CI is responsible for pushing - it has to be fixed too _before_ merging this.  I don't have access there AFAIK.

praiskup:
[test]

praiskup:
[test]

praiskup:
[test]

praiskup:
[test]

praiskup:
[test]

praiskup:
LGTM

praiskup:
Generated branch is not in sync;  I'll try to fix the jenkins job and resubmit the push event.

PR #287, opened 2018-08-07
Adding postgres 10 imagestream and templates for ppc64le

jeffdyoung:
Adding postgres 10 imagestream and templates for ppc64le


hhorak:
Merging, thanks!

PR #288, opened 2018-08-13
make upgrade code more general

pkubatrh:
Introduce a way for overriding the default SCL-specific paths in `run_pgupgrade`
via `$POSTGRESQL_PREV_BINDIR` and `POSTGRESQL_BINDIR` so that images
that are not using SCLs (eg. Fedora) can run the automatic upgrade feature.

Allow any combination of previous and current version of postgresql in `run_upgrade_test`

@praiskup PTAL

### DIFF
--- a/src/root/usr/share/container-scripts/postgresql/common.sh
+++ b/src/root/usr/share/container-scripts/postgresql/common.sh
@@ -259,8 +259,8 @@ run_pgupgrade ()
     old_collection=rh-postgresql$old_raw_version
   fi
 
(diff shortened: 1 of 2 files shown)


pkubatrh:
>I'm not sure this will work nicely for Fedora, as we don't know what version we upgrade from. Plus we should install postgresql-upgrade package, e.g.

This was with `postgresql-upgrade` in mind - so only being able to upgrade from the previous version as provided by the rpm. The user would have to make sure to apply proper images in a succession if upgrading from older versions, same as in the SCL-based scenario.

omron93:
[test]

PR #289, opened 2018-08-15
update imagestreams to registry.redhat.io

bparees:


praiskup:
I'm just curious, what  is the difference between registry.access.redhat.com and registry.redhat.io?  I had to e.g. add NO_PROXY=registry.redhat.io into /etc/sysconfig/docker to make it work.

bparees:
> I'm just curious, what is the difference between registry.access.redhat.com and registry.redhat.io? 

registry.redhat.io requires credentials to pull images.  that is the only meaningful difference.  I'm not sure why you would have had to change proxy settings.

praiskup:
> registry.redhat.io requires credentials to pull images

It is not (at least on my box).

> I'm not sure why you would have had to change proxy settings.

https://access.redhat.com/solutions/3245201

I actually don't have an opinion here.  I'd like to here something like "doing <this> because of <sth>" in the commit message.  Reasons could be that we are shutting down access.redhat.com, or that registry.redhat.io is considered the default one (per some document, e.g.).

bparees:
> It is not (at least on my box).

either you've already logged in at some point, or registry.redhat.io auth was temporarily disabled for some reason, but in general it does/will.

> Reasons could be that we are shutting down access.redhat.com, or that registry.redhat.io is considered the default one (per some document

registry.redhat.io is the future of the redhat container catalog, registry.access.redhat.com is going to be shut down at some point in the future.  The new registry requires authentication and the local reference policy allows the openshift internal registry to perform that authentication on the user's behalf so that every use on a cluster doesn't have to provide their own registry.redhat.io credentials.

praiskup:
> registry.redhat.io is the future of the redhat container catalog, registry.access.redhat.com is going to be shut down at some point in the future

It would be awesome if such reasoning was in the commit message, otherwise lgtm.

bparees:
@praiskup commit message updated. ptal/merge.

praiskup:
Thanks a lot!  lgtm

PR #291, opened 2018-08-22
README: simplify searching for older documentation

praiskup:
Complements #290, so users can easily search even documentation
for older images.


pkubatrh:
lgtm, thanks!

PR #292, opened 2018-09-19
URL of OKD is docs.okd.io now

liangxia:
@bparees PTAL


praiskup:
Looks fine to me.

praiskup:
docs.openshift.org redirects to docs.okd.io

PR #294, opened 2018-10-04
tests: Do not rely on docker/podman re-using IPs in master restart test case

pkubatrh:
Podman does not re-use the same IP at all and docker documentation says it is not ensured anyway so the best thing to do is to change the master server address manually during testing.

Modifying the /etc/hosts has been chosen as podman has very limited networking capabilities compared to docker (eg. no 'podman network' or 'podman run --link ...').


praiskup:
lgtm

PR #298, opened 2018-10-24
Generate Fedora Dockerfile using distgen.

omron93:
Generate Fedora Dockerfile using distgen.

Remove release label - in Fedora it's updated automatically by
build system.

Remove arch label - 'Optional: if omitted, it will be built for
all supported Fedora Architectures'

Skip upgrade and migration test for Fedora images - uses
completely different system how to get name of an image which
contains previous version of postgresql

+ make fedora Dockerfile more similar to other OSes

@pkubatrh @praiskup Please take a look and merge.

### DIFF
--- a/specs/multispec.yml
+++ b/specs/multispec.yml
@@ -27,6 +27,9 @@ specs:
                prepare-yum-repositories rhel-server-rhscl-7-rpms && \
        rhscl_version:
          development: 3.1
+    fedora:
+      distros:
+        - fedora-29-x86_64
 
   version:
     "9.5":
@@ -69,3 +72,9 @@ specs:
       latest_fedora: "f29"
       has_devel_repo:
         centos: True
+
+matrix:
+  exclude:
+    - distros:
+        - fedora-29-x86_64
+      version: "9.5"
--- a/specs/multispec.yml
+++ b/specs/multispec.yml
@@ -40,6 +40,7 @@ specs:
       common_image_name: "{{ spec.org }}/postgresql-95-{{ spec.prod }}"
       rhel_image_name: "rhscl/postgresql-95-rhel7"
       centos_image_name: "centos/postgresql-95-centos7"
+      latest_fedora: "f25"
 
     "9.6":
       version: "9.6"
@@ -52,6 +53,7 @@ specs:
       common_image_name: "{{ spec.org }}/postgresql-96-{{ spec.prod }}"
       rhel_image_name: "rhscl/postgresql-96-rhel7"
       centos_image_name: "centos/postgresql-96-centos7"
+      latest_fedora: "f27"
 
     "10":
       version: "10"
@@ -64,5 +66,6 @@ specs:
       common_image_name: "{{ spec.org }}/postgresql-10-{{ spec.prod }}"
       rhel_image_name: "rhscl/postgresql-10-rhel7"
       centos_image_name: "centos/postgresql-10-centos7"
+      latest_fedora: "f29"
       has_devel_repo:
         centos: True
--- a/manifest.sh
+++ b/manifest.sh
@@ -13,9 +13,6 @@ DISTGEN_RULES="
 
     src=src/root/usr/share/container-scripts/postgresql/common.sh
     dest=root/usr/share/container-scripts/postgresql/common.sh;
-
-    src=src/Dockerfile.fedora
-    dest=Dockerfile.fedora;
 "
 
 # Files containing distgen directives, which are used for each
@@ -25,7 +22,10 @@ DISTGEN_MULTI_RULES="
     dest=Dockerfile;
 
     src=src/Dockerfile
-    dest=Dockerfile.rhel7
+    dest=Dockerfile.rhel7;
+
+    src=src/Dockerfile.fedora
+    dest=Dockerfile.fedora;
 "
 
 # Symbolic links
--- a/manifest.sh
+++ b/manifest.sh
@@ -12,7 +12,10 @@ DISTGEN_RULES="
     dest=root/usr/share/container-scripts/postgresql/README.md;
 
     src=src/root/usr/share/container-scripts/postgresql/common.sh
-    dest=root/usr/share/container-scripts/postgresql/common.sh
+    dest=root/usr/share/container-scripts/postgresql/common.sh;
+
+    src=src/Dockerfile.fedora
+    dest=Dockerfile.fedora;
 "
 
 # Files containing distgen directives, which are used for each
--- a/src/Dockerfile.fedora
+++ b/src/Dockerfile.fedora
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:26
+FROM registry.fedoraproject.org/{{ spec.latest_fedora }}/s2i-core:latest
 
 # PostgreSQL image for OpenShift.
 # Volumes:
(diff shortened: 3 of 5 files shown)


omron93:
[test-openshift]

praiskup:
```
$ make build OS=fedora
...
Sending build context to Docker daemon 70.66 kB
Error response from daemon: Syntax error - can't find = in "EXPOSE". Must be of the form: name=value
```

Seems like there's trailing backslash symbol in "usage" LABEL.

Otherwise I like this approach as a first step.  As a second step, it would be awesome to get
something like
```
10/Dockerfile.fedora28
10/Dockerfile.fedora29
10/Dockerfile.fedora30
```
Same as we have `Dockerfile.rhel<VERSION>`.  But this doesn't have to happen right here in this PR, and it will require other changes then in this repository.

omron93:
Thanks, fixed.

> Otherwise I like this approach as a first step.  As a second step, it would be awesome to get something like 
> ```
> 10/Dockerfile.fedora28
> 10/Dockerfile.fedora29
> 10/Dockerfile.fedora30
> ```
> Same as we have Dockerfile.rhel<VERSION>.  But this doesn't have to happen right here in this PR, and it will require other changes then in this repository.

I was thinking about it too. Definitely it would simplify taking changes from upstream into fedora git.

But it doesn't make sense for images without distgen - because it's a duplication of the same code and all of these files need to be changed in a PR.
And for some images (e.g. httpd, nginx) it would end up with a lot of fedora dockerfiles in one version-folder (sw version is same in all supported fedora releases). And all these files would be almost the same - only differing in FROM line (1-2 character diff).

So I would vote for having only one fedora dockerfile in upstream repo for earch version (using latest fedora which contains required sw version) and do rebuilds into fedora registry properly so user could always use image with fedora version he wants.

omron93:
[test-openshift]

omron93:
@praiskup Is it possible to:
- skip Dockerfile.fedora generation for 9.5/ 
- or add .exclude-fedora file only to 9.5/

?

praiskup:
> But it doesn't make sense for images without distgen - because it's a duplication of the same code and all of these files need to be changed in a PR.

Those who'd want to avoid the large duplication could just have one versioned fedora Dockerfile, say Dockerfile.fedora28 instead of Dockerfile.fedora.  At least in theory.

> @praiskup Is it possible to:

I think you only can create `.exclude-fedora` at this point, unless you plan to fix the build scripts.

omron93:
> Those who'd want to avoid the large duplication could just have one versioned fedora Dockerfile, say Dockerfile.fedora28 instead of Dockerfile.fedora. At least in theory.

It makes sense. But then the benefit is only a naming consistency with centos/rhel.

> I think you only can create .exclude-fedora at this point, unless you plan to fix the build scripts.

@praiskup  How to do this with distgen?

praiskup:
On Thursday, October 25, 2018 12:05:14 PM CEST Marek Skalický wrote:
> > Those who'd want to avoid the large duplication could just have one
> > versioned fedora Dockerfile, say Dockerfile.fedora28 instead of
> > Dockerfile.fedora. At least in theory.
> 
> It makes sense. But then the benefit is only a naming consistency with
> centos/rhel.

Not only that, because at the same time we could start supporting multiple
versions PostgreSQL server per Fedora release (if desired, e.g. in
postgresql).  It's not really easy at this point, but things will be much
more interesting once we are able to support multiple versions of
PostgreSQL server in single Fedora version (blocker here
https://pagure.io/modularity/issue/109 so far).

omron93:
@pkubatrh @praiskup Please check that excluding 9.5 Fedora dockerfile is "acceptable".

[test-openshift]

praiskup:
To me, excluding 9.5 is acceptable.

I'm not sure that this PR treats the Fedora versioning correctly; yes, it's acceptable work-around to me (similarly to the previous non-scalable layout), but it is still not something which I'd like to call "standard" so other containers are OK to C&P.  I really think that we should provide versioned fedora Dockerfiles (same as we do for centos/rhel).  **especially** in PostgreSQL case -- where we do have multiple versions of packages in one fedora release (e.g. 9.6/10 in f28) already.

omron93:
> but it is still not something which I'd like to call "standard" so other containers are OK to C&P.

How to implement best the case with separated fedora and CentOS/RHEL templates?

> I really think that we should provide versioned fedora Dockerfiles (same as we do for centos/rhel). especially in PostgreSQL case -- where we do have multiple versions of packages in one fedora release (e.g. 9.6/10 in f28) already.

For fedora registry: Then which postgresql version choose for specific Fedora version? I know that modularity allows that, but single version per fedora release can be in registry (fXY/ is in image name) - same idea proposed in sclorg/s2i-ruby-container#194.
(or other option is that the name of the container/pagure/... could contain PG version, but it doesn't scale)

The question is how many users use fedora dockerfiles from github and if in the container world OS version matters?

[test-openshift]

omron93:
[test][test-openshift]

omron93:
[test]

pkubatrh:
>single version per fedora release can be in registry

How about using tags for different versions? As in having `f29/postgresql:11-$release` and `f29/postgresql:10-$release`. You would not be able to use the `latest` tag this way but tbh it nearly never makes sense to use it.

In sclorg/s2i-ruby-container#194 you said that this is against Fedora's Container Guidelines [1] but the only mention I found is in the versioning section [2] which says that the `version` label should be 0 so that there is no chance of building an image containing a different than expected (major?) version of the rpm (in case of an rpm rebase) - but this is impossible with modularity, since for a new (major?) version of the rpm one would create a new stream (and a new image tag).

[1] https://fedoraproject.org/wiki/Container:Guidelines
[2] https://fedoraproject.org/wiki/Container:Guidelines#VERSIONING

omron93:
> In sclorg/s2i-ruby-container#194 you said that this is against Fedora's Container Guidelines [1] but the only mention I found is in the versioning section [2] which says that the version label should be 0 so that there is no chance of building an image containing a different than expected (major?) version of the rpm (in case of an rpm rebase) - but this is impossible with modularity, since for a new (major?) version of the rpm one would create a new stream (and a new image tag).

Wording is `VERSION which at this time needs to be 0`. Well, some images already sets different VERSION and I don't think this is  unbreakable rule.

What I think is more serious is Registry Layout[1] and that's the problem I meant in ruby comment: That you can't name image without `fXY` namespace.

> How about using tags for different versions? As in having f29/postgresql:11-$release and f29/postgresql:10-$release. You would not be able to use the latest tag this way but tbh it nearly never makes sense to use it.

That's true. But how to do it in pagure? Currently there is only one branch for each fedora release. It could be possible to build both images from such one branche, but rebuilding would create a real mess in git history :-)
I don't know if relengs would be able to create special branches which would allow regular container builds. Any experience?

@praiskup @pkubatrh I'm not against having more containers per fedora felease - at least to try it for some containers and see statistics. But I would prefer to have at least one working up to date container per release. Are you against merging this PR? and continue with discussion...

[1] https://fedoraproject.org/wiki/Container:Guidelines#Registry_Layout

specs/multispec.yml
```diff
@@ -27,6 +27,9 @@ specs:
                prepare-yum-repositories rhel-server-rhscl-7-rpms && \
        rhscl_version:
          development: 3.1
+    fedora:
+      distros:
+        - fedora-29-x86_64
```
praiskup: I guess this means that even `f27` is generated using `f29` config.

praiskup:
I don't want to underestimate the fact that this approach doesn't scale;  it is not only Fedora, but with RHEL/CentOS we'll face the same situation soon.  Otherwise I'm not against merging this.

omron93:
Tests are passing. Merging for now, hope that discussion about using modules to make fedora containers better is going to continue.

PR #303, opened 2018-12-04
Do not test against 9.2 image anymore

pkubatrh:
Workaround for [BZ#1655214](https://bugzilla.redhat.com/show_bug.cgi?id=1655214) to get postgresql tests passing on RHEL with Docker rpms.

We could wait for the fix in Docker rpms but do we want to test upgrades and migrations from images we have not supported for a long time anyway?

### DIFF
--- a/test/run_test
+++ b/test/run_test
@@ -637,7 +637,7 @@ run_upgrade_test ()
 {
     [ "${OS}" == "fedora" ] && return 0
 
-    local upgrade_path="none 9.2 9.4 9.5 9.6 10 none" prev= act=
+    local upgrade_path="none 9.4 9.5 9.6 10 none" prev= act=
     for act in $upgrade_path; do
         if test "$act" = $VERSION; then
             break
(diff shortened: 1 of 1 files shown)


pkubatrh:
We could also downgrade the docker rpms to docker-1.13.1-75 as a workaround in the meantime if testing against 9.2 is still needed.

praiskup:
I don't think we'd **have to** test 9.2 images, but it sort of makes sense as 9.2 is the RHEL 7 default, which could make the migration path useable. Would it make sense to drop 9.2 only temporarily?

omron93:
@pkubatrh @praiskup Any idea what could cause fedora failure: `out='rpc error: code = 2 desc = oci runtime error: exec failed: container_linux.go:247: starting container process caused "process_linux.go:110: decoding init error from pipe caused \"read parent: connection reset by peer\""` ?

pkubatrh:
Same thing, since we are running Fedora tests on a CentOS host.

praiskup:
Thanks, merged to not block other issues/PRs.

pkubatrh:
>I don't think we'd have to test 9.2 images, but it sort of makes sense as 9.2 is the RHEL 7 default, which could make the migration path useable. Would it make sense to drop 9.2 only temporarily?

Ah right I forgot rhel7 still has 9.2 so it would make sense to have a migration path via containers. Im ok with adding test case back after the docker rpms are fixed.

PR #306, opened 2018-12-13
exclude container user's id when setting up passwd file

pkubatrh:
When a container is run with a recent enough podman version and using the --user cli argument the /etc/passwd file inside the container includes a line with the user defined via --user:

```
$ podman run --rm -ti --user 1001 rhel7 bash -c "cat /etc/passwd | grep 1001"
1001:x:1001:0:container user:/:/bin/sh
```

Since postgresql creates a postgres user with the user id of the container user we need to exclude the passwd line containing said user from the generated passwd file else the initial database will not be owned by the postgres user.

### DIFF
--- a/src/root/usr/share/container-scripts/postgresql/common.sh
+++ b/src/root/usr/share/container-scripts/postgresql/common.sh
@@ -137,7 +137,7 @@ function generate_postgresql_recovery_config() {
 function generate_passwd_file() {
   export USER_ID=$(id -u)
   export GROUP_ID=$(id -g)
-  grep -v ^postgres /etc/passwd > "$HOME/passwd"
+  grep -v -e ^postgres -e ^$USER_ID /etc/passwd > "$HOME/passwd"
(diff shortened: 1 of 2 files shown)


pkubatrh:
[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
Openshift tests failing due to:

```
Error from server (Forbidden): pods is forbidden: User "developer" cannot list pods in the namespace "sclorg-test-23185": User "developer" cannot list pods in project "sclorg-test-23185"
```
That seems like some issue with the tests/openshift instance

pkubatrh:
[test-openshift]

pkubatrh:
@praiskup PTAL

praiskup:
looks fine to me

pkubatrh:
Thanks! Merging

PR #308, opened 2019-02-03
Add tests for verifying postgresql image in a remote OpenShift cluster

hhorak:
This adds a new make option `make test-openshift-remote-cluster` that runs `test/run-openshift-remote-cluster` and it runs the test within a remote OpenShift cluster. It requires `oc` command to exist and being logged in in a working cluster. The same functions are also added into `run-openshift` script, so those are tested during PR tests.


hhorak:
[test-openshift]

hhorak:
BTW this PR requires also a change in container-common-scripts: https://github.com/sclorg/container-common-scripts/pull/108/files

@praiskup @pkubatrh WDYT? I'd like to create a similar PR for other container images, but would like to see whether it is the correct way to do it first.

praiskup:
Looks like just file removal, and adding one new test which is possible to run against local or remote OpenShift cluster. If I interpret that correctly, it sounds wise - we should move as many tests as possible into the "common" tests subset... LGTM.

hhorak:
Yes, it is exactly that. We might move the rest or the functions as well, although I doubt they will "just work" on remote cluster as they are written now. It will probably need some more adjustments,  that I did not try yet. Merging now, and a new issue reported to track the porting of rest of the tests as #311.

praiskup:
thanks

Issue #309, opened 2019-02-04
PostgreSQL pod fails to restart properly if DB initialization didn't finish properly in previous pod run


iankko:
When the OpenShift cluster is under higher load (hundreds of tests running in sequential run, 3-5 tests running in parallel), it might take longer time for the PostgreSQL pod to start. If the database initialization of the pod started in 1th run, but didn't correctly finish, e.g.:

```
$ cat first_pod_log.txt 
The files belonging to this database system will be owned by user "postgres".
This user must also own the server process.

The database cluster will be initialized with locale "en_US.utf8".
The default database encoding has accordingly been set to "UTF8".
The default text search configuration will be set to "english".

Data page checksums are disabled.

fixing permissions on existing directory /var/lib/pgsql/data/userdata ... ok
creating subdirectories ... ok
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting dynamic shared memory implementation ... posix
creating configuration files ... ok
creating template1 database in /var/lib/pgsql/data/userdata/base/1 ... ok
initializing pg_authid ... ok
initializing dependencies ... ok
creating system views ... ok
loading system objects' descriptions ... ok
creating collations ... ok
creating conversions ... ok
creating dictionaries ... ok
setting privileges on built-in objects ... ok
creating information schema ... ok
loading PL/pgSQL server-side language ... ok
vacuuming database template1 ... ok
copying template1 to template0 ... ok
```
And the readiness / liveness probe "decided" to pod in question needs to be restarted, a subsequent pod will end up in _Crash Loop Back-off_ state with the msg like the following one in the pod log:

```
$ cat pod_first_restart_log.txt 
pg_ctl: another server might be running; trying to start server anyway
waiting for server to start....LOG:  redirecting log output to logging collector process
HINT:  Future log output will appear in directory "pg_log".
........................................................ done
server started
=> sourcing /usr/share/container-scripts/postgresql/start/set_passwords.sh ...
psql: FATAL: database "postgres" does not exist
```

And this scenario (try to restart the PostgreSQL pod, which subsequently fails with ```psql: FATAL: database "postgres" does not exist```) is then retried couple of times, till the default deploymentconfig timeout (600seconds IIRC) is reached.

JFTR, the aforementioned behaviour was observed with ```rhscl/postgresql-95-rhel7:latest``` image (but assuming looking at the code, the different image versions might be prone to the very same issue).

iankko:
[...]

PR #310, opened 2019-02-04
[run-postgresql] Make the DB initialization atomic

iankko:
(complete all of its parts or retry it if some failed)

Might possibly fix: https://github.com/sclorg/postgresql-container/issues/309 (Didn't test it though)

Signed-off-by: Jan Lieskovsky <jlieskov@redhat.com>

### DIFF
--- a/src/root/usr/bin/run-postgresql
+++ b/src/root/usr/bin/run-postgresql
@@ -20,7 +20,9 @@ generate_postgresql_config
 # Is this brand new data volume?
 PG_INITIALIZED=false
 
-if [ ! -f "$PGDATA/postgresql.conf" ]; then
+# Perform the DB initialization as an atomic operation,
+# IOW either complete all of its phases or redo it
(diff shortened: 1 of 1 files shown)


praiskup:
That variable is always false at that point.

iankko:
Closing, this will need to be redone yet based on the discussion in the original issue.

iankko:
Yet, re: _That variable is always false at that point._ - that was actually the intention. Always try to retry the DB init if the variable isn't set to something else than _"false"_ yet. AFAICT this should be safe (if the DB entry already exists, calling next DB init won't touch it). 

But the problem is, if not the whole DB initialization process finished before, the missing phases won't be retried in the 2nd run :(.

praiskup:
> AFAICT this should be safe (if the DB entry already exists, calling next DB init won't touch it).

But re-initdb would fail, and that would fail the whole script/container run.

PR #312, opened 2019-02-28
[KEYCLOAK-8742] [run-postgresql] Make the DB initialization atomic

iankko:
Either complete the DB init as a whole, or start it from the beginning
if the pod got restarted while DB init was in progress, but didn't finish
properly

Signed-off-by: Jan Lieskovsky <jlieskov@redhat.com>

Fixes: https://github.com/sclorg/postgresql-container/issues/309

### DIFF
--- a/src/root/usr/bin/run-postgresql
+++ b/src/root/usr/bin/run-postgresql
@@ -20,8 +20,14 @@ generate_postgresql_config
 # Is this brand new data volume?
 PG_INITIALIZED=false
 
-if [ ! -f "$PGDATA/postgresql.conf" ]; then
-  initialize_database
+if [ ! -f "${PGDATA}/postgresql.conf" ]; then
+  # Minimize the race condition of unfinished DB initialization vs pod restart
+  # First initialize the DB into temporary directory
+  TEMP_PGDATA=$(mktemp -d $(printf '%0.sX' {1..32}))
+  initialize_database "${TEMP_PGDATA}"
+  # If DB initialization succeeded, move the content of temporary directory to
+  # final PGDATA & set PG_INITIALIZED flag
+  mv -T "${TEMP_PGDATA}" "${PGDATA}"
   PG_INITIALIZED=:
 else
   try_pgupgrade
--- a/src/root/usr/share/container-scripts/postgresql/common.sh
+++ b/src/root/usr/share/container-scripts/postgresql/common.sh
@@ -153,10 +153,16 @@ initdb_wrapper ()
 }
 
 function initialize_database() {
-  initdb_wrapper initdb
+  local PG_INIT_DATA_DIR=${1:-"$PGDATA"}
+  if [ -n "$(find ${PG_INIT_DATA_DIR} -prune -empty)" ]; then
+    initdb_wrapper initdb --pgdata="${PG_INIT_DATA_DIR}"
+  else
+    >&2 echo "${FUNCNAME[0]}: ${PG_INIT_DATA_DIR} exists but is not empty"
+    exit 1
+  fi
 
   # PostgreSQL configuration.
-  cat >> "$PGDATA/postgresql.conf" <<EOF
+  cat >> "${PG_INIT_DATA_DIR}/postgresql.conf" <<EOF
 
 # Custom OpenShift configuration:
 include '${POSTGRESQL_CONFIG_FILE}'
@@ -165,7 +171,7 @@ EOF
   # Access control configuration.
   # FIXME: would be nice-to-have if we could allow connections only from
   #        specific hosts / subnet
-  cat >> "$PGDATA/pg_hba.conf" <<EOF
+  cat >> "${PG_INIT_DATA_DIR}/pg_hba.conf" <<EOF
 
 #
 # Custom OpenShift configuration starting at this point.


iankko:
@praiskup @hhorak @pkubatrh PTAL (once got a chance)

Thank you, Jan

hhorak:
[test]

hhorak:
[test-openshift]

src/root/usr/bin/run-postgresql
```diff
@@ -20,8 +20,14 @@ generate_postgresql_config
...
+  # Minimize the race condition of unfinished DB initialization vs pod restart
+  # First initialize the DB into temporary directory
+  TEMP_PGDATA=$(mktemp -d $(printf '%0.sX' {1..32}))
+  initialize_database "${TEMP_PGDATA}"
+  # If DB initialization succeeded, copy content of temporary directory to
+  # final PGDATA, remove the temporary directory & set PG_INITIALIZED flag
+  cp -prT "${TEMP_PGDATA}" "${PGDATA}" && rm -rf "${TEMP_PGDATA}"
   PG_INITIALIZED=:
 else
   try_pgupgrade
```
praiskup: @pkubatrh (brainstorm) if "initializing" has a problems with timeouts, how `try_pgupgrade` is affected?

src/root/usr/bin/run-postgresql
```diff
@@ -20,8 +20,14 @@ generate_postgresql_config
...
-if [ ! -f "$PGDATA/postgresql.conf" ]; then
-  initialize_database
+if [ ! -f "${PGDATA}/postgresql.conf" ]; then
+  # Minimize the race condition of unfinished DB initialization vs pod restart
+  # First initialize the DB into temporary directory
+  TEMP_PGDATA=$(mktemp -d $(printf '%0.sX' {1..32}))
+  initialize_database "${TEMP_PGDATA}"
+  # If DB initialization succeeded, copy content of temporary directory to
+  # final PGDATA, remove the temporary directory & set PG_INITIALIZED flag
+  cp -prT "${TEMP_PGDATA}" "${PGDATA}" && rm -rf "${TEMP_PGDATA}"
```
praiskup: I really like the hack point of this hack, thank you for the idea!

I sort of doubt that **if** I/O is blocking us and causes unreasonable delays so container timeouts **/bin/cp** will make it that better (+ even doing `rm -rf` isn't for free).  Would it be possible to rather work with `mv` and also fix the `set_pgdata` function?

src/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -153,10 +153,11 @@ initdb_wrapper ()
 }
 
 function initialize_database() {
-  initdb_wrapper initdb
+  PG_INIT_DATA_DIR="${1:-"$PGDATA"}"
```
praiskup: a=${b:-"$c"} is enough

praiskup: could we use `local`?

src/root/usr/bin/run-postgresql
iankko: @praiskup Will change _cp_ to _mv_, but what you mean under _also fix the ```set_pgdata``` function_? Just change the name, or should it also use some temporary dir?

praiskup: I was thinking about `mv "$TEMP_PGDATA" "$PGDATA"`, but for this we would have to have non-existing "$PGDATA" (so we'd have to work on `set_pgdata` a bit, it deserves it anyways).

But if there will be safe `mv` statement in subdirectory (I think) it will be good enough as well.

praiskup:
Btw., initdb refuses to work if "$PGDATA" is non empty.  We should probably keep this behaviour as well, to have at least some protection against concurrently running initdb processes (== PostgreSQL containers).

src/root/usr/bin/run-postgresql
iankko: Changed to use ```mv``` instead.

src/root/usr/share/container-scripts/postgresql/common.sh
iankko: Fixed both per your suggestions.

iankko:
@praiskup Updated the PR & squashed the commit into the original one.

Re: _Btw., initdb refuses to work if "$PGDATA" is non empty. We should probably keep this behaviour as well, to have at least some protection against concurrently running initdb processes (== PostgreSQL containers)._

Something like in the current form, is it enough?

src/root/usr/bin/run-postgresql
```diff
@@ -20,8 +20,14 @@ generate_postgresql_config
 # Is this brand new data volume?
 PG_INITIALIZED=false
 
-if [ ! -f "$PGDATA/postgresql.conf" ]; then
-  initialize_database
+if [ ! -f "${PGDATA}/postgresql.conf" ]; then
+  # Minimize the race condition of unfinished DB initialization vs pod restart
+  # First initialize the DB into temporary directory
+  TEMP_PGDATA=$(mktemp -d $(printf '%0.sX' {1..32}))
```
praiskup: Preferrably `TEMP_PGDATA=$(mktemp -d "$(printf '%0.sX' {1..32})")`.

praiskup: Btw., do we have to have the printf magic at all?

praiskup:
> Something like in the current form, is it enough?

I don't know, I'd personally prefer something stronger (an explicit check for datadir emptyness).

Though I missed one thing entirely:  Shouldn't we rather have predictable directory for db initializing, instead of temporary (of generated name)?  Seems like we otherwise could get a serious mountpoint bloat which would be never removed.

src/root/usr/bin/run-postgresql
iankko: @praiskup That printf magic is there just to have 32-times the _X_ character (the final temporary dir to be "strong enough", not guessable). But can replace it with the final form (since better to understand later).

iankko:
> > Something like in the current form, is it enough?
> 
> I don't know, I'd personally prefer something stronger (an explicit check for datadir emptyness). **EDIT:** This is OK since `mv -T` refuses to move the directory if the target is non-empty.
> 
> Though I missed one thing entirely: Shouldn't we rather have predictable directory for db initializing, instead of temporary (of generated name)? Seems like we otherwise could get a serious mountpoint bloat which would be never removed.

Yes, the temporary dir has one bottleneck now -- the log prints something like:

```
...
Success. You can now start the database server using:

    pg_ctl -D KKaop5p8x12cQlfG3qjT0PEYuBGv0vWa -l logfile start
...
```
There's this ugly temporary dir name. Will think about this a bit yet.

praiskup:
> Will think about this a bit yet.

Ping.  By "predictable" I meant something like `/var/lib/pgsql/data/init_userdata` only.  Would that work?

iankko:
Hey Pavel,

> > Will think about this a bit yet.
> 
> Ping. By "predictable" I meant something like `/var/lib/pgsql/data/init_userdata` only. Would that work?

The point of the `mktemp -d` trick is / was, by calling the `initdb` into temporary directory we didn't need to take care for `$PGDATA` directory cleanup if something went wrong during cluster initialization.

This is not possible to do when there's a need to have "predicatable" init data directory (we could use `/var/lib/pgsql/data/init_userdata` for the 1th time, but what if initdb got interrupted couple of times in a row due to e.g. network latencies?)

So it could be `/var/lib/pgsql/data/init_userdata-XXXX` but still it would require some space for flexibility (some space for `mktemp -d` to have ability to create multiple directories). So far didn't find a way, how to:

1. Still use the `mktemp -d` approach as before,
2. But make it "predictable".

But tried different approach, which seems to be working for me. See:
  https://github.com/sclorg/postgresql-container/pull/319

for the updated PR.

praiskup:
On Monday, March 11, 2019 6:37:18 PM CET Ján Lieskovský wrote:
> The point of the `mktemp -d` trick is / was, by calling the `initdb` into
> temporary directory we didn't need to take care for `$PGDATA` directory
> cleanup if something went wrong during cluster initialization.

The /var/lib/pgsql/data/initdb_userdata was meant to be temporary as well.

> This is not possible to do when there's a need to have "predicatable" init
> data directory (we could use `/var/lib/pgsql/data/init_userdata` for the 1th
> time, but what if initdb got interrupted couple of times in a row due to e.g.
> network latencies?)

If /var/lib/pgsql/data/initdb_userdata exists, you can just remove it, or?

The network latencies on NFS you speak about probably only mean that you
shouldn't run PostgreSQL on top of them, I would say...

> But tried different approach, which seems to be working for me. See:
> https://github.com/sclorg/postgresql-container/pull/319

That one is too dangerous IMO.  But others may comment as well.

iankko:
@praiskup 

>> On Monday, March 11, 2019 6:37:18 PM CET Ján Lieskovský wrote: The point of the `mktemp -d` trick is / was, by calling the `initdb` into temporary directory we didn't need to take care for `$PGDATA` directory cleanup if something went wrong during cluster initialization.

> The /var/lib/pgsql/data/initdb_userdata was meant to be temporary as well.

Ok, will have a look yet.

>> This is not possible to do when there's a need to have "predicatable" init data directory (we could use `/var/lib/pgsql/data/init_userdata` for the 1th time, but what if initdb got interrupted couple of times in a row due to e.g. network latencies?)

> If /var/lib/pgsql/data/initdb_userdata exists, you can just remove it, or?

Not completely when the PVC is NFS based. It can be done successfully (from previous testing) when it doesn't contain stale NFS file handles yet (the client tried to write something to the NFS export dir, but got interrupted). In that case, even `umount -lf /nfs/mount` will fail and call to `initdb` in the next pod launch will detect `$PGDATA` isn't empty, and fail to regenerate it.

The temporary directory approach allows to forget about the need to remove the previous invalid `$PGDATA` content (since you get new directory for which `initdb` succeeded, and only in that case copy the initialized content). The invalid content in previous directory would be left to be removed by next host restart (since `/tmp` based one).

> The network latencies on NFS you speak about probably only mean that you shouldn't run PostgreSQL on top of them, I would say...

NFS is just one PVC plugin, where this kind of issues can be easily demonstrated. But happens e.g. with GlusterFS based ones too (since in this case also have three infra nodes, where the heketi-storage pod is running. If some of the infra nodes has network latencies, PVC read/write gets slow, and you face the same issue again).

>> But tried different approach, which seems to be working for me. See: #319

> That one is too dangerous IMO. But others may comment as well.

Fair enough. Can you be more specific why?

praiskup:
> Not completely when the PVC is NFS based. It can be done successfully (from previous testing) when it doesn't contain stale NFS file handles yet (the client tried to write something to the NFS export dir, but got interrupted). 

Well, if it was initialized correctly, it wouldn't be on the */initdb_userdata path anymore, but on */userdata, or?  It is completely OK to fail if container isn't able to remove `intidb_userdata` directory,  or create it, or so.

> Can you be more specific why?

Doing `rm -rf "$PGDATA"` is just too error prone to accept, you never know where it vanishes a real data.  Doing `rm -rf /var/lib/pgsql/initdb_data` might only ever remove data which were not yet moved to it's final destination.  Even that is easy to make a mistake, but still slightly better.... we discussed it before, people might update from previous version of the image and the code there had no idea about the code in the pull-request.

praiskup:
The `TEMP_PGDATA=$(mktemp -d $(printf '%0.sX' {1..32}))`, after several initdb failures would only waste a disk space after several container-startup crashes.

praiskup:
Note that we are considering to disable livenessProbe, because it doesn't seem to be useful anyways.  Would it resolve your original problem?

iankko:
Not needed anymore (original issue got fixed by https://github.com/sclorg/postgresql-container/pull/320), closing

PR #314, opened 2019-03-04
Tighten (and document) the permission setup by fix-permissions

praiskup:
There's no need to have write-able $APP_DATA, so tighten this.
Also, I keep re-observing the reasons why we add 'postgres' user
into 'root' group all the time -- so let's have this documented as
well.

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -76,8 +76,7 @@ RUN {{ spec.environment_setup }}
     localedef -f UTF-8 -i en_US en_US.UTF-8 && \
     test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)" && \
     mkdir -p /var/lib/pgsql/data && \
-    /usr/libexec/fix-permissions /var/lib/pgsql && \
-    /usr/libexec/fix-permissions /var/run/postgresql
+    /usr/libexec/fix-permissions /var/lib/pgsql /var/run/postgresql
 
 # Get prefix path and path to scripts rather than hard-code them in scripts
 ENV CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/postgresql \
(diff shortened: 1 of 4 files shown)


src/Dockerfile
```diff
@@ -96,10 +95,22 @@ ENV BASH_ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
...
+#    Such process wouldn't be able to execute the default 'assemble' script
+#    correctly (it transitively executes 'fix-permissions' script).  So let's
+#    add the 'postgres' user into 'root' group here
+#
+# 2. we call fix-permissions on $APP_DATA here directly (UID=0 during build
+#    anyways) to assure that s2i process is actually able to _read_ the
+#    user-specified scripting.
+#
+# WARNING: this doesn't work in rootless s2i build because editing /etc/group
+# has no effect there (processes are never able to call set*id syscalls).
```
praiskup: This WARNING probably isn't valid.  I'm able to `su - ..` now in rootless container.

praiskup:
[test-openshift]

praiskup:
[test] there's some race in master-slave test, it is not first time it failed for me.  [test-openshift]

src/root/usr/libexec/fix-permissions
```diff
@@ -1,7 +1,39 @@
...
+
+uid=26
+write=w
+
+usage ()
+{
+    cat >&2 <<EOF
+$0: Error: ${1-usage error}
+
+Usage: $1 [--read-only] DIR [DIR ..]
```
pkubatrh: `$0` instead of `$1`?

praiskup: correct

praiskup:
Updated & rebased, ptal. [test][test-openshift]

praiskup:
[test][test-openshift]

praiskup:
[test-openshift]

pkubatrh:
lgtm now, thanks!

Issue #313, opened 2019-03-01
Document that migration/upgrade needs to raise openshift timeouts


praiskup:
While taking a look at #312, I realized that the liveness/readiness probes would likely kill upgrade or migration scenario for rather bulkier databases.

PR #316, opened 2019-03-06
Ensure we have enough time to do crash recovery

praiskup:
Fixes: #297

PR #320, opened 2019-03-20
make livenessProbe to return always true

praiskup:
Since we have no real deadlocks to be detected in livenessProbe
for now, lets have it (effectively) disabled for now.

The liveness probe used to cause issues before, namely because it:
- killed initdb on a rather slow storage
- would pg_upgrade for rather large data directory
- killed pod when PostgreSQL was in crash recovery

Fixes: #313, #309
Relates: #316

### DIFF
--- a/src/root/usr/libexec/check-container
+++ b/src/root/usr/libexec/check-container
@@ -9,31 +9,19 @@
 
 test -z "$ENABLED_COLLECTIONS" || . scl_source enable $ENABLED_COLLECTIONS
 
-check=ready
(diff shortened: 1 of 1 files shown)


praiskup:
@bparees, @hhorak, @pkubatrh, @kubco2 please take a look.

Please consider whether we shouldn't raise the readiness check timeout from 1s, too.

praiskup:
[test-openshift]

pkubatrh:
lgtm, thanks!

PR #327, opened 2019-04-26
Add support for POSTGRESQL_CREATEUSER_OPTIONS

atodorov:
allows to create users with --createdb permissions for example, which may be used during testing and development **WITHOUT** having to build a customized image.

The values of these variables are directly passed to `createuser`.

Background:
I am using the `centos/postgresql-10-centos7` container with Django which when running `manage.py test` will create a separate database for testing and needs sufficient permissions.

When can we expect to get this merged and a new image on Docker hub to be available?

### DIFF
--- a/src/root/usr/share/container-scripts/postgresql/README.md
+++ b/src/root/usr/share/container-scripts/postgresql/README.md
@@ -62,6 +62,9 @@ initialization by passing `-e VAR=VALUE` to the Docker run command.
 **`POSTGRESQL_USER`**  
        User name for PostgreSQL account to be created
 
+**`POSTGRESQL_CREATEUSER_OPTIONS`**
+       Options passed directly to PostgreSQL's createuser. Default: no options
+
 **`POSTGRESQL_PASSWORD`**  
        Password for the user account
 
--- a/src/root/usr/share/container-scripts/postgresql/common.sh
+++ b/src/root/usr/share/container-scripts/postgresql/common.sh
@@ -179,14 +179,16 @@ host replication all all md5
 EOF
 }
 
+export POSTGRESQL_CREATEUSER_OPTIONS=${POSTGRESQL_CREATEUSER_OPTIONS:-}
+export POSTGRESQL_MASTER_CREATEUSER_OPTIONS=${POSTGRESQL_MASTER_CREATEUSER_OPTIONS:-}
 function create_users() {
   if [[ ",$postinitdb_actions," = *,simple_db,* ]]; then
-    createuser "$POSTGRESQL_USER"
+    createuser "$POSTGRESQL_CREATEUSER_OPTIONS" "$POSTGRESQL_USER"
     createdb --owner="$POSTGRESQL_USER" "$POSTGRESQL_DATABASE"
   fi
 
   if [ -v POSTGRESQL_MASTER_USER ]; then
-    createuser "$POSTGRESQL_MASTER_USER"
+    createuser "$POSTGRESQL_MASTER_CREATEUSER_OPTIONS" "$POSTGRESQL_MASTER_USER"
   fi
 }
 


praiskup:
> The values of these variables are directly passed to createuser.

I'd suggest the variable to be named `*_CREATEUSER_* ` then...

praiskup:
Per irc chat, this is non-openshift scenario - thus s2i is a welcomed option, then I'd consider this:
```
[root@ci-vm-10-0-136-154 ~]# cat postgresql-init/test.sh 
createuser new-user --createdb # or whatever else you prefer, perhaps replace set_passwords.sh
[root@ci-vm-10-0-136-154 ~]# docker run ... -v `pwd`/postgresql-init:/opt/app-root/src/postgresql-init ...
```
The problem of the PR is that the container scripting creates the user only once (during initial deployment);  and then we "only" always "ALTER USER" in set_passwords.sh.  So if user password is changed, it get's updated..  The "createdb" permission is very similar parameter of user/role as the password is, and it IMO should be updated as well during redeployment...

atodorov:
`postgresql-init/test.sh` approach sounds promising but it fails for me:

When the contents is:
```
createuser --createdb "$POSTGRESQL_USER"
```

I get an error:
```
postgresql_database | server started
postgresql_database | /var/run/postgresql:5432 - accepting connections
postgresql_database | => sourcing /opt/app-root/src/postgresql-init/test-db-init.sh ...
postgresql_database | createuser: creation of new role failed: ERROR:  role "kiwi" already exists
```

When the contents of the init script is 
```
psql -U postgres -c "ALTER USER \"$POSTGRESQL_USER\" WITH CREATEDB;"
```

then I get another error:
```
postgresql_database | server started
postgresql_database | /var/run/postgresql:5432 - accepting connections
postgresql_database | => sourcing /opt/app-root/src/postgresql-init/test-db-init.sh ...
postgresql_database | ERROR:  role "kiwi" does not exist
```


when I add some debugging statements at the top of the init script:
```
psql -U postgres -c "\du"
```

I see no role with the reported name "kiwi" so it baffles me why it will report as already existing if I want to create it:

```
postgresql_database |                                    List of roles
postgresql_database |  Role name |                         Attributes                         | Member of 
postgresql_database | -----------+------------------------------------------------------------+-----------
postgresql_database |  postgres  | Superuser, Create role, Create DB, Replication, Bypass RLS | {}
```

pkubatrh:
You are not getting the error from inside your script (the user gets created correctly) but from the image's call to `create_users` that follows the sourcing of init scripts:

```
server started
=> sourcing /opt/app-root/src/postgresql-init/test.sh ...
++ createuser --createdb pkubat
+ break
+ read filename
+ migrate_db
+ test ,simple_db = ,migration
+ return 0
+ create_users
+ [[ ,,simple_db, = *,simple_db,* ]]
+ createuser pkubat
createuser: creation of new role failed: ERROR:  role "pkubat" already exists
```

Considering we provide a way to create users inside the init scripts, we should allow for failures inside `create_users` when the user/database already exists.

atodorov:
> Considering we provide a way to create users inside the init scripts, we should allow for failures inside create_users when the user/database already exists

How do you propose we do this? From what I can tell if the user is already created then `createuser` will fail. Will a simple if-then statement suffice here ?

praiskup:
@pkubatrh wrote:
> Considering we provide a way to create users inside the init scripts, we should allow for failures inside create_users when the user/database already exists.

Hmpfs, I don't think that users' init scripting should fight with distribution default
init scripting.

@atodorov don't you want to use `postgresql-init/set_passwords.sh` filename, instead
of `postgresql-init/test.sh` (as suggested above)?  That would effectively disable the
distribution-default set_passwords.sh script, and you would have everything under
your own control.

PR #328, opened 2019-05-06
Generated

arnabnandy1706:
I need this Dockerfile to biuld my own postgresql image


pkubatrh:
We will not be adding the generated files to the master branch.

What were you trying to do with the PR? If you just want to build the image you can generate the sources yourself locally using distgen.

PR #334, opened 2019-07-08
remove versions that are EOL

gabemontero:
per request from @siamaksade from OpenShift PM, this PR removes EOF version from the imagestream file

### DIFF
--- a/imagestreams/postgresql-centos7.json
+++ b/imagestreams/postgresql-centos7.json
@@ -27,60 +27,6 @@
         }
       },
       {
-        "name": "9.2",
-        "annotations": {
-          "openshift.io/display-name": "PostgreSQL 9.2",
-          "openshift.io/provider-display-name": "Red Hat, Inc.",
-          "description": "Provides a PostgreSQL 9.2 database on CentOS 7. For more information about using this database image, including OpenShift considerations, see https://github.com/sclorg/postgresql-container/blob/master/README.md.",
-          "iconClass": "icon-postgresql",
-          "tags": "hidden,postgresql",
-          "version": "9.2"
-        },
-        "from": {
-          "kind": "DockerImage",
-          "name": "docker.io/openshift/postgresql-92-centos7:latest"
-        },
-        "referencePolicy": {
-          "type": "Local"
-        }
-      },
-      {
-        "name": "9.4",
-        "annotations": {
-          "openshift.io/display-name": "PostgreSQL 9.4",
-          "openshift.io/provider-display-name": "Red Hat, Inc.",
-          "description": "Provides a PostgreSQL 9.4 database on CentOS 7. For more information about using this database image, including OpenShift considerations, see https://github.com/sclorg/postgresql-container/blob/master/README.md.",
-          "iconClass": "icon-postgresql",
-          "tags": "hidden,database,postgresql",
-          "version": "9.4"
-        },
-        "from": {
-          "kind": "DockerImage",
-          "name": "docker.io/centos/postgresql-94-centos7:latest"
-        },
-        "referencePolicy": {
-          "type": "Local"
-        }
-      },
-      {
-        "name": "9.5",
-        "annotations": {
-          "openshift.io/display-name": "PostgreSQL 9.5",
-          "openshift.io/provider-display-name": "Red Hat, Inc.",
-          "description": "Provides a PostgreSQL 9.5 database on CentOS 7. For more information about using this database image, including OpenShift considerations, see https://github.com/sclorg/postgresql-container/blob/master/README.md.",
-          "iconClass": "icon-postgresql",
-          "tags": "hidden,database,postgresql",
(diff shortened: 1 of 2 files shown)


gabemontero:
@hhorak PTAL

pkubatrh:
Thanks for the PR. Please remove the same versions from the CentOS imagestream file as well.

hhorak:
LGTM, although would like to rather see an ack from @bparees as well, as this is different approach from what we've used previously -- can you also confirm this is ok?

siamaksade:
@hhorak what approach was used previously?

bparees:
> @hhorak what approach was used previously?

previously i think we just marked them hidden (as these already are) in order to avoid breaking someone who had pods using these imagestreamtags in their cluster already.  They may be EOL but that's no reason to break their pods outright.

That said, the main determinant of whether removing the tags would break someone is what happens when we "update" the imagestreams in both 3.x and 4.x.  Since I don't think "oc apply" will remove existing tags, and i'm pretty sure we are effectively doing "oc apply" in both 3.x and 4.x upgrades, removing the tags is probably ok (new installs won't get them, existing installs will keep them).

That behavior should be confirmed though, before we remove these tags and potentially break a bunch of user applications in existing clusters.

siamaksade:
cc @gabemontero

hhorak:
@bparees, thanks for the feedback. @gabemontero, would you be able to confirm the behaviour @bparees describes above?

gabemontero:
On Tue, Jul 9, 2019 at 11:13 AM Honza Horak <notifications@github.com>
wrote:

> @bparees <https://github.com/bparees>, thanks for the feedback.
> @gabemontero <https://github.com/gabemontero>, would you be able to
> confirm the behaviour @bparees <https://github.com/bparees> describes
> above?
>
yep that was already on my list...just need to move it to the top

I had forgotten about the hidden vs actual remove element.

Once I’ve sorted out the imagestream goclient options and capabilities I
will report back.

But I have some business travel coming up so it might be a few days.

gabemontero:
OK, good news and bad news.

Good news:  I managed to get to it today, and confirmed that the imagestream goclient has an update and a updatestatus, which means that even if a given tag is removed from the spec, the tag still remains in the status, using the current approach of calling update on the client.

Now, if the user were to delete / recreate the imagestream, we would then lose the EOL image information in the imagestream status, as the recreate would be with the new content where the tag is removed vs. just hidden.

@bparees and I discussed this a little, and at first blush, we are OK with this, as a cluster admin would have to make this choice to delete and recreate.

But certainly chime in @siamaksade @hhorak @adambkaplan with opinions, etc. as you see fit.

Also, in testing this, I discovered that the samples operator currently does not properly deal with differing tag array lengths between spec and status ... such conditions do not occur currently, but would if we make changes like this.

So I need to fix that first even if we agree on the removal.

gabemontero:
PR https://github.com/openshift/cluster-samples-operator/pull/157 addresses the samples operator issue referenced earlier.

Once it merges, we can merge any changes here that delete tags.

gabemontero:
centos imagestream updated

gabemontero:
samples operator PR to properly handle removed tags merged .... this PR should be mergeable now @hhorak @bparees

hhorak:
@gabemontero, may I ask for some further explanation, please? I'm not familiar with "samples operator" concept much, so don't have a full picture in my mind, what happens with the image streams and templates on OpenShift side.

I know that the templates and image streams are synced periodically into https://github.com/openshift/library, and from that repo they are taken into releases. But what releases actually? Are those taken only into the future major versions (4.2, 4.3, ...) or will those be taken also into the next point releases (3.11.x)?

IOW, ensuring we don't only make one version of OpenShift happy, but all versions of OpenShift are not affected by this change negatively, so being extra cautious.

Anyway, I would feel better if even @bparees can provide +1 here and we'll merge the other PRs then.

Thanks a lot guys!

gabemontero:
Sure @hhorak - as part of the install in 4.x moving away from openshift/ansible and moving to the operator based install, we created the samples operator to install the imagestreams and templates into the openshift namespace.

We periodically transfer the artifacts from openshift/library into the samples operator payload, similar to what the openshift ansible based install does.

So wrt being sure we do not break either 3.11 or 4.x, the key point here was to nail down the client invocations used.
1) in 3.x, we already knew they used the `oc apply` invocation and that invocation would only affect the spec's of the imagestreams and not the statuses
2) in 4.x, my recent investigation confirmed that our use of the go-client code effectively did the same thing.

But yes, ultimately openshift/library will pull in these changes, and affect both active 3.11.x and 4.x releases.

I'm fine with waiting for @bparees 's +1

thanks

hhorak:
@gabemontero thanks a lot for explanation, it looks like you have it pretty well under control, so will merge the PRs, hopefully later today.

bparees:
+1

PR #336, opened 2019-07-19
Generated

joselramalho:
jkklhkljk


pkubatrh:
I expect this was filed in error, so closing.

If this was not an error, feel free to reopen with a proper comment but considering we are storing generated content in the `generated` branch for the sole reason of not having it directly in `master` I will likely close it anyway.

PR #337, opened 2019-08-01
Fix file permissions

kubco2:
This is just a fast way how to fix permissions  inside container if files on host are with incorrect permissions and probably not complete.

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -112,6 +112,11 @@ ENV BASH_ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
 RUN usermod -a -G root postgres && \
     /usr/libexec/fix-permissions --read-only "$APP_DATA"
 
+# fix incorrect permissions copied from host
+RUN chmod a+rx /usr /usr/bin /usr/libexec /usr/share
+RUN /usr/libexec/fix-permissions --read-only /usr/share/container-scripts
+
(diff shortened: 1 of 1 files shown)


kubco2:
[test]

Makefile
```diff
@@ -21,3 +21,4 @@ upgrade-tests: $(VERSIONS)
 
 build-serial: generate
 test-openshift-remote-cluster: generate
+
```
praiskup: Can you please document the reasons for this? :-)

pkubatrh:
Please add the new RUN commands to the previous one so that we do not unnecessarily add new layers.

pkubatrh:
Also this looks like it would be useful for other images as well. Maybe https://github.com/sclorg/s2i-base-container would be a better place for this?

src/Dockerfile
```diff
@@ -112,6 +112,11 @@ ENV BASH_ENV=${CONTAINER_SCRIPTS_PATH}/scl_enable \
 RUN usermod -a -G root postgres && \
     /usr/libexec/fix-permissions --read-only "$APP_DATA"
 
+# fix incorrect permissions copied from host
```
pkubatrh: Some more explanation why this is needed (what issue was caused when not present) could be added either here as a comment, or inside the commit.

PR #340, opened 2019-08-30
tests: change lower-or-equal to eq in get_image_id

pkubatrh:
We only need to work with postgresql92 image now (to test upgrade from RHEL7) and not with older images.
This fixes get_image_id for versions 10+ as they would get the "openshift3" namespaced image names generated.


pkubatrh:
Only used in upgrade/migration tests so no need to run openshift tests.

PR #342, opened 2019-09-17
 add example for enabling ssl connections

pkubatrh:


pkubatrh:
seems like some timeout happened during rhel8 tests, trying once more
[test]

pkubatrh:
@phracek Can you take a quick look at this PR?

phracek:
Definitely. I will do it.

phracek:
Based on the documentation https://www.postgresql.org/docs/9.6/app-psql.html test looks good to me.

LGTM.

phracek:
I am missing only documentaion about SSL. But this can be solved in the other pull request

phracek:
https://github.com/sclorg/postgresql-container/issues/345

pkubatrh:
Thanks for review! Tests are green, so merging.

pkubatrh:
Now that I am looking closely on the test case during working on an unrelated issue I wonder how the CI did not manage to fail when the function to run the test is called `run_s2i_ssl_enable_test` but the `TEST_LIST` entry is called `run_s2i_enable_ssl_test` ...

pkubatrh:
The masked failure should be fixed by https://github.com/sclorg/container-common-scripts/pull/125

Issue #300, opened 2018-11-27
Limit run_migration_test based on architecture


lukaszachy:
run_migration_test tries to migrate from older images. However for non-intel architectures some images are not released and the test obviously fails (e.g. `no supported platform found in manifest list`).
It would be nice to be able to limit this test to architectures, where older images exist.

pkubatrh:
Looking into this.
For some reason docker/podman is completely ok with pulling an x86_64 postgresql-92 image on a non-x86_64 host ...

PR #348, opened 2019-10-23
check if the image is available first in upgrade/migration tests

pkubatrh:
only check a subset of the migration path on non-intel architectures
as postgresql images older than v10 only support x86_64

Resolves: #300

### DIFF
--- a/test/run_test
+++ b/test/run_test
@@ -648,6 +648,8 @@ run_upgrade_test ()
         prev=$act
     done
     test "$prev" != none
+    # Check if the previous image is available in the registry
+    docker pull "$(get_image_id "$prev:remote")" || return 0
 
     # TODO: We run this script from $VERSION directory, through test/run symlink.
     test/run_upgrade_test "$prev:remote" "$VERSION:local"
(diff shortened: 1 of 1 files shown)


pkubatrh:
Note that rhel8 is missing registry.redhat.io credentials but the CI failure is getting ignored by this PR now (as we skip unavailable images)

test/run_test
```diff
@@ -648,6 +648,8 @@ run_upgrade_test ()
         prev=$act
     done
     test "$prev" != none
+    # Check if the previous image is available in the registry
+    docker pull "$(get_image_id "$prev:remote")" || return 0
```
phracek: If the previous image does not exist, then I would prefer to return 1, which could indicate a problem and we can track it in CI logs.

Just an improvement.

test/run_test
```diff
@@ -658,16 +660,19 @@ run_migration_test ()
...
 
     local from_version
-    if [ "${OS}" == "rhel8" ]; then
-        local upgrade_path="9.6 10"
-    else
+    # Only test a subset of the migration path on non-intel hosts
+    if [ "$(uname -i)" == "x86_64" ]; then
         local upgrade_path="9.2 9.4 9.5 9.6 10"
+    else
+        local upgrade_path="10"
```
phracek: This means the latest version and therefore I would prefer to use
```
local upgrade_path="$VERSION"
```

so it will not be updated during each new version.

phracek:
Only two comments. Otherwise LGTM. Also tests are passing :)

test/run_test
pkubatrh: >If the previous image does not exist, then I would prefer to return 1 ...

That is what is happening now. Which is a problem as the tests fail on architectures where previous image versions are unavailable. See #300

test/run_test
pkubatrh: This will be upgraded for each new version since the upgrade path will grow as we add new images. Check #343 for example.

pkubatrh:
>Also tests are passing :)

They are, but they should not. :)

pkubatrh:
[test]

phracek:
> > Also tests are passing :)
> 
> They are, but they should not. :)

Tests should pass and not fail. That is expected. If the tests are not passing then the container has a problem. Why the tests should fail?

pkubatrh:
@phracek is this good to go?

PR #355, opened 2019-12-03
Run all tests by default, allow to fail immediately and improve results reporting

hhorak:
Previously the test-suite failed immediately after some issue was found. This is not good for general verification, where we should see ideally full results, to see how badly the image is broken. However, for debugging purposes, it is better to see the tests failed right away, so for that purpose, there is `FAIL_QUICKLY` that can be set and tests then fail quickly.

It was also not easy to tell from the test log whether the tests failed, other than checking the exit code. Seeing a clear message that the tests failed should make it more clear.

### DIFF
--- a/test/run_test
+++ b/test/run_test
@@ -5,6 +5,9 @@
 # IMAGE_NAME specifies the name of the candidate image used for testing.
 # The image has to be available before this script is executed.
 #
+# By default, all the tests are run. For debugging purposes, FAIL_QUICKLY
+# variable can be set to 1, which makes the test-suite to fail right after any
+# of the tests fails.
 
 set -exo nounset
 shopt -s nullglob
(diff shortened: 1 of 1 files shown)


hhorak:
We should probably move some of those functions to container-common-scripts, and make the functions more generic.

test/run_test
```diff
@@ -832,10 +845,18 @@ run_s2i_bake_data_test ()
...
 
 function run_all_tests() {
+  local suite_result=0
   for test_case in $TEST_LIST; do
     : "Running test $test_case"
-    $test_case
+    if $test_case ; then
+      printf -v test_short_summary "${test_short_summary}[PASSED] $test_case\n"
+    else
+      printf -v test_short_summary "${test_short_summary}[FAILED] $test_case\n"
```
praiskup: why not:
`var="${var}..`

praiskup: ah ... the newline...  it would require `$nl` define like
```
nl='
'
test_short_summary="$test_short_summary[PASSED] $test_case$nl"

praiskup: We are in postgresql-container, in container-common-scripts this is used e.g. in generate.sh.

PR #360, opened 2020-02-04
add syspath packages for easier work with postgresql SCLs

pkubatrh:


pkubatrh:
[test-openshift]

phracek:
Fedora BASE image is a bit out (f27). Can we use `f31/s2i-core` which is the latest greatest?

pkubatrh:
>Can we use f31/s2i-core which is the latest greatest?

This is what I am working on right now. Needs a bit more work, will add the changes in a separate PR and rebase this one once it is merged.

pkubatrh:
[test-openshift]

phracek:
I know that this PR adds syspath packages for postgresql SCLs, but the question is if RHEL8 contains also -syspath packages. If so, I would add them as well.

pkubatrh:
rhel8-based images have no -syspath packages as rhel8 postgresql uses modules instead of SCLs, which means all the files are already in their expected locations.

PR #361, opened 2020-02-06
fix Fedora builds

pkubatrh:
move to latest stable (f31) s2i-core base image
use modules for postgresql builds
install pgaudit in postgresq 12 and later

Needs changes in https://github.com/sclorg/s2i-base-container/pull/199 synced to Fedora 31 first

### DIFF
--- a/src/Dockerfile.fedora
+++ b/src/Dockerfile.fedora
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/{{ spec.latest_fedora }}/s2i-core:latest
+FROM registry.fedoraproject.org/f{{ config.os.version }}/s2i-core:latest
 
 # PostgreSQL image for OpenShift.
 # Volumes:
@@ -46,6 +46,10 @@ COPY root/usr/libexec/fix-permissions /usr/libexec/fix-permissions
 # to make sure of that.
 RUN INSTALL_PKGS="rsync tar gettext bind-utils postgresql-server postgresql-contrib nss_wrapper " && \
     INSTALL_PKGS+="findutils python2 xz" && \
+{% if spec.version not in ["9.6", "10"] %}
+    INSTALL_PKGS+=" pgaudit" && \
+{% endif %}
+    dnf -y module enable postgresql:{{ spec.version }} && \
     dnf -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     dnf clean all && \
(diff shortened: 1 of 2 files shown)


pkubatrh:
s2i-core should be updated in registries now, so lets try again
[test]

pkubatrh:
```
09:10:42 Trying to pull repository registry.fedoraproject.org/f31/s2i-core ... 
09:10:43 received unexpected HTTP status: 503 Service Temporarily Unavailable
```

Weird, I have no issues pulling it locally

src/Dockerfile.fedora
```diff
@@ -46,6 +46,10 @@ COPY root/usr/libexec/fix-permissions /usr/libexec/fix-permissions
 # to make sure of that.
 RUN INSTALL_PKGS="rsync tar gettext bind-utils postgresql-server postgresql-contrib nss_wrapper " && \
     INSTALL_PKGS+="findutils python2 xz" && \
```
mkutlak: Shouldn't `python2` be replaced with `python3`, or is it still necessary?

pkubatrh:
[test]

src/Dockerfile.fedora
pkubatrh: Thanks for poiting this out. I am actually not sure if we could not get away with python3 now, but considering rhel8 vrersion does not use python2 anymore as far as I can see there should not be a problem.
Will try out.

pkubatrh:
[test]

src/Dockerfile.fedora
```diff
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/{{ spec.latest_fedora }}/s2i-core:latest
```
phracek: What about to use at least one tag from https://registry.fedoraproject.org/repo/f31/s2i-core/tags/
s2i-core w/o latest is really general.

pkubatrh: This was just a blind try to work around the weird CI failure:
```
received unexpected HTTP status: 503 Service Temporarily Unavailable
```

But seems like tags do not have an effect on this

pkubatrh:
CI finally green now. @phracek can you take a look?

src/Dockerfile.fedora
```diff
@@ -1,4 +1,4 @@
...
+FROM registry.fedoraproject.org/f{{ config.os.version }}/s2i-core:latest
```
phracek: Where is it come from? From distgen?

phracek: I see from make generate command.

pkubatrh:
Thanks for review! Merging

PR #363, opened 2020-03-14
[DO NOT MERGE] Demonstrate problem in running tests

hhorak:
This PR adds `false` call in the beginning of tests. Adding such a `false` call anywhere
in the test function should make the tests failed. It is not happening though.

This is caused by the change in running tests, when we wanted to run all tests by default.
However, when doing so, `set -e` setting does not work, because the -e is ignored in all
commands under if statement. This is the relevant documentation in the man page bash(1).

```
If a compound command or shell function executes in a context where -e is being ignored,
none of the commands executed within the compound command or function body will be affected
by the -e setting, even if -e is set and a command returns a failure status.

If a compound command or shell function sets -e while executing in a context where -e is
ignored, that setting will not have any effect until the compound command or the command
containing the function call completes.
```

### DIFF
--- a/test/run_test
+++ b/test/run_test
@@ -300,6 +300,7 @@ function try_image_invalid_combinations() {
 }
 
 function run_container_creation_tests() {
+false
   echo "  Testing image entrypoint usage"
   try_image_invalid_combinations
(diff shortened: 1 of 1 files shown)


hhorak:
A PR that removes the problematic code in #365.

fila43:
Thanks for taking care, I am closing it

PR #365, opened 2020-03-17
Remove FAIL_QUICKLY feature because it hides failures in the test functions

hhorak:
There is a problem in the way test cases could be run all, even if they failed. Running the test function inside an IF statment disables 'set -e' in the function and cannot be re-set. The problem is demonstrated in https://github.com/sclorg/postgresql-container/pull/363 which added 'false' call in the beginning of the test functions, but the test suite still passed.

This fixes a problem demonstrated in #363.

### DIFF
--- a/test/run_test
+++ b/test/run_test
@@ -5,9 +5,6 @@
 # IMAGE_NAME specifies the name of the candidate image used for testing.
 # The image has to be available before this script is executed.
 #
-# By default, all the tests are run. For debugging purposes, FAIL_QUICKLY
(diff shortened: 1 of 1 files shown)


pkubatrh:
[test]

pkubatrh:
lgtm, but trying to rerun tests

pkubatrh:
rhel8 failure looks like https://github.com/containers/libpod/issues/4326
I will fix any Fedora failures in the version 11 PR

pkubatrh:
other tests green, merging. Thanks for the fix!

PR #370, opened 2020-04-01
Turn the data_sync_retry to on for all versions

hhorak:
Because the change was done for all supported versions in upstream:
https://www.postgresql.org/about/news/1920/

Previously work-arounded for PostgreSQL v12 only:
https://github.com/sclorg/postgresql-container/pull/354

We hit this with the 10.12 version
Resolves: #1819122


pkubatrh:
LGTM, thanks a lot for looking into the issue Honza!

PR #371, opened 2020-04-07
turn on data_sync_retry in rhel8 images as well

pkubatrh:


pkubatrh:
@hhorak PTAL as you are the author of the original patch
To me it did not make much sense to apply the setting only in rhel8 images so I removed that condition

hhorak:
LGTM, it makes sense to let containers behave the same anyway.

hhorak:
I see the RHEL-8 test in Jenkins to be failing for some time already (the upgrade test part). It doesn't seem to be caused by this change.

pkubatrh:
the rhel8 fail still looks like a podman issue I managet to hit in Fedora a while ago:
https://github.com/containers/libpod/issues/4326
I expect the fix did not get to rhel8 yet

PR #373, opened 2020-04-14
add note about reserved user

bscheshirwork:


pkubatrh:
LGTM, thanks!

Doc only, so no need for tests.

PR #375, opened 2020-04-23
Move input into the container

hhorak:
Otherwise it does not work in podman 1.6.x due to https://bugzilla.redhat.com/show_bug.cgi?id=1827324

### DIFF
--- a/test/run_test
+++ b/test/run_test
@@ -60,6 +60,9 @@ add_cleanup_command ()
 '
 }
 function cleanup() {
+  # Print a big fat separator to find the error easier
+  echo "=================================== Cleanup begins here ============================="
+
   for cidfile in $CIDFILE_DIR/* ; do
     CONTAINER=$(cat $cidfile)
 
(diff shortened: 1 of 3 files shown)


test/run_test
```diff
@@ -872,19 +875,19 @@ run_pgaudit_test()
   wait_ready "$name"
 
   # enable the pgaudit extension
-  docker exec -i $(get_cid "$name") bash -c psql <<EOSQL
+  docker exec -i $(get_cid "$name") bash -c "psql <<EOSQL
```
praiskup, line 878: Hehe, what a hack :-).  Don't you want to add inline coment with the reason why the
heredoc separator went into the string? (rhbz# reference at least).

hhorak: Good idea, done.

pkubatrh:
There are still some heredocs where the same is needed. Eg.
https://github.com/sclorg/postgresql-container/blob/master/test/pg-test-lib.sh#L61

hhorak:
> There are still some heredocs where the same is needed. Eg.
> https://github.com/sclorg/postgresql-container/blob/master/test/pg-test-lib.sh#L61

Fixed now as well. Although the biggest challenge was work-arounding the `docker exec ... <file` in https://github.com/sclorg/postgresql-container/blob/master/test/pagila.sh#L37. Let's see whether this works now.

pkubatrh:
lgtm now, thanks!

PR #380, opened 2020-07-13
Add RHEL8 imagestreams

yselkowitz:


phracek:
@yselkowitz These imagestreams should be deleted https://github.com/sclorg/postgresql-container/blob/master/imagestreams/postgresql-rhel7-ppc64le.json, https://github.com/sclorg/postgresql-container/blob/master/imagestreams/postgresql-rhel7-s390x.json, right? Can you please do it in this pull request, like for the others.

yselkowitz:
If I could have, I would have.  They cannot be removed at this time because 9.6 is still supported on x86_64 but was never containerized (for RHEL 7) on P&Z.

hhorak:
Thanks for the PR, LGTM.

hhorak:
Image stream verified manually.

PR #394, opened 2020-12-18
Add example how to enable extension

hhorak:
TODO: This would deserve a test as well.

PR #398, opened 2021-02-12
WIP: Add 2 helpers for adding extension more easily

mscherer:
This PR should go along #394, that as I as a basis to get postgis in postgresql.

I used that dockerfile:
```
FROM registry.fedoraproject.org/f33/postgresql
USER 0
RUN mkdir -p postgresql-cfg && echo "shared_preload_libraries='postgis-3'" > postgresql-cfg/postgis.conf
RUN mkdir -p postgresql-init && echo -en 'echo "CREATE EXTENSION IF NOT EXISTS postgis;" | psql' > postgresql-init/enable_postgis.sh
RUN dnf install --setopt=install_weak_deps=False --best -y postgis && dnf clean all
USER 26
```

However, with the addition of the 2 helpers, I think this could be simplified to be:
```        
FROM postgres-fedora
USER 0
RUN dnf install --setopt=install_weak_deps=False --best -y postgis && dnf clean all
RUN load_pg_library postgis-3 
RUN create_extension postgis 
USER 26
```

This permit to hide implementation details, and avoid cut and paste (since postgis provides 4 extensions), and stupid errors (i took me a while to get a working container).

I marked the PR as WIP since there is no doc as I do not know where you would like to add it, and there is also no test (but I guess this can be done with pgaudit).

### DIFF
--- a/src/root/usr/bin/create_extension
+++ b/src/root/usr/bin/create_extension
@@ -1,0 +1,5 @@
+#!/bin/sh
+EXT=$1
+mkdir -p ~/postgresql-start
+echo "psql -c 'CREATE EXTENSION IF NOT EXISTS ${EXT};' \$POSTGRESQL_DATABASE" > ~/postgresql-start/enable_${EXT}.sh
+
--- a/src/root/usr/bin/load_pg_library
+++ b/src/root/usr/bin/load_pg_library
@@ -1,0 +1,4 @@
+#!/bin/sh
+LIB=$1
+mkdir -p ~/postgresql-cfg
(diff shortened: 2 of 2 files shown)


mscherer:
So, upon further experimentaion, I think the part about creating the extension may not always work (as I need to create it per database, and I do not know enough of postgres if that's a good or bad idea). I guess we could add a 3rd hook for that (create in the database). 

Is it safe to assume that the container will be used with 1 single database, and not shared accross multiple pod with multiple db ?

mscherer:
Replaced by https://github.com/sclorg/postgresql-container/pull/399

PR #403, opened 2021-04-12
Fix build to quay

phracek:
In multispec.yaml is wrong definition for CentOS7 img_name. It should be without quay.io registry.


phracek:
[test]

hhorak:
Cool, tests seem to pass, so merging. Thanks.

Issue #421, opened 2021-10-20
run_doc_test test depend on go-md2man version to work


mscherer:
It seems that on newer version of go-md2man (the one in Fedora, 2.0.1), some characters are no longer escaped, thus making the test on https://github.com/sclorg/postgresql-container/blob/master/test/run_test#L702-L707 failling when run on those platform.

This is caused by https://github.com/cpuguy83/go-md2man/commit/1c88e957a8346fc8d089a114f59a30489e4f083b who was merged for 2.0.1 version.

PR #422, opened 2021-10-20
Fix run_doc_test to work with newer go-md2man

mscherer:
Should fix https://github.com/sclorg/postgresql-container/issues/421


mscherer:
I tested that one only on a Fedora host.

phracek:
[test]

PR #427, opened 2021-11-24
Actually set MEMORY_LIMIT_IN_BYTES based on cgroups limits

jpopelka:
Hi,

in [common.sh](https://github.com/sclorg/postgresql-container/blob/60f95e96aa3fe079222ca90e41f2285a158b6f14/src/root/usr/share/container-scripts/postgresql/common.sh#L14) `POSTGRESQL_SHARED_BUFFERS` and `POSTGRESQL_EFFECTIVE_CACHE_SIZE` are calculated based on `MEMORY_LIMIT_IN_BYTES` (if set).
But I don't see `MEMORY_LIMIT_IN_BYTES` set anywhere and it doesn't seem to be set automatically in a container with memory limits specified (for example in Openshift/k8s or when you run a container via podman/docker and specify `--memory`).

FYI: I'm experimenting with memory-related settings because I'm seeing #396 (postgresql-12-centos7) and have no idea what else to try.

### DIFF
--- a/src/root/usr/share/container-scripts/postgresql/common.sh
+++ b/src/root/usr/share/container-scripts/postgresql/common.sh
@@ -2,8 +2,19 @@
 export POSTGRESQL_MAX_CONNECTIONS=${POSTGRESQL_MAX_CONNECTIONS:-100}
 export POSTGRESQL_MAX_PREPARED_TRANSACTIONS=${POSTGRESQL_MAX_PREPARED_TRANSACTIONS:-0}
 
-# Perform auto-tuning based on the container cgroups limits (only when the
-# limits are set).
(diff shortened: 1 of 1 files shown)


pkubatrh:
Hi thanks for the PR.

We are actually using the cgroups limites, but it might not be immediately obvious. The environment variables are set here: https://github.com/sclorg/postgresql-container/blob/master/src/root/usr/bin/run-postgresql#L6
And the `cgroups-limits` script is available here: https://github.com/sclorg/s2i-base-container/blob/master/core/root/usr/bin/cgroup-limits

pkubatrh:
Also @phracek might be able to help with the debugging from the container image/Openshift side, @fila43 from the postgresql side

jpopelka:
Ok, that's why I haven't found it, thanks. Closing then.

PR #428, opened 2021-11-29
Check for generated files by distgen

phracek:
This commit allows checking files generated by dist-gen.

If files are not generated properly, then the test fails.


test/run_test
```diff
@@ -16,6 +16,7 @@ shopt -s nullglob
 . test/pg-test-lib.sh
 
 TEST_LIST="\
+ct_test_distgen_generated_files
```
pkubatrh, line 19: Very much doubt the test should be run over here, part of the regular test suite. How will this work when we run the test downstream?

pkubatrh:
Would be much better to add a test for this part of the https://github.com/sclorg/container-common-scripts/blob/master/test.sh script. That way it would only be checked during the test upstream (or locally), it would fail very early (before the rest of the tests is run) and we would not have to add a test case to every single repository (a submodule update would be enough).
@phracek WDYT?

test/run_test
phracek: Good point. I have missed the downstream part. The test suite should not be a part of downstream. It does not make sense at all. Thank you.

phracek:
@pkubatrh you are right. Just update container-common-scripts and check if files are generated properly before any other tests, makes sense.

phracek:
@pkubatrh The container-common-scripts repository contains pull request where before calling any test generated files are checked. https://github.com/sclorg/container-common-scripts/pull/218

phracek:
This pull request is not needed at all. We have PR https://github.com/sclorg/container-common-scripts/pull/275 which does it automatically.

PR #434, opened 2022-03-21
use composite build-and-push-action

zmiklank:
- and rebuild container image every wednesday

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -3,133 +3,64 @@ on:
   push:
     branches:
       - master
+  schedule:
+      - cron: '0 1 * * 3'
 
 jobs:
   build-and-push:
+    if: github.repository_owner == 'sclorg'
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-push.yml
```diff
@@ -3,133 +3,64 @@ on:
...
 jobs:
   build-and-push:
+    if: github.repository_owner == 'sclorg'
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         include:
-          - version: "9.6"
+          - dockerfile_path: "9.6"
```
phracek, line 17: This is obsolete. Version 9.6 reached EOL.

phracek (changes requested):
Thanks for the PR!. Version 9.6 reached EOL.

.github/workflows/build-and-push.yml
zmiklank: I removed it.

PR #441, opened 2022-05-03
Support C8S images for postgresql-10 and postgresql13 on quay.io/sclorg

phracek:
Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>


phracek:
[test-all]

PR #448, opened 2022-06-09
Update description of POSTGRESQL_SHARED_BUFFERS & POSTGRESQL_EFFECTIVE_CACHE_SIZE default values

jpopelka:
The previous description was not completely correct, because when the container is run in Kubernetes/Openshift, the default values [are calculated](https://github.com/sclorg/postgresql-container/blob/master/src/root/usr/share/container-scripts/postgresql/common.sh#L12) based on [memory limit](https://github.com/sclorg/container-common-scripts/blob/master/shared-scripts/core/usr/bin/cgroup-limits).


phracek (approved):
Thanks for the updated postgresql documentation.

[test-all]

phracek:
OpenShift 4 failure is not caused by this PR. Let's fix it by the other PR.

PR #455, opened 2022-07-11
Postgresql-12 is RHEL 8 Full Life Application Stream

fila43:
Removing it from exclude


phracek:
[test-all]

phracek:
OpenShift 4 cluster should be up again. Let's test it 

[test-openshift]

PR #451, opened 2022-06-17
Update Dockerfile.rhel8 to fix the postgresql upgrade issue from v12 to v13.

dinaharan:
I was trying to test postgresql upgrade to version 13 from version 12 using RHEL8 based image but it's not working.  I see the following env is missing in RHEL8 based postgresql 13 image

POSTGRESQL_PREV_VERSION=12

Here is the link to problematic Dockerfile.

https://github.com/sclorg/postgresql-container/blob/generated/13/Dockerfile.rhel8

The same works fine with RHEL7 and other images.

Issue #452, opened 2022-06-23
Database version upgrade from postgresql 12 to postgresql 13 is not working 


dinaharan:
I was trying to test postgresql upgrade to version 13 from version 12 using RHEL8 based image but it's not working. I see the following env is missing in RHEL8 based postgresql 13 Dockerfile.

POSTGRESQL_PREV_VERSION=12

Here is the link to problematic Dockerfile.

https://github.com/sclorg/postgresql-container/blob/generated/13/Dockerfile.rhel8

I tried the same upgrade using RHEL7 based image and it works fine.

dinaharan:
PR #451 submitted for this issue.

PR #456, opened 2022-07-18
Update Dockerfile.rhel8 to fix the postgresql upgrade issue from v12 to v13.

dinaharan:
Updated docker file to set the POSTGRESQL_PREV_VERSION environment variable for postgresql version 13.

Here is the OLD PR: #451

### DIFF
--- a/test/run_test
+++ b/test/run_test
@@ -652,7 +652,7 @@ run_upgrade_test ()
 {
     # Do not run on Fedora or RHEL8 until the upgrade script
     # is fixed for non-SCL use cases
-    { [ "${OS}" == "fedora" ] } && return 0
+    { [ "${OS}" == "fedora" ]; } && return 0
 
     local upgrade_path="none 9.2 9.4 9.5 9.6 10 12 13 none" prev= act=
     for act in $upgrade_path; do
--- a/test/run_test
+++ b/test/run_test
@@ -652,7 +652,7 @@ run_upgrade_test ()
 {
     # Do not run on Fedora or RHEL8 until the upgrade script
     # is fixed for non-SCL use cases
-    { [ "${OS}" == "fedora" ] || [ "${OS}" == "rhel8" ]; } && return 0
+    { [ "${OS}" == "fedora" ] } && return 0
 
     local upgrade_path="none 9.2 9.4 9.5 9.6 10 12 13 none" prev= act=
     for act in $upgrade_path; do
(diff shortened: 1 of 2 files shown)


phracek:
@dinaharan Why do you create a new PR? It should be enough to update the old one. Is there any reason?

phracek:
[test-all]

phracek (approved):
The change itself looks good to me and tests are passing.

phracek:
@dinaharan As @pkubatrh mentioned the `upgrade` tests are not run for `RHEL8`  https://github.com/sclorg/postgresql-container/blob/master/test/run_test#L655. It would be nice to also fix https://github.com/sclorg/postgresql-container/blob/master/test/run_upgrade_test to support test_upgrade also for `RHEL8`.

phracek (changes requested):
This allows to upgrade into version 13. Please modify also run_upgrade_test so we have covered test suite.

hhorak:
[test]

hhorak:
Looking at the code, I'm worried the upgrade that is handled based on `$POSTGRESQL_PREV_VERSION` variable only works on RHSCL (RHEL-7), does't it? The SCL path is hardcoded in the function for example:
https://github.com/sclorg/postgresql-container/blob/705b6558/src/root/usr/share/container-scripts/postgresql/common.sh#L295

Also, I don't see we'd install `postgresql-upgrade` package anywhere, which would be needed for the upgrade to proceed. Or do I miss something here?

test/run_test
```diff
@@ -652,7 +652,7 @@ run_upgrade_test ()
 {
     # Do not run on Fedora or RHEL8 until the upgrade script
     # is fixed for non-SCL use cases
-    { [ "${OS}" == "fedora" ] || [ "${OS}" == "rhel8" ]; } && return 0
+    { [ "${OS}" == "fedora" ] } && return 0
```
hhorak, line 655: The semicolon needs to stay.

hhorak:
Anyway, #452 that requests upgrade to work is a valid request, we just need to do more than what this PR currently changes. At minimum, install `postgresql-upgrade` package and also change the `common.sh` to work with non-scl packages.

phracek:
[test-all]

phracek:
[test-all]

phracek:
[test-all]

phracek:
@dinaharan Can you please provide updated https://github.com/sclorg/postgresql-container/pull/456#issuecomment-1191370668. Otherwise we are not able to merge this pull request. Also please rebase this pull request agains master. THank you.

PR #462, opened 2022-09-14
Get rid off check for .exclude files. It does not make sense

phracek:
Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>


phracek:
[test-all]

PR #464, opened 2022-09-16
Use generic run tests

zmiklank:


zmiklank:
[test-all]

test/run_test
```diff
@@ -36,14 +36,12 @@ test -n "${IMAGE_NAME-}" || false 'make sure $IMAGE_NAME is defined'
 test -n "${VERSION-}" || false 'make sure $VERSION is defined'
 test -n "${OS-}" || false 'make sure $OS is defined'
 
-CIDFILE_DIR=$(mktemp --suffix=postgresql_test_cidfiles -d)
+CID_FILE_DIR=$(mktemp --suffix=postgresql_test_cidfiles -d)
```
phracek, line 39: Why this is needed?

phracek: This does not break functionality. And only cosmetic issue does not make sense to me.

phracek (changes requested):
Please look at my comments. I don't think, renaming variable is needed.

phracek:
[test]

phracek:
RHEL-9 Unsubscribed host is removed by #466

test/run_test
```diff
@@ -735,8 +706,6 @@ ${mount_opts}
   # need this to wait for the container to start up
   PGUSER=user     PASS=password           test_connection "$container_name"
   PGUSER=backuser PASS=pass     DB=backup test_connection "$container_name"
-
-  docker stop "$(get_cid $container_name)" >/dev/null
```
phracek, line 739: Why `docker stop` is removed from here?

phracek (changes requested):

test/run_test
zmiklank: Because it is done in `ct_cleanup`.

phracek (approved):
LGTM. Thanks for your responses. Tests are passing.

PR #472, opened 2022-10-14
Remove version 11 from build and push action

frenzymadness:
Version 11 has been removed from sources.


frenzymadness:
Version 10 might need to be removed as well?

zmiklank:
Version 10 has EOL on Dec 2022. So I would leave it there. But yes, seems like 11 could be removed, thanks.

zmiklank (approved):

frenzymadness:
Now I see where the misunderstanding is. Let me open a new PR for it.

PR #475, opened 2022-10-19
Fix distgen generation

zmiklank:
This commit should get the source code to the state before introduction of the new generator.

### DIFF
--- a/12/Dockerfile.centos7
+++ b/12/Dockerfile.centos7
@@ -1,0 +1,90 @@
+FROM quay.io/centos7/s2i-core-centos7
+
+# PostgreSQL image for OpenShift.
+# Volumes:
+#  * /var/lib/pgsql/data   - Database cluster for PostgreSQL
+# Environment:
+#  * $POSTGRESQL_USER     - Database user name
+#  * $POSTGRESQL_PASSWORD - User's password
+#  * $POSTGRESQL_DATABASE - Name of the database to create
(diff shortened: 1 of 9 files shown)


zmiklank:
@fila43 Could you please check, if all supported versions are available in the sources now?

zmiklank:
[test-all]

phracek (changes requested):
Based on the https://src.fedoraproject.org/rpms/postgresql version 10 is out of scope for Fedora. In Fedora 35 we have version 13. Therefore Dockerfile.fedora for version `10` does not make sense.

Postgresql 10 will have EOL in December 2022 for RHEL7 and therefore I do not know if want to have it. But I am not against it.

12/Dockerfile.centos7
```diff
@@ -0,0 +1,90 @@
+FROM quay.io/centos7/s2i-core-centos7
```
phracek, line 1:
```suggestion
FROM quay.io/centos7/s2i-core-centos7:latest
```

phracek: or `:centos7`. This could be better.

13/Dockerfile.centos7
```diff
@@ -0,0 +1,90 @@
+FROM quay.io/centos7/s2i-core-centos7
```
phracek, line 1:
```suggestion
FROM quay.io/centos7/s2i-core-centos7:centos7
```

phracek (changes requested):
Some nit picks.

PR #477, opened 2022-11-15
Add a test in Dockerfile directly to check the version

hhorak:
Otherwise when a wrong stream is installed, it's only found by scl test, which is too late in the process and might even go through uncought, especially when the scl test is removed for non-SCL distros at some point.


hhorak:
[test]

phracek:
[test-openshift]

phracek (approved):
Everything LGTM. Thanks.

phracek:
Let's test one more time

[test-all]

phracek:
[test-all]

phracek:
[test-all]

phracek:
OpenShift tests are failing. Let's fix them by separate PR.

Issue #296, opened 2018-10-16
Run init scripts after database creation


carlcantprogram:
Hi,

I'd like to add a schema setup script to the init directory so that basic users, permissions etc are always available in the database. The init scripts fail as the database doesn't exist yet. (only the default postgres database exists at init time). The schema setup script would read the new database name from POSTGRESQL_DATABASE and run the schema setup in that database.

commonquail:
I also have this problem. The image forces me to create a (user, password, database) but then I don't have a hook to modify that database _just once_ so I can't really bootstrap it. I could make my own user and database, then bootstrap that database, but only by setting a remote admin password, which I'd rather avoid.

In particular, I'd like to run e.g.

```sql
REVOKE ALL ON DATABASE mydb FROM public;
```

before any other user gets access to the database.

I _think_ that my specific obstacles could be solved by moving them from the `init` to the `pre-start` hook but that's not conceptually correct. I'd rather have a `post-init` hook that runs [after creating database and user](https://github.com/sclorg/postgresql-container/blob/649145a19ab1d60f184a6e9c24e868c32282f962/10/root/usr/bin/run-postgresql#L47-L48).

It came as a big surprise that `init` happens _before_ behaviour that is very difficult to opt out of.

commonquail:
Because of #325 I'm forced to create a custom image either way. Since I have to do that I've now also extracted `run-postgresql` from `centos/postgresql-10-centos7@sha256:e5be79e98d93399de3f6b6d9554efed94648b7454cd1f96f17f5af6d4651eedc` and extended it with a `post-init` hook after `create_users`. If maintainers are interested in this I can refine and submit it for inclusion.

exhuma:
@commonquail Do you have any more information on your solution? From what I can tell by going through the docker-files, the OpenShift image is not based on the official Postgres images (I might be wrong here as the Dockerfiles are templated). So the usual init-scripts won't run anyway.

Either way, I also need some way to "bootstrap" the database before a container can be run.

I have migrations which run in the context of a container execution and require a postgres extension to be present. Installing the extension requires superuser privs in Postgres which the container does not have.

I can run this manually, but it would be way cleaner to have this properly initialise when a container is created and run.

commonquail:
[...]

PR #481, opened 2022-12-13
add initialization example script

aushacker:
A simple example showing how to configure a database with some test data.

When using ephemeral storage the database name specified by POSTGRESQL_DATABASE must be different from that created in the user provided script, otherwise the container will crash loop. This is due to the default scripts hard failing when trying to create a database that the user supplied init script has already created.

See issue #296 for narrative (known problem).

### DIFF
--- a/examples/custom-data/postgresql-init/initialize.sh
+++ b/examples/custom-data/postgresql-init/initialize.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+cat <<EOF | psql
+create database quotes;
+\c quotes
+create table quotes ( id serial primary key, speaker varchar(30), quote varchar(200) );
+insert into quotes (speaker, quote) values ( 'Groucho', 'Outside of a dog, a book is man''s best friend. Inside of a dog, it''s too dark to read.' );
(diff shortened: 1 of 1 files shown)


phracek:
[test-all]

phracek (commented):
Container tests are passing. Unfortunately, I am not a database expert. @hhorak @fila43 WDYT?
As soon as @fila43 or @hhorak approve this pull request, we can merge it.

@aushacker Do we have a test for it? If not please create at least one.

aushacker:
@phracek Have had a look through the available tests, none of them seem to match the required procedure.

What is needed to test would be something like:
1. Build base image
2. Create a new image, layering in postgresql-init/initialize.sh
3. Spin up the image from 2 in podman/docker
4. Use a postgresql client to connect and query the quotes database

How does that fit in with the current testing procedure?

fila43:
I do not see any problem with proposed example

fila43:
[test-all]

phracek:
@aushacker Thank you for this pull request. Please do following steps:

* rebase it agains the master
* call `make clean-versions`
* call `make generate-all` which is mandatory, because of `distgen check` is failing. It means that generated changes are different then we expect it.

phracek (approved):
LGTM. Thanks for proposed example.

PR #488, opened 2023-01-26
Allow loading secrets from mounted files rather than ENV vars

bdunne:
This allows keys from kubernetes secrets to be mounted as files to avoid exposing them in the ENV.

### DIFF
--- a/10/root/usr/bin/run-postgresql
+++ b/10/root/usr/bin/run-postgresql
@@ -2,6 +2,15 @@
 
 export ENABLE_REPLICATION=${ENABLE_REPLICATION:-false}
 
+# Load secrets from files mounted on the filesystem and only expose them to this process.
+# This allows kubernetes to mount secrets as files in a directory rather than exposing them in the ENV.
+# For example: mount secret with key username as /run/secrets/postgresql/POSTGRESQL_USER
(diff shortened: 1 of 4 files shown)


hhorak:
Thanks for contribution, supporting secrets generally looks like a good idea to me and the code is pretty effective. What might not be perfect is to have all secrets available as ENV for the main container process. So, these are my thoughts:

We should likely allow-list variables handled like this. There might be other secrets (private/public keys, certificate) that must be handled differently than putting to the variable and also all the real secretes (passwords, private keys) should not be set for the postmaster process (should be explicitly unset), because the databases itself is so complex, that getting it from the process might be possible by unauthorized users and thus would create a security issue.

hhorak:
Plus we should have a test for this, and include it in the README.md.

hhorak:
I cannot find any guidance what path is recommended for the secrets like this, `/run/secrets/postgresql/` does not look wrong, just would like to use something as close as possible to some usual usage (if there is something like this).

hhorak:
@pkubatrh I wonder whether this might be a good learning task for Ondra or Adam (includes some small piece of research).

fila43:
[test]

fila43:
[test]

PR #490, opened 2023-02-09
Demonstrate broken tests didn't show red

hhorak:
See the list of tests include many tests:

```
TEST_LIST="\
run_container_creation_tests
run_general_tests
run_change_password_test
run_replication_test
run_master_restart_test
run_doc_test
run_s2i_test
run_test_cfg_hook
run_s2i_bake_data_test
run_s2i_enable_ssl_test
run_upgrade_test
run_migration_test
run_pgaudit_test
"
```

Then check the results log of some of the green tasks, like https://artifacts.dev.testing-farm.io/fded2fa4-7148-4f0e-8e94-adb5ae50e7e2/, which includes:

```
 [PASSED] for 'postgresql-container_tests' run_container_creation_tests (00:00:31)
 [PASSED] for 'postgresql-container_tests' run_general_tests (00:01:38)
 [PASSED] for 'postgresql-container_tests' run_change_password_test (00:00:10)
```

and not others. Also, mind the error in the log, which is just before the cleanup starts:

```
ERROR:  syntax error at or near "BREAKING"
LINE 1: BREAKING TESTS CREATE TABLE t1 (a integer); INSERT INTO t1 V...
        ^
```

This means the test suite fails in 4th test, which fails and yet the whole suite returns green.

PR #489, opened 2023-02-09
Add PostgreSQL 15

hhorak:
Except obvious changes, there is a change in providing extra privileges to the `master` user. There is a change in v15, that public schema is not accessible for non-superuser as before.

Relevant upstream change:
Remove PUBLIC creation permission on the [public schema](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PUBLIC) (Noah Misch)

PR #492, opened 2023-02-14
make sure, that testsuite fails if any test fail

zmiklank:
The set -e shell option was removed, and now the tests need to be adjusted to return valid values and do not rely on the exit on error. All of the container tests are adjusted for this behaviour in this commit.

Resolves: https://github.com/sclorg/postgresql-container/pull/490

### DIFF
--- a/test/run_test
+++ b/test/run_test
@@ -118,7 +118,6 @@ function test_connection() {
   local sleep_time=2
   for i in $(seq $max_attempts); do
     echo "    Trying to connect..."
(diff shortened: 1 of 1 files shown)


zmiklank:
[test]

hhorak:
Similar changes also done as part of #489

zmiklank:
Closing in favor of https://github.com/sclorg/postgresql-container/pull/489.

PR #494, opened 2023-02-17
Add build-and-push action config to v15

hhorak:


hhorak:
Don't want to bother people during learning day, the config looks super-obvious, so giving it a try.

PR #366, opened 2020-03-23
Allow logging to /dev/stderr

hhorak:
For easier debugging, it should be pretty easy to switch daemon to log into the /dev/stderr.
This adds such a feature by setting POSTGRESQL_LOG_DESTINATION environment variable to /dev/stderr.

This should address #353.

PR #499, opened 2023-02-28
Fixes OpenShift 4 test for postgresql-container

phracek:
This pull request fixes OpenShift 4 tests.

What is changed:
- container-common-scripts are updated
- added tests `run_latest_imagestreams_test`
- test `test_postgresql_integration` checks as ephemeral-template as persistent-template

### DIFF
--- a/test/postgresql-persistent-template.json
+++ b/test/postgresql-persistent-template.json
@@ -1,0 +1,1 @@
+../examples/postgresql-persistent-template.json
--- a/test/postgresql-persistent-template.json
+++ b/test/postgresql-persistent-template.json
@@ -1,1 +1,0 @@
-../examples/postgresql-persistent-template.json
--- a/test/test-lib-postgresql.sh
+++ b/test/test-lib-postgresql.sh
@@ -37,6 +37,13 @@ function test_postgresql_imagestream() {
   elif [ "${OS}" == "rhel9" ]; then
     tag="-el9"
   fi
+  # TODO
+  # Delete this as soon as postgresql image reached GA
+  if [ "${VERSION}" == "15" ]; then
+    echo "WARNING: Version ${VERSION} does not reach GA."
+    echo "WARNING: Remove this condition as soon as it will be available."
+    return 0
+  fi
   ct_os_test_image_stream_template "${THISDIR}/imagestreams/postgresql-${OS%[0-9]*}.json" "${THISDIR}/examples/postgresql-ephemeral-template.json" postgresql "-p POSTGRESQL_VERSION=${VERSION}${tag}"
 }
 
--- a/test/test-lib-postgresql.sh
+++ b/test/test-lib-postgresql.sh
@@ -18,7 +18,7 @@ function test_postgresql_integration() {
   postgresql-persistent-template.json"
   for template in $TEMPLATES; do
     ct_os_test_template_app_func "${IMAGE_NAME}" \
-                                 "${THISDIR}/${template}" \
+                                 "${THISDIR}/examples/${template}" \
                                  "${service_name}" \
                                  "ct_os_check_cmd_internal '<SAME_IMAGE>' '${service_name}-testing' 'PGPASSWORD=testp pg_isready -t 15 -h <IP> -U testu -d testdb' 'accepting connections' 120" \
                                  "-p POSTGRESQL_VERSION=${VERSION} \
--- a/test/test-lib-postgresql.sh
+++ b/test/test-lib-postgresql.sh
@@ -14,16 +14,19 @@ source "${THISDIR}"/test-lib-remote-openshift.sh
 
 function test_postgresql_integration() {
   local service_name=postgresql
-  ct_os_template_exists postgresql-ephemeral && t=postgresql-ephemeral || t=postgresql-persistent
-  ct_os_test_template_app_func "${IMAGE_NAME}" \
-                               "${t}" \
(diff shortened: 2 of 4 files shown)


phracek:
This pull request modifies only OpenShift test. Let's only test them.

[test-openshift]

test/test-lib-postgresql.sh
```diff
@@ -37,4 +40,16 @@ function test_postgresql_imagestream() {
   ct_os_test_image_stream_template "${THISDIR}/imagestreams/postgresql-${OS%[0-9]*}.json" "${THISDIR}/examples/postgresql-ephemeral-template.json" postgresql "-p POSTGRESQL_VERSION=${VERSION}${tag}"
 }
 
+
+function run_latest_imagestreams_test() {
```
zmiklank, line 44: This function is the same as in mysql container. Could that be rewritten so that both container images can use the same funcion?

https://github.com/sclorg/mariadb-container/blob/655aa8de41d12946b0d23c73bb06dbedb0bd3089/test/test-lib-mysql.sh#L148

test/test-lib-postgresql.sh
```diff
@@ -37,4 +40,16 @@ function test_postgresql_imagestream() {
...
+  local result=1
+  # Switch to root directory of a container
+  echo "Testing the latest version in imagestreams"
+  pushd "${THISDIR}/.." >/dev/null || return 1
+  ct_check_latest_imagestreams
+  result=$?
+  popd >/dev/null || return 1
+  return $result
+}
+
```
zmiklank, line 54: The `test_postgresql_imagestream` still fails.
Log:
```
Warning: unknown field "spec.template.spec.containers[0].capabilities"
deploymentconfig.apps.openshift.io/postgresql created
.
.
.
View details with 'oc describe <resource>/<name>' or list resources with 'oc get all'.
Build config bc/ does not exist for some reason.
Import probably failed.
project.project.openshift.io "sclorg-test-28907" deleted
Project sclorg-test-28907 was deleted properly
Test for image 'rhel8/postgresql-15:1' FAILED (exit code: 1)
```

test/postgresql-persistent-template.json
```diff
@@ -0,0 +1 @@
+../examples/postgresql-persistent-template.json
```
zmiklank, line 1: Why creating symlinks? Isn't it possible to use the templates directly from `examples` directory? So there is not that much noise in the repo root?

zmiklank (changes requested):
Thank you for your work. Please see my comments.

test/test-lib-postgresql.sh
phracek: The best way would be to move database functions to container-common-scripts. 

See issue here: https://github.com/sclorg/container-common-scripts/issues/332

In case, I will have a time, I will work on it.

But Fixing tests across https://github.com/sclorg/ organization especially OpenShift 4 tests took a more time then I have expected.

test/test-lib-postgresql.sh
phracek: This should be fixed by https://github.com/sclorg/postgresql-container/pull/499/commits/47aa17220c4214ae7f652d6ac9a97d90f933356a. We do not have yet registry.redhat.io/rhel9/postgresql-15 image yet. The same for RHEL8. postgresql-15 image is not yet generally available.

test/postgresql-persistent-template.json
phracek: Good point. I need to clarify it If it works.

test/test-lib-postgresql.sh
zmiklank: Ok.

test/test-lib-postgresql.sh
zmiklank: Should we re-run the tests then?

test/postgresql-persistent-template.json
phracek: It works, but during the syncing sources from upstream -> downstream, double symlinks are not allowed and there there is this symlink.

phracek:
[test-openshift]

test/test-lib-postgresql.sh
phracek: Already done.

zmiklank:
[test]

test/postgresql-persistent-template.json
zmiklank: What do you mean? `examples/postgresql-*-template.json` do not seem to be symlinks.

phracek:
OpenShift tests are passing. The container tests are not valid for this pull request. It should be solved by #366 .

test/postgresql-persistent-template.json
phracek: You are right, basaed on the dist-git repo . I will fix it.

phracek: This should be fixed by https://github.com/sclorg/postgresql-container/pull/499/commits/29f518a1fe81fd8650b3600c546f6f7f3af91111

phracek:
This pull request fixes OpenShift 4 tests. Container tests are not needed to run.

[test-openshift]

zmiklank (approved):

phracek:
Let's re-run one more time

[test-openshift]

phracek:
Subscriptions should be now solved.

[test-openshift]

PR #500, opened 2023-03-07
tests: use the tfaga wrapper in the testing actions

zmiklank:
And [use better error logging in container tests action](https://github.com/sclorg/postgresql-container/commit/da95f53c294f80eb6a226ee5efbe69b7f241bbd6).

### DIFF
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -41,10 +41,10 @@ jobs:
         run: |
           _result=${{ steps.check.outputs.result }}
           if [ "$_result" == failure ]; then
-            echo "Distgen-generated files are not regenerated properly."
-            echo "Please regenerate them with:"
-            echo "'make clean-versions'"
-            echo "'make generate-all'"
+            echo "::error::Distgen-generated files are not regenerated properly."
+            echo "::warning::Please regenerate them with:"
+            echo "::warning::'make clean-versions'"
+            echo "::warning::'make generate-all'"
             exit 1
           fi
   container-tests:
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -49,105 +49,24 @@ jobs:
           fi
   container-tests:
     needs: distgen-check
-    # This job only runs for '[test]' pull request comments by owner, member
     runs-on: ubuntu-20.04
-    name: "Container tests: ${{ matrix.version }} - ${{ matrix.context }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/container-tests.yml
```diff
@@ -41,113 +41,32 @@ jobs:
         run: |
           _result=${{ steps.check.outputs.result }}
           if [ "$_result" == failure ]; then
-            echo "Distgen-generated files are not regenerated properly."
-            echo "Please regenerate them with:"
-            echo "'make clean-versions'"
-            echo "'make generate-all'"
+            echo "::error::Distgen-generated files are not regenerated properly."
```
phracek, line 44: This part is missing in `.github/workflows/openshift-tests.yml`

phracek (changes requested):
Please address my comment.

.github/workflows/container-tests.yml
zmiklank: Yes, it was intended. The same is in python and varnish

Point is, that we want only one status line for distgen check. Not one from container tests and one from OCP tests.
Thus the status is being created only from container tests.

I realize, that this is not ideal. Do you have any better idea?

phracek: What about if we schedule tests only for "OpenShift"? Then dist-gen check is skipped.

phracek: I don't see any problem with duplication. distgen report should be checked always.

phracek (changes requested):

.github/workflows/container-tests.yml
zmiklank: But what if the content is not regenerated properly? Then we will be testing incorrect source codes.

zmiklank: Oh, I see, there is a check in OCP tests too, but does not produce status.

zmiklank: Please note that 'distgen check for OCP tests feature' got LGTM when it was implemented[1]. 
Also the distgen check is not connected to this PR.

If you want to change the functionality I would suggest discussing this on the container meeting and then maybe creating and issue or PR.


[1] https://github.com/sclorg/s2i-python-container/pull/558

phracek: In case of dist-gen check failed, it would be nice to have a reason, but let's not block this pull request. It is a nice to have.

phracek (approved):

PR #501, opened 2023-03-08
fix wrong nss_wrapper behavior when used with --userns=keep-id

pkubatrh:


pkubatrh:
[test]

15/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -177,7 +177,7 @@ function generate_postgresql_recovery_config() {
 function generate_passwd_file() {
   export USER_ID=$(id -u)
   export GROUP_ID=$(id -g)
-  grep -v -e ^postgres -e ^$USER_ID /etc/passwd > "$HOME/passwd"
+  grep -v -e ^postgres -e ^$USER_ID -e ^$(id -un) /etc/passwd > "$HOME/passwd"
```
phracek, line 180: We are there mixing `^postgres` `^$USER_ID` and `^$(id -un)`. What about to add there `local USERNAME=$(id -un)` so we do not mix three type of grep expressions. Only nit-picking.

And does it mean, that functionality without `userns=keep-id` works as well with this change?

phracek (commented):
The change itself LGTM. Can you please answer and/or address my comment?

15/root/usr/share/container-scripts/postgresql/common.sh
pkubatrh: We have `USER_ID` and `GROUP_ID` variables because the values are used in multiple places so it makes sense there. The username is only used in the `grep` call, so I would rather not introduce a variable for it.

>And does it mean, that functionality without userns=keep-id works as well with this change?

It should, yes. The tests are green after all. :)

phracek (approved):
Thanks for fixing this. LGTM. Feel free to merge it and think about if it feasible to introduce this change as well in https://github.com/sclorg/s2i-python-container/blob/master/src/root/opt/app-root/etc/generate_container_user#L12

PR #504, opened 2023-03-31
Add support for building and pushin to quay.io version 12

phracek:


zmiklank (approved):
Based on [1] we support postgresql 12 in RHEL8 until 2029.
LGTM.

[1] https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle

PR #510, opened 2023-05-18
This test `run_master_restart_test` is removed from the test suite

phracek:
This test is removed from the test suite.
In our CI is failing sporadically and solution is not available yet.
Let's remove it from test but the test function remains.


phracek:
[test]

fila43:
Thanks, it seems to be LGTM.

PR #519, opened 2023-07-12
add pipefail to migrate_db function

jonathankingfc:
Currently when the remote migration fails, the `grep` will catch the error and return status 0. This is unexpected behavior. Adding the `pipefail` will catch errors at any point in the pipe and fail the process.

### DIFF
--- a/10/root/usr/share/container-scripts/postgresql/common.sh
+++ b/10/root/usr/share/container-scripts/postgresql/common.sh
@@ -250,6 +250,7 @@ migrate_db ()
 
     # Migration path.
     (
+        set -o pipefail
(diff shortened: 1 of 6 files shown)


zmiklank:
Thank you for the contribution.

[test-all]

zmiklank:
> @zmiklank Are the failed tests anything to be concerned about?

No, the tests are failing because of an infrastructure issue in Testing Farm.

The change itself LGTM, however, because it changes the behavior of the container for the user I would rather wait for ack from @fila43.

fila43:
It seems to be LGTM to me. Thanks for your contribution.

PR #521, opened 2023-07-17
move pipefail outside of pg_dump block

jonathankingfc:
The pipe fail needs to be moved outside of the parenthesis in order to catch all errors inside of the code block.

### DIFF
--- a/10/root/usr/share/container-scripts/postgresql/common.sh
+++ b/10/root/usr/share/container-scripts/postgresql/common.sh
@@ -248,9 +248,9 @@ migrate_db ()
 {
     test "$postinitdb_actions" = ",migration" || return 0
 
+    set -o pipefail
     # Migration path.
     (
-        set -o pipefail
         if [ ${POSTGRESQL_MIGRATION_IGNORE_ERRORS-no} = no ]; then
             echo '\set ON_ERROR_STOP on'
         fi
--- a/12/root/usr/share/container-scripts/postgresql/common.sh
+++ b/12/root/usr/share/container-scripts/postgresql/common.sh
@@ -248,9 +248,9 @@ migrate_db ()
 {
     test "$postinitdb_actions" = ",migration" || return 0
 
+    set -o pipefail
     # Migration path.
     (
-        set -o pipefail
         if [ ${POSTGRESQL_MIGRATION_IGNORE_ERRORS-no} = no ]; then
             echo '\set ON_ERROR_STOP on'
         fi
--- a/13/root/usr/share/container-scripts/postgresql/common.sh
+++ b/13/root/usr/share/container-scripts/postgresql/common.sh
@@ -248,9 +248,9 @@ migrate_db ()
 {
     test "$postinitdb_actions" = ",migration" || return 0
 
+    set -o pipefail
     # Migration path.
     (
-        set -o pipefail
         if [ ${POSTGRESQL_MIGRATION_IGNORE_ERRORS-no} = no ]; then
             echo '\set ON_ERROR_STOP on'
         fi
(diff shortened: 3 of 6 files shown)


hhorak:
[test]

hhorak:
The failure is some infra issue.

However, this change has a bigger impact than the previous #519 because leaving it outside of the sub-shell will enable pipefail in the rest of the scripts later. I'm not aware of any specific reason this would be problem, we should just think twice before merging it, maybe checkeing more carefully those further parts that might be influenced or disable pipefail after the sub-shell.... Let me think a bit more about this...

@fila43 your opinion would be welcome as well

hhorak:
Well, giving it more thoughts, I'd rather turn off the pipefail option before the function returns, otherwise the behavior of pipes in the create_users or other scripts injected into the image that we even don't have control on.

hhorak:
So, I created https://github.com/jonathankingfc/postgresql-container/pull/1 but thought github would understand that I want to adjust this PR, which it does not. So there is a duplicate PR #522 just to see CI result... we can go with either of PRs.

hhorak:
@jonathankingfc, please, kindly verify that the suggested change still fixes the issue in your environment.

fila43:
> The failure is some infra issue.
> 
> However, this change has a bigger impact than the previous #519 because leaving it outside of the sub-shell will enable pipefail in the rest of the scripts later. I'm not aware of any specific reason this would be problem, we should just think twice before merging it, maybe checkeing more carefully those further parts that might be influenced or disable pipefail after the sub-shell.... Let me think a bit more about this...
> 
> @fila43 your opinion would be welcome as well

I have the same opinion, I was pretty fine with the `set` inside sub-shell but enabling this feature for the entire script I don't see as a good idea without a really deep investigation of the impact.

fila43:
@jonathankingfc I am not a bash expert, but I am wondering why it isn't enough to set `pipefail` inside the subshell?
Pipes are used only inside. So could you provide more information about your use case?

hhorak:
> why it isn't enough to set `pipefail` inside the subshell?

I actually understand this part -- the output of the sub-shell is then piped to `psql`, so if the sub-shell fails, it would still pass the (likely wrong) input to psql that might end successfully without `pipefail`. With `pipefail` outside of the sub-shell, the `| psql` part fails as well if the sub-shell fails.

fila43:
> > why it isn't enough to set `pipefail` inside the subshell?
> 
> I actually understand this part -- the output of the sub-shell is then piped to `psql`, so if the sub-shell fails, it would still pass the (likely wrong) input to psql that might end successfully without `pipefail`. With `pipefail` outside of the sub-shell, the `| psql` part fails as well if the sub-shell fails.

Ah, sorry the `psql` command is outside of the subshell I didn't check it correctly. So it makes definitely sense this change.

pkubatrh:
Just wondering, will the pipefail option actually get inherited by the subshell by default?

jonathankingfc:
@hhorak I've updated this PR to reflect the changes you've mentioned and have confirmed it working with our implementation. 

Reproduction steps:


```
docker run --rm -e POSTGRESQL_MIGRATION_REMOTE_HOST=invalid-host-name -e POSTGRESQL_MIGRATION_ADMIN_PASSWORD=password centos/postgresql-13-centos7 run-postgresql --version ; echo $?
```
This outputs 0 even though the hostname is invalid

With Update:
```
docker run --rm -e POSTGRESQL_MIGRATION_REMOTE_HOST=invalid-host-name -e POSTGRESQL_MIGRATION_ADMIN_PASSWORD=password updated-image:latest run-postgresql --version ; echo $?
```
This outputs 1 which is expected as the upgrade process was unsuccessful

hhorak:
> Just wondering, will the pipefail option actually get inherited by the subshell by default?

Yes, I had the same question, it should work the way that sub-shell is just a process fork (so inherits everything, unlike running a sub-command, where this would not work). I also tested that in a short script and it seems to really work this way.

jonathankingfc:
@hhorak Can we retest?

PR #523, opened 2023-07-20
Add a test suggested in PR where we added pipefail

hhorak:
https://github.com/sclorg/postgresql-container/pull/521

### DIFF
--- a/test/run_migration_test
+++ b/test/run_migration_test
@@ -56,8 +56,7 @@ assert_migration_succeeds ()
 # test case suggested in https://github.com/sclorg/postgresql-container/pull/521
 assert_migration_fails()
 {
-    docker run --rm -e POSTGRESQL_MIGRATION_REMOTE_HOST=invalid-host-name -e POSTGRESQL_MIGRATION_ADMIN_PASSWORD=password "$image_to" run-postgresql --version
-    if [ $? -eq 0 ] ; then
+    if docker run --rm -e POSTGRESQL_MIGRATION_REMOTE_HOST=invalid-host-name -e POSTGRESQL_MIGRATION_ADMIN_PASSWORD=password "$image_to" run-postgresql --version ; then
       info "The image exited with 0 despite migration obviously did not work, that's a problem."
       return 1
     else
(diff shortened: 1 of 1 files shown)


hhorak:
It's expected that this test fails before #521 is merged. we should rebase and re-run after it is merged and it should hopefully pass.

hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

test/run_migration_test
```diff
@@ -53,4 +53,17 @@ assert_migration_succeeds ()
...
 
+# test case suggested in https://github.com/sclorg/postgresql-container/pull/521
+assert_migration_fails()
+{
+    docker run --rm -e POSTGRESQL_MIGRATION_REMOTE_HOST=invalid-host-name -e POSTGRESQL_MIGRATION_ADMIN_PASSWORD=password "$image_to" run-postgresql --version
+    if [ $? -eq 0 ] ; then
+      info "The image existed with 0 despite migration obviously did not work, that's a problem."
+      return 1
+    else
+      info "The image existed with non-zero exit code when migration obviously did not work, which is correct"
```
pkubatrh, line 64:
```suggestion
      info "The image exited with 0 despite migration obviously did not work, that's a problem."
      return 1
    else
      info "The image exited with non-zero exit code when migration obviously did not work, which is correct"
```

pkubatrh:
otherwise lgtm

hhorak:
[test]

test/run_migration_test
```diff
@@ -53,4 +53,17 @@ assert_migration_succeeds ()
     container_from= container_to=
 }
 
+# test case suggested in https://github.com/sclorg/postgresql-container/pull/521
+assert_migration_fails()
+{
+    docker run --rm -e POSTGRESQL_MIGRATION_REMOTE_HOST=invalid-host-name -e POSTGRESQL_MIGRATION_ADMIN_PASSWORD=password "$image_to" run-postgresql --version
```
pkubatrh, line 59: Did not notice we still have `set -e` in here. This will need to be moved directly into the if statement below to not trigger early exit.

pkubatrh (changes requested):

test/run_migration_test
hhorak: Good point, thanks for it.

pkubatrh:
[test]

pkubatrh (approved):

PR #525, opened 2023-07-25
update imagestreams for CentOS Stream

zmiklank:


phracek (approved):
LGTM. Let's see what's happened with tests
[test-all]

zmiklank:
> LGTM. Let's see what's happened with tests [test-all]

AFAIK we do not test the centos stream imagestreams.

Issue #542, opened 2024-01-04
run_change_password_test() does not test password change


zmiklank:

hhorak:
The attempt to fix the SELinux issues #538 added creation of a new volume, that would make the test invalid, because the new container would work with empty/new data, which would not test changing password itself. However, the newly created volume directory was not used actually, so the test worked with previous data actually and likely did not fix the SELinux messages. In any case, #543 should fix this issue with SELinux messages reported originally as #536.

PR #543, opened 2024-01-04
Gracefully shut down the original container

### LINKED ISSUE #536: avc:  denied  raised during the `run_change_password_test`

hhorak:
With keeping the container running, we see SELinux error messages and a sudden crash of the PostgreSQL container. Let's properly shut down the container before using its volume privately mounted into a different container.

This should fix #536 and also reverts previous attempt (#538) to fix the issues with SELinux messages.

Background:
podman's :Z modificator for volumes works the way that the latest container run with the same volume directory mounted with :Z modificator has access, the previous containers are kept running, but access to the shared directory is suddenly removed. That caused the first instance of PostgreSQL server to crash with SIGSEGV actually, while triggering some SELinux error message during that.

Related to #542.

### DIFF
--- a/test/run_test
+++ b/test/run_test
@@ -634,6 +634,13 @@ $volume_options
 
   echo "  Changing passwords"
 
+  echo "Kill the previous container and create a new one"
+  local cidfile=$CID_FILE_DIR/"${name}"
+  docker kill $(cat $cidfile)
+  docker rm -f $(cat $cidfile)
+  # Don't forget to remove its .cid file
+  rm $cidfile
+
   DOCKER_ARGS="
 -e POSTGRESQL_DATABASE=${database}
 -e POSTGRESQL_USER=${user}
--- a/test/run_test
+++ b/test/run_test
@@ -634,11 +634,6 @@ $volume_options
 
   echo "  Changing passwords"
 
-  # create separate mounting directory for second container, as selinux does
-  # not allow two containers accesing one mounting directory if mounted with
-  # Z option
-  create_volume_dir || ret=1
-
   DOCKER_ARGS="
 -e POSTGRESQL_DATABASE=${database}
 -e POSTGRESQL_USER=${user}


hhorak:
[test]

test/run_test
```diff
@@ -634,10 +634,12 @@ $volume_options
 
   echo "  Changing passwords"
 
-  # create separate mounting directory for second container, as selinux does
-  # not allow two containers accesing one mounting directory if mounted with
-  # Z option
-  create_volume_dir || ret=1
+  echo "Kill the previous container and create a new one"
+  local cidfile=$CID_FILE_DIR/"${name}"
```
zmiklank, line 638: Do you think we could benefit from using the generic [ct_get_cid](https://github.com/sclorg/container-common-scripts/blob/1e82b93382d056dd17b90b8e0e637e5bfb9cbb88/test-lib.sh#L335) function from test-lib here?

zmiklank (commented):
Looks good, thanks for the thorough investigation and for great commit messages!

I have just one small note, see in inline comments.

test/run_test
hhorak: Yes, it makes sense to me, although it would be better to be done in the whole script to keep it consistent. I'd not combine it to this PR now, but might create a separate PR for this and other functions from test-lib that are duplicated here.

hhorak: Some small refactoring prepared here: https://github.com/sclorg/postgresql-container/pull/545 but this and #544 must be merged before, I built those changes based on the commits here in order to avoid merge conflicts later.

zmiklank: Agreed.

zmiklank (approved):

PR #544, opened 2024-01-04
Do not ignore SQL failures and check that we still work with previous data when changing password

hhorak:
Fix also postgresql_cmd function that should not ignore SQL errors (thus using ON_ERROR_STOP option now) and was often called with an interactive input using <<< but podman run is run in non-interactive mode, resulting in the SQL commands to not be printed, so checking them later did not work and errors are not visible. Using `-c` option solves this.

Function test_postgresql was called with a parameter that was not used anywhere. This function used `CREATE EXTENSION` to test proper PostgreSQL functionality, that only admin have permissions for.
So, the parameter of test_postgresql function was changed to have a useful meaning. The meaning is that only if the parameter is "admin", the `CREATE EXTENSION` statement is run to not see
"Not enough permissions" error when the contianer is run as a normal user.

### DIFF
--- a/test/run_test
+++ b/test/run_test
@@ -107,7 +107,7 @@ function get_ip_from_cid() {
 }
 
 function postgresql_cmd() {
-  docker run --rm -e PGPASSWORD="$PASS" "$IMAGE_NAME" psql "postgresql://$PGUSER@$CONTAINER_IP:5432/${DB-db}" "$@"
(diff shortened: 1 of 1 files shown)


hhorak:
[test]

hhorak:
[test]

zmiklank (approved):
Thanks for this PR.

LGTM. I would consider adding '-At -c' params into postgresql_cmd function directly (even though I noticed, that some commands do not use the '-At' params currently), but I am also OK with this solution.

hhorak:
> LGTM. I would consider adding '-At -c' params into postgresql_cmd function directly (even though I noticed, that some commands do not use the '-At' params currently), but I am also OK with this solution.

I'd rather leave it flexible, so the caller of `postgresql_cmd` can decide what format and way of input to use.

PR #549, opened 2024-01-23
add YAML config for imagestreams generator

zmiklank:
Depends on: https://github.com/sclorg/ci-scripts/pull/119

PR #553, opened 2024-03-12
Add missing imagestreams for Postgresql 16

phracek:
This pull request adds missing imagestreams for Postgresql16.

### DIFF
--- a/test/test-lib-postgresql.sh
+++ b/test/test-lib-postgresql.sh
@@ -18,10 +18,14 @@ function test_postgresql_integration() {
     namespace_image="rhscl/postgresql-${VERSION}-rhel7"
   else
     namespace_image="${OS}/postgresql-${VERSION}"
-    # In case we test postgresql-16 that is not released yet
-    # let's use postgresql-15
-    if [ "${VERSION}" == "16" ]; then
-      namespace_image="${OS}/postgresql-15"
+    # Check if the current version is already GA
+    # This directory is cloned from TMT plan repo 'sclorg-tmt-plans'
+    local devel_file="/root/sclorg-tmt-plans/devel_images"
+    if [ -f "${devel_file}" ]; then
+      if grep -q "${OS}=postgresql-image=${VERSION}" "$devel_file" ; then
+        echo "This version is currently developed, so skipping this test."
+        return
+      fi
     fi
   fi
   TEMPLATES="postgresql-ephemeral-template.json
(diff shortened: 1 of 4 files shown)


phracek:
This pull request adds two missing imagestreams. Only running OpenShift tests are mandatory

[test-openshift]

zmiklank:
Hi, can we do this change with the generator?

I can do that if you want.

https://github.com/sclorg/postgresql-container/pull/549

phracek:
> you

Just let me know or write how.

phracek:
[test-openshift]

test/test-lib-postgresql.sh
```diff
@@ -18,6 +18,11 @@ function test_postgresql_integration() {
     namespace_image="rhscl/postgresql-${VERSION}-rhel7"
   else
     namespace_image="${OS}/postgresql-${VERSION}"
+    # In case we test postgresql-16 that is not released yet
+    # let's use postgresql-15
+    if [ "${VERSION}" == "16" ]; then
```
zmiklank, line 23: Can we wait with adding imagestreams for v16 instead of introducing this fix?
There is high possibility we will forget about this in the future.
Or let's have the tests failing for a while, I am OK with it.

phracek: Of course, but nightly builds are failing. I check them each day or two. It is boring. For this reason, we can stop it.

Or I can file an issue to remove it.

zmiklank: Issue here on sclorg's Github? If yes, then let's remove them in https://github.com/sclorg/postgresql-container/pull/549. It is only a simple change in the imagestream.yaml.

phracek: Removing from #549 is not the right way. I will try to explain where the trouble is.

We have to do two things for each new container version in the upstream repository.
First of all, add a new version here, and second, add a new version into imagestreams as well.

But this is a problem. When we test the newest version in OpenShift 4, like in this case `registry.redhat.io/rhel8/postgresql-16`,  then the OpenShift tests fail because this imagestream does not exist yet. It is still not generally available

But OpenShift 4 team gets regularly all our imagestreams from our repositories to the newly developed OCP4 version, like OCP 4.16 now.
When the latest imagestreams are missing in our repositories, then these imagestreams are also missing in OpenShift developed versions, which is not expected.
We can of course test in our test suites if 'developed' version is present in `devel_images` file and then skip it. 
This solution is feasible and realistic not only for this pull request but also for the later-on new images.

zmiklank: But until the image is out the imagestreams will not work for the OpenShift team too.

>We can of course test in our test suites if 'developed' version is present in devel_images file and then skip it.
This solution is feasible and realistic not only for this pull request but also for the later-on new images.

Sounds sensible to me. I would also be OK with adding the imagestreams when the image is ready (the correct timing could be a problem here) OR let the tests fail until then. But I would rather avoid the option of testing imagestreams for pgsql16 with pgsql 15 images.

phracek: I have fixed it by commit https://github.com/sclorg/postgresql-container/pull/553/commits/f8e6b12e83cc685783540c458b25ee4f473d0d7e

phracek:
[test-openshift]

phracek:
[test-openshift]

phracek:
[test-openshift]

phracek:
[test-openshift]

test/test-lib-postgresql.sh
phracek: It looks like everything is greeen.

zmiklank (approved):

Issue #550, opened 2024-02-19
Upgrade fails due to postmaster.pid


eifrach:
### Container platform

OCP 4, Podman/Docker

### Version

12, 13

### OS version of the container image

RHEL 7, RHEL 8, CentOS 7, CentOS Stream 8

### Bugzilla, Jira

_No response_

### Description

We are trying to work on and upgrade flow from version 12 to version 13
we are using k8s deployment  pods 
when we try to deploy the new image we get an error during the upgrade 

```
==========  $PGDATA upgrade: 12 -> 13  ==========                                                                                                                                                 
                                                                                                                                                                                                  
                                                                                                                                                                                                  
===>  Starting old postgresql once again for a clean shutdown...                                                                                                                                  
                                                                                                                                                                                                  
pg_ctl: another server might be running; trying to start server anyway                                                                                                                            
waiting for server to start....2024-02-14 12:46:35.514 UTC [31] FATAL:  lock file "postmaster.pid" already exists                                                                                 
2024-02-14 12:46:35.514 UTC [31] HINT:  Is another postmaster (PID 1) running in data directory "/var/lib/pgsql/data/userdata"?           
```

It seems that the `postmaster.pid` is always there also on restart / redeployment and the service recovers without an issue. Only during the upgrade do we face the problem.

deleting the file and redeploying the upgrade solves it. 



my question is,
1.  Is that the normal behavior  ?
2.  What is the best approach to do deployment upgrades? 

not sure if deleting a PID is good solution to run in Production enviorment  




### Reproducer

deploy version 12, and redeploy version 13 with the upgrade env

can be done with `podman play` or `minikube`

fila43:
Hello, thank you for your report. Since every layer (podman, openshift,..) complicates the debugging, it would be really helpful to provide a reproducer just using podman or on rpm level.

eifrach:
BTW, when you restart the image you can see that the `postmater.pid` is there even without the upgrade 
it's not been deleted when shutdown the postgres services

fila43:
It seems to be caused by the orchestration. I tested the scenario just by deploying the image by podman and then redeploying and the upgrade works fine.

eifrach:
1. make sure you are using persistent data, could be  you are not migrating the DB.
2. make soure you have the `postmaster.pid`  created before starting the upgrade.
 
but I did it with podman and minikube and openshift.

fila43:
[...]

PR #554, opened 2024-03-13
The .pid file persists after shut down

fila43:
.pid is present during postgresql-server run. It should be removed with server stop. in some circumstances, it persists after the container shuts down. If it happens, this check should ensure the state is correct before the upgrade. It should remove the redundant file.

### DIFF
--- a/10/root/usr/share/container-scripts/postgresql/common.sh
+++ b/10/root/usr/share/container-scripts/postgresql/common.sh
@@ -296,6 +296,11 @@ function wait_for_postgresql_master() {
 
 run_pgupgrade ()
 (
+  # Remove .pid file if the file persists after ugly shut down
+  if [ -f "$PGDATA/postmaster.pid" ] && ! pgrep -f "postgres" > /dev/null; then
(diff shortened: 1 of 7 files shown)


fila43:
related https://github.com/sclorg/postgresql-container/issues/550

zmiklank:
[test-all]

10/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -296,6 +296,11 @@ function wait_for_postgresql_master() {
 
 run_pgupgrade ()
 (
+  # Remove .pid file if the file persists after ugly shut down
+  if [ -f "$PGDATA/postmaster.pid" ] && ! ps -aux | awk '{print $11}' | grep -q postgres; then
```
zmiklank, line 300: What do we want to check with this condition?

Is it if postgres process is running? If yes, could we use pgrep for that instead of ps?

zmiklank (approved):

PR #556, opened 2024-03-20
Remove testing postgresql-container on OpenShift 3

phracek:
Remove testing postgresql-container on OpenShift 3


phracek:
[test-openshift]

PR #562, opened 2024-04-11
Upgrade flow for images RHEL/CentOS 8+

eifrach:
Creating flow for upgrading postgresSQL 

1. add `install_weak_deps=false` to installation which reduce the image size
2. install `postgresql-upgrade` package for the upgrade

### DIFF
--- a/13/root/usr/share/container-scripts/postgresql/common.sh
+++ b/13/root/usr/share/container-scripts/postgresql/common.sh
@@ -297,7 +297,7 @@ function wait_for_postgresql_master() {
 run_pgupgrade ()
 (
   # Remove .pid file if the file persists after ugly shut down
-  if [ -f "$PGDATA/postmaster.pid" ] && ! pgrep -f "postgres" > /dev/null; then
+  if [ -f "$PGDATA/postmaster.pid" ] && ! pg_isready > /dev/null; then
     rm -rf "$PGDATA/postmaster.pid"
   fi
 
@@ -311,8 +311,17 @@ run_pgupgrade ()
     old_collection=rh-postgresql$old_raw_version
   fi
 
-  old_pgengine=/opt/rh/$old_collection/root/usr/bin
-  new_pgengine=/opt/rh/rh-postgresql${new_raw_version}/root/usr/bin
+  # Backward compatibility for RHEL/CentOS 7
+  source /etc/os-release
+
+  if [[ $VERSION_ID -lt 8 ]]; then
+    old_pgengine=/opt/rh/$old_collection/root/usr/bin
+    new_pgengine=/opt/rh/rh-postgresql${new_raw_version}/root/usr/bin
+  else
+    old_pgengine=/usr/lib64/pgsql/postgresql-$old_raw_version/bin
+    new_pgengine=/usr/bin
+  fi
+
   PGDATA_new="${PGDATA}-new"
 
   printf >&2 "\n==========  \$PGDATA upgrade: %s -> %s  ==========\n\n" \
@@ -350,9 +359,9 @@ run_pgupgrade ()
   # boot up data directory with old postgres once again to make sure
   # it was shut down properly, otherwise the upgrade process fails
   info_msg "Starting old postgresql once again for a clean shutdown..."
-  "${old_pgengine}/pg_ctl" start -w --timeout 86400 -o "-h ''"
+  "${old_pgengine}/pg_ctl" start -w --timeout 86400 -o "-h 127.0.0.1''"
   info_msg "Waiting for postgresql to be ready for shutdown again..."
-  "${old_pgengine}/pg_isready"
+  "${old_pgengine}/pg_isready" -h 127.0.0.1
   info_msg "Shutting down old postgresql cleanly..."
   "${old_pgengine}/pg_ctl" stop
 
--- a/src/root/usr/share/container-scripts/postgresql/common.sh
+++ b/src/root/usr/share/container-scripts/postgresql/common.sh
@@ -298,7 +298,7 @@ function wait_for_postgresql_master() {
 run_pgupgrade ()
 (
   # Remove .pid file if the file persists after ugly shut down
-  if [ -f "$PGDATA/postmaster.pid" ] && ! pgrep -f "postgres" > /dev/null; then
+  if [ -f "$PGDATA/postmaster.pid" ] && ! pg_isready > /dev/null; then
     rm -rf "$PGDATA/postmaster.pid"
   fi
 
@@ -312,8 +312,17 @@ run_pgupgrade ()
     old_collection=rh-postgresql$old_raw_version
   fi
 
-  old_pgengine=/opt/rh/$old_collection/root/usr/bin
-  new_pgengine=/opt/rh/rh-postgresql${new_raw_version}/root/usr/bin
+  # Backward compatibility for RHEL/CentOS 7
+  source /etc/os-release
+
+  if [[ $VERSION_ID -lt 8 ]]; then
+    old_pgengine=/opt/rh/$old_collection/root/usr/bin
+    new_pgengine=/opt/rh/rh-postgresql${new_raw_version}/root/usr/bin
+  else
+    old_pgengine=/usr/lib64/pgsql/postgresql-$old_raw_version/bin
+    new_pgengine=/usr/bin
+  fi
+
   PGDATA_new="${PGDATA}-new"
 
   printf >&2 "\n==========  \$PGDATA upgrade: %s -> %s  ==========\n\n" \
@@ -351,9 +360,9 @@ run_pgupgrade ()
   # boot up data directory with old postgres once again to make sure
   # it was shut down properly, otherwise the upgrade process fails
   info_msg "Starting old postgresql once again for a clean shutdown..."
-  "${old_pgengine}/pg_ctl" start -w --timeout 86400 -o "-h ''"
+  "${old_pgengine}/pg_ctl" start -w --timeout 86400 -o "-h 127.0.0.1''"
   info_msg "Waiting for postgresql to be ready for shutdown again..."
-  "${old_pgengine}/pg_isready"
+  "${old_pgengine}/pg_isready" -h 127.0.0.1
   info_msg "Shutting down old postgresql cleanly..."
   "${old_pgengine}/pg_ctl" stop
 
--- a/10/root/usr/share/container-scripts/postgresql/common.sh
+++ b/10/root/usr/share/container-scripts/postgresql/common.sh
@@ -297,7 +297,7 @@ function wait_for_postgresql_master() {
 run_pgupgrade ()
 (
   # Remove .pid file if the file persists after ugly shut down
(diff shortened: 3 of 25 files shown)


eifrach:
[test]

13/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -297,7 +297,7 @@ function wait_for_postgresql_master() {
 run_pgupgrade ()
 (
   # Remove .pid file if the file persists after ugly shut down
-  if [ -f "$PGDATA/postmaster.pid" ] && ! pgrep -f "postgres" > /dev/null; then
```
fila43, line 300: Why are you removing the condition? We do not want to remove `.pid` file of  the running process

13/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -352,7 +352,6 @@ run_pgupgrade ()
   info_msg "Starting old postgresql once again for a clean shutdown..."
   "${old_pgengine}/pg_ctl" start -w --timeout 86400 -o "-h ''"
   info_msg "Waiting for postgresql to be ready for shutdown again..."
-  "${old_pgengine}/pg_isready"
```
fila43, line 355: Why are you removing this safety procedure?

13/root/usr/share/container-scripts/postgresql/common.sh
```diff
@@ -308,11 +308,11 @@ run_pgupgrade ()
   if test "$old_raw_version" = 92; then
     old_collection=postgresql92
   else
-    old_collection=rh-postgresql$old_raw_version
+    old_collection=postgresql-$old_raw_version
   fi
```
fila43, line 312: This does not correspond with the current state of PostgreSQL rpm names. We need to adjust it and check if the `old_colllection` variable is used

fila43:
Overall, it looks good. Thank you for your contribution. We will need to generalize it and implement it into distgen templates.

13/Dockerfile.c8s
```diff
@@ -44,7 +44,8 @@ COPY root/usr/libexec/fix-permissions /usr/libexec/fix-permissions
 RUN yum -y module enable postgresql:13 && \
     INSTALL_PKGS="rsync tar gettext bind-utils nss_wrapper postgresql-server postgresql-contrib" && \
     INSTALL_PKGS="$INSTALL_PKGS pgaudit" && \
-    yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
+    INSTALL_PKGS="$INSTALL_PKGS procps-ng util-linux postgresql-upgrade" && \
+    yum -y --setopt=tsflags=nodocs --setopt=install_weak_deps=false install $INSTALL_PKGS && \
```
pkubatrh, line 48: What difference in packages is there when using this option? I would rather not introduce it for already built image versions if the change is too disruptive.

13/root/usr/share/container-scripts/postgresql/common.sh
eifrach: So I removed it because I had some issues when I was testing the upgrade. I will look into it deeper.
we should check if the PID  is up first

13/root/usr/share/container-scripts/postgresql/common.sh
eifrach: On my testing it returned exit code 2 - I did manual checks the socket was up and running but the `pg_isready` wasn't catching it.
On the above line we use `-w` which is waiting for the DB to be ready. I thought maybe we don't need to double check it unless there is another reason for it?

13/Dockerfile.c8s
eifrach: it not a must, it removes about 100MB of size - those are packages that are extra doesn't change / affect the core application

13/root/usr/share/container-scripts/postgresql/common.sh
eifrach: so this is from a running container I added a stop before the line
seem that the service is not running, also there isn't any open socket

I'm assuming the it is partially match `run-postgresql` 
```
bash-4.4$ pgrep -f "postgres" 
1
17

 ps -ef
UID          PID    PPID  C STIME TTY          TIME CMD
postgres       1       0  0 12:18 pts/0    00:00:00 /bin/bash /usr/bin/run-postgresql
postgres      17       1  0 12:18 pts/0    00:00:00 /bin/bash /usr/bin/run-postgresql
postgres      18      17  0 12:18 pts/0    00:00:00 /usr/bin/coreutils --coreutils-prog-shebang
postgres      19       0  0 12:19 pts/1    00:00:00 bash
postgres      27      19  0 12:22 pts/1    00:00:00 ps -ef
```

how about 
```suggestion
  if [ -f "$PGDATA/postmaster.pid" ] && ! pg_isready > /dev/null; then
 ```

File: 13/root/usr/share/container-scripts/postgresql/common.sh (hunk 2 above)
Comment (eifrach) [reply 13/root/usr/share/container-scripts/postgresql/common.sh:355]: after some searching, It seems we start the server as local only -- which mean it doesn't create a TCP port instead it creates a Linux socket.

the 'pg_isready' tries the port and fails

File: 13/root/usr/share/container-scripts/postgresql/common.sh (hunk 3 above)
Comment (eifrach) [reply 13/root/usr/share/container-scripts/postgresql/common.sh]: changed

File: 13/root/usr/share/container-scripts/postgresql/common.sh (hunk 2 above)
Comment (eifrach) [reply 13/root/usr/share/container-scripts/postgresql/common.sh:355]: fixed

**Comment by eifrach** [comment]
> Overall, it looks good. Thank you for your contribution. We will need to generalize it and implement it in the master branch. https://github.com/sclorg/postgresql-container/blob/master/src/root/usr/share/container-scripts/postgresql/common.sh

I've issue fixes for all the remarks 
once they approved I'll create a PR with changes to all common.sh files

File: 13/root/usr/share/container-scripts/postgresql/common.sh (hunk 1 above)
Comment (eifrach) [reply 13/root/usr/share/container-scripts/postgresql/common.sh:300]: changed

Comment (fila43) [reply 13/root/usr/share/container-scripts/postgresql/common.sh:300]: makes sense +1

File: 13/Dockerfile.c8s (hunk 4 above)
Comment (fila43) [reply 13/Dockerfile.c8s:48]: For the existing images, I would choose the more defensive approach. I would just add the necessary packages.

File: 13/root/usr/share/container-scripts/postgresql/common.sh (hunk 2 above)
Comment (fila43) [reply 13/root/usr/share/container-scripts/postgresql/common.sh:355]: +1

File: src/root/usr/share/container-scripts/postgresql/common.sh [hunk 5]
```diff
@@ -312,7 +312,7 @@ run_pgupgrade ()
     old_collection=rh-postgresql$old_raw_version
   fi
 
-  old_pgengine=/opt/rh/$old_collection/root/usr/bin
+  old_pgengine=/usr/lib64/pgsql/postgresql-$old_raw_version/bin
   new_pgengine=/opt/rh/rh-postgresql${new_raw_version}/root/usr/bin
```
fila43, line 316: `/usr/bin` ?

fila43:
It makes a good impression on me. Please implement the changes into `/src` directory. Once it is there, we will check and adjust the compatibility for rhel7/centos7 vs rhel8+ versions, or we can wait for the EOL of centos7/rhel7 images. Also, there should be the prepared upgrade test that needs to be enabled for rhel8+ containers.

eifrach:
[test]

eifrach:
cc @danmanor

fila43:
LGTM

fila43:
@phracek would it be possible to enable this test also for RHEL8+ ? 
https://github.com/sclorg/postgresql-container/blob/d0cecca7766a6150489228dc2670143bf73a3997/test/run_test#L682

fila43:
[test]

erthalion:
Hi,

Thanks for the PR! I've got a couple of general commentaries about PostgreSQL
major version upgrade, and although they may not necessarily be related to the
PR, I reckon it's a good idea to collect them here. As an outside person I
could be missing some details of this project, please let me know if that's the
case.

* `POSTGRESQL_UPGRADE` seems to allow only `--link` and `--copy`, but since PG
  11 there is also `--clone` option, which somewhat combines best of both
  worlds, but requires newer Linux kernel versions.

* The docs mention `--link` option as an unsafe, but the old cluster should not
  be used only of the new cluster was started. If it wasn't the case, the old
  cluster was unmodified and should be safe to start.

* It seems that it's possible to provide custom initdb options via
  `POSTGRESQL_UPGRADE_INITDB_OPTIONS`, but some of them must be the same as the
  old cluster (collate, encoding, data checksums, wal segment size). It sounds
  a bit unreliable to leave it to the container user, could those options be
  enforced?

* It's worth running `pg_upgrade --check` to verify compatibility.

* Maybe a good idea to add `-j` to `upgrade_cmd` for additional parallelism.

* Usually triggering a vacuum after the upgrade to rebuild statistics is a
  good idea as well.

* More open-ended question, what do you think about those situations where
  some coordination is needed? E.g. when upgrading a replica, most likely rsync
  has to be used in coordination with the primary upgrade. Or upgrading
  extensions, which is not handled via `pg_upgrade` either.

  It's probably beyond what you want to achieve within a container, but it's
  at least worth mentioning in the documentation, right?

https://www.postgresql.org/docs/current/pgupgrade.html

fila43:
[test]

fila43:
[test]

fila43:
[test]

fila43:
[test]

fila43:
@phracek @zmiklank any idea about c9s failures? RHEL9 passed but c9s failed

zmiklank:
> @phracek @zmiklank any idea about c9s failures? RHEL9 passed but c9s failed

This is reproducible also in the master branch - so not seem to be related to this PR. Seems like a segfault (double free) from `pg_isready` binary.

Aren't there any differences in c9s and rhel9 binaries?

zmiklank:
So this may be an postgresql[1] issue which is present only when using newer openssl (3.2.1) that got into c9s, but not into rhel9 yet. postgresql upstream patch seeems to exist already, more info is in the linked mailing thread[1].

[1] https://www.postgresql.org/message-id/CAN55FZ1eDDYsYaL7mv%2BoSLUij2h_u6hvD4Qmv-7PK7jkji0uyQ%40mail.gmail.com

eifrach:
sorry for the delay, I was on PTO 
looking into the failures.

zmiklank:
> sorry for the delay, I was on PTO looking into the failures.

No problem at all. Just note that c8s images will be deprecated on May the 31th. So no need to work further on them I guess. (This PR needs to merged first: https://github.com/sclorg/postgresql-container/pull/567)

fila43:
[test]

fila43:
I tested RHEL8 failed cases locally and it works, c9s fails are known issues so I would merge it.  @eifrach could you please confirm my observation to be sure before merge?

eifrach:
> I tested RHEL8 failed cases locally and it works, c9s fails are known issues so I would merge it. @eifrach could you please confirm my observation to be sure before merge?

yes it works for me - also I've tested the data migration

PR #572, opened 2024-06-10
Add file show_all_imagestreams.py from container-common-scripts

phracek:
This pull request only adds show_all_imagestreams.py so it is working.

The file `show_all_imagestreams.py` shows all available imagestreams for this container, that are specified
in the directory `imagestreams`.

The example output will be seen in the next comment.


phracek:
[test-openshift]

PR #580, opened 2024-07-19
Use VERSION instead of SINGLE_VERSION, like we do in whole org

phracek:


phracek:
[test-openshift-pytest]

PR #582, opened 2024-07-30
Change the default from md5 to scram-sha-256 in pg_hba.conf

bdunne:

### DIFF
--- a/12/root/usr/share/container-scripts/postgresql/common.sh
+++ b/12/root/usr/share/container-scripts/postgresql/common.sh
@@ -216,10 +216,10 @@ EOF
 #
 
 # Allow connections from all hosts.
-host all all all md5
+host all all all scram-sha-256
 
 # Allow replication connections from all hosts.
-host replication all all md5
+host replication all all scram-sha-256
 EOF
 }
 
--- a/13/root/usr/share/container-scripts/postgresql/common.sh
+++ b/13/root/usr/share/container-scripts/postgresql/common.sh
@@ -216,10 +216,10 @@ EOF
 #
 
 # Allow connections from all hosts.
-host all all all md5
+host all all all scram-sha-256
 
 # Allow replication connections from all hosts.
-host replication all all md5
+host replication all all scram-sha-256
 EOF
 }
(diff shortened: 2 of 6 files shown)


kbrock:
Looks like this has been supported since postgres 10 and is the standard as of postgres 14

```
# postgresql.conf
password_encryption = scram-sha-256
```

Someone just needs to set the password after changing that setting and all systems go. (even if `pg_hba.conf` is still set to `md5`

To see if you need to upgrade:
```
SELECT
    rolname, rolpassword ~ '^SCRAM-SHA-256\$' AS has_upgraded
FROM pg_authid
WHERE rolcanlogin;
```

/via https://www.crunchydata.com/blog/how-to-upgrade-postgresql-passwords-to-scram

hhorak:
We are not able to follow the steps to convert all passwords as the article above describes, because that's what only the users of the postgresql image can do.

What we can do is making the default algorithm configurable and changing the default -- that should not affect new deployments because the pg_hba.conf file is created only when creating the database, but we can still surprise users that might use older clients.

Anyway, the change is inevitable at some point, so we should start somewhere (changing the default only in RHEL10/c10s might be the least painful way).

@fila43 WDYT?

kbrock:
@hhorak I think we are close to the same page.

This migration is definitely on our radar since many companies are requiring the migration to `scram`.

It is really too bad that there is not a way to upgrade the passwords automatically without risking writing the plain text passwords to disk.

So what I heard you say:

- Set the postgres.conf default to `scram-sha-256`.
  - Only needed for postgres 10-13.
  - Can remove in postgres 14 since that is the default.
- Add a parameter for `POSTGRES_AUTH_METHOD`, defaulting it to `md5`.
- Use the parameter in `pg_hba.conf`.
- Change the default value in RHEL10/c10s to `scram-sha-256`.

Questions:

- Is this parameter exposed to end users and accessible via something like `podman -e POSTGRES_AUTH_METHOD scram-sha-256`?


Cheers to Brno. When we worked for RH, we enjoyed our great team members over there.

PR #588, opened 2024-09-03
Remove scl check

hhorak:

### DIFF
--- a/12/Dockerfile.rhel8
+++ b/12/Dockerfile.rhel8
@@ -54,7 +54,8 @@ RUN yum -y module enable postgresql:12 && \
     localedef -f UTF-8 -i en_US en_US.UTF-8 && \
     test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)" && \
     mkdir -p /var/lib/pgsql/data && \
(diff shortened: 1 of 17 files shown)


hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

hhorak:
A comment about pagila - we should rather download directly from github (issue to be created).

phracek (approved):
Awesome fix. Please create an issue for closing pagila instead of downloading RPM.

LGTM. Rewards will be sent soon.

hhorak:
> A comment about pagila - we should rather download directly from github (issue to be created).

Reported as #589

test/pagila.sh
SlouchyButton, line 1: The reason why I don't think this is a good idea long term (and why the separate issue was created) is that since we are linking to a specific version on the repository that is not supported anymore, the link can stop working anytime.

Silently missing, this is resolved with the check that the download succeeded, which is a great addition!

test/run_test
SlouchyButton, line 1: I have mentioned this change in the Google doc to keep it updated.

SlouchyButton (approved):
Amazing work

PR #591, opened 2024-09-16
Add Helm chart tests for PSQL

phracek:
Add Helm chart tests for PSQL

We need to test if helm charts work in postgresql container

In OpenShift we deliver helmcharts and we need to test them by each pull request.


phracek:
Tests have passed. Let's merge it.

Issue #595, opened 2024-10-08
`run_pgaudit_test` crashing the test suite


SlouchyButton:
This issue was introduced by https://github.com/sclorg/postgresql-container/pull/399.

A new test introduced this line in `run_pgaudit_test`, which in the PR had an error:
`-v DOCKER_EXTRA_ARGS || cp -r "$test_dir"/examples/pgaudit/* "$config_dir"`

The line is missing test at the beginning, so that the new test `run_new_pgaudit_test`, which doesn't really test pgaudit, rather custom extension loading via container argument, takes precedence over the custom config that is in the `run_pgaudit_test` to load pgaudit.

The problem is that if the config is not present, the whole test crashes with the error:
```
-----------------------------------------------
Running test run_pgaudit_test (starting at 2024-10-08 15:02:02+02:00) ...
-----------------------------------------------
29eecc21255a18dee73d3ca9d81786c5aafae722344559577f70160d482cd83a
Created container 29eecc21255a18dee73d3ca9d81786c5aafae722344559577f70160d482cd83a
Asserting runtime option pg-test-pgaudit
pg-test-pgaudit shared_preload_libraries pgaudit

test/run: line 352: $1: unbound variable
Tests finished with EXIT=1
==============================================
Cleaning of testing containers and images started.
It may take a few seconds.
```

Interesting is that it fully stops further testing, but won't really crash the testsuite as the testsuite will gracefully clean all the containers.

Another weird thing is that it fails on `run_pg_audit_test` already, not the new `run_new_pgaudit_test`.

Ideally, the test suite should still continue and mention that this test failed as described previously in this issue. 
This is related to https://github.com/sclorg/postgresql-container/issues/537 where this issue was first spotted while reviewing aforementioned PR. At some later point a mistake was introduced into the PR which made test always copy the config files and generating an error (that was not caught, so tests passed). This mistake was not caught when merging the PR at that point and also made all tests pass as the changes weren't used in the test.

This mistake then started generating `-v: command not found` error in logs.

After fixing this issue, tests are not passing - with the same problem as from the original PR.

SlouchyButton:
Problem is that `DOCKER_EXTRA_ARGS` variable is set already which it shouldn't be, so the `test` command returns 0, so no `cp` happens and we get the crash. Variable is empty, but apparently do exist as `test -v DOCKER_EXTRA_ARGS` is `TRUE`.

This can be fixed with replacing check for existing variable with check for empty variable.

Issue #537, opened 2023-11-21
Testsuite seems to be passing, even though a specific tests fails.


zmiklank:
Strange behaviour.

There is an error in `test/run` script:
```bash
-----------------------------------------------
Running test run_pgaudit_test (starting at 2023-10-03 10:26:59+00:00) ... 
-----------------------------------------------
c644c8223c59eb64a932a2d1cd2e13e9f3bee4a1c7fd79ec69e1df6e37a788c0
Created container c644c8223c59eb64a932a2d1cd2e13e9f3bee4a1c7fd79ec69e1df6e37a788c0
test/run: line 340: $1: unbound variable
Tests finished with EXIT=1

```

Tests are passing:

```bash
==============================================
Tests were run for image quay.io/sclorg/postgresql-15-c9s:1
==============================================
Test cases results:

 [PASSED] for 'postgresql-container_tests' run_container_creation_tests (00:00:19)
 [PASSED] for 'postgresql-container_tests' run_general_tests (00:00:39)
 [PASSED] for 'postgresql-container_tests' run_change_password_test (00:00:07)
 [PASSED] for 'postgresql-container_tests' run_replication_test (00:00:06)
 [PASSED] for 'postgresql-container_tests' run_doc_test (00:00:00)
 [PASSED] for 'postgresql-container_tests' run_s2i_test (00:00:14)
 [PASSED] for 'postgresql-container_tests' run_test_cfg_hook (00:00:08)
 [PASSED] for 'postgresql-container_tests' run_s2i_bake_data_test (00:00:12)
 [PASSED] for 'postgresql-container_tests' run_s2i_enable_ssl_test (00:00:08)
 [PASSED] for 'postgresql-container_tests' run_upgrade_test (00:00:00)
 [PASSED] for 'postgresql-container_tests' run_migration_test (00:00:23)
```

_Originally posted by @phracek in https://github.com/sclorg/postgresql-container/issues/399#issuecomment-1744700208_

zmiklank:
It is little bit confusing what the log line `Tests finished with EXIT=1` is saying to us.
In reality it means, that the exit code of the test script was not zero, however, this does not mean, that the TESTSUITE_RESULT was success or that the tests finished successfully.
I do not see the results of  `run_pgaudit_test` in the logs that Peto has found. That could mean, that unbound variable led to exiting the whole testsuite even before the test result itself was recorded.
[...]

PR #399, opened 2021-02-17
Add support for extensions loading and creation.

mscherer:
Using POSTGRESQL_LIBRARIES and POSTGRESQL_EXTENSIONS allow
users to enable extensions (some bundled in the container) without
having to do anything but adding 2 variables.

PR #596, opened 2024-10-08
Fix error in `run_pgaudit_test()`

SlouchyButton:
Initially command `test was missing`, but after it was added test failed and crashed in a way that no more tests were run. This behavior is known issue tracked here in issue #537

Problem is that `DOCKER_EXTRA_ARGS` variable is set already which it shouldn't be, so the `test` command returns 0, so no `cp` happens and we get the crash.

Variable is empty, but apparently do exist as
`test -v DOCKER_EXTRA_ARGS` is `TRUE`.

Introduced by PR #399

Fixes: #595

### DIFF
--- a/test/run_test
+++ b/test/run_test
@@ -934,7 +934,7 @@ run_pgaudit_test()
   # create a dir for config
   config_dir=$(mktemp -d --tmpdir pg-hook-volume.XXXXX)
   add_cleanup_command /bin/rm -rf "$config_dir"
-  -v DOCKER_EXTRA_ARGS || cp -r "$test_dir"/examples/pgaudit/* "$config_dir"
+  test -n "$DOCKER_EXTRA_ARGS" || cp -r "$test_dir"/examples/pgaudit/* "$config_dir"
   setfacl -R -m u:26:rwx "$config_dir"
 
   # create a dir for data


SlouchyButton:
[test]

test/run_test
```diff
@@ -934,7 +934,7 @@ run_pgaudit_test()
   # create a dir for config
   config_dir=$(mktemp -d --tmpdir pg-hook-volume.XXXXX)
   add_cleanup_command /bin/rm -rf "$config_dir"
-  -v DOCKER_EXTRA_ARGS || cp -r "$test_dir"/examples/pgaudit/* "$config_dir"
+  test -n "$DOCKER_EXTRA_ARGS" || cp -r "$test_dir"/examples/pgaudit/* "$config_dir"
```
phracek, line 937: This means if the string is not empty. That's correct. Definition is here https://github.com/sclorg/postgresql-container/blob/master/test/run_test#L41. It means, that if variable is defined, then pgaudit is copied. Am I right?

phracek (commented):
Only if my expectations or understanding of the code is correct.

test/run_test
phracek: Otherwise LGTM.

phracek:
[test]

test/run_test
SlouchyButton: > This means if the string is not empty.

This is correct

> if variable is defined, then pgaudit is copied

Yes as long as the variable is empty, if it's not, it's not copied.

PR #597, opened 2024-10-18
Disable tests depending on  help.1 file presence

lukaszachy:
This file will no longer be included in images produced by Konflux, thus we shouldn't test it.


phracek:
[test]

PR #603, opened 2025-01-15
Enable testing postgresql-container on RHEL10 quest.

phracek:
No code is changed. GitHub Action should not be needed to call


phracek:
[test]

PR #608, opened 2025-03-24
Rename package bind-utils to bind9.18-utils

phracek:
Replace 'bind-utils' with 'bind9.18-utils'

See Jira issue https://issues.redhat.com/browse/RHEL-80345

It is valid only for RHEL9


phracek:
[test][test-openshift]

fila43:
LGTM

PR #640, opened 2026-01-06
Container PyTest suite for Postgresql-container

phracek:
This pull request adds PyTest suite migrated from `run_pytest` scripts.

```
run_container_creation_tests -> test_container_configuration.py
run_general_tests -> test_container_general.py
run_change_password_test -> test_container_password.py
run_replication_test -> test_container_replication.py
run_s2i_test -> test_container_basics.py
run_test_cfg_hook -> test_container_configuration.py
run_s2i_bake_data_test -> test_container_ssl.py
run_s2i_enable_ssl_test -> test_container_ssl.py
run_pgaudit_test -> test_container_extensions.py
run_pgvector_test -> test_container_extensions.py
run_env_extension_load_test -> test_container_extensions.py
run_logging_test -> test_container_logging.py
```

All tests are migrated except `run_migration_test` and `run_upgrade_test`.


phracek:
Let's try first round 
[test-pytest]

phracek:
Let's try next round
[test-pytest]

test/test_container_configuration.py
```diff
@@ -0,0 +1,216 @@
...
+        """
+        assert self.db.assert_container_creation_succeeds(
+            container_args=[
+                psql_user,
+                psql_password,
+                psql_database,
+                psql_admin_password,
+            ],
+            command="",
+        )
```
pkhartsk: Unlike in the bash tests, this doesn't try to actually connect to the container after creation as `test_connection_func` and `connection_params` are missing. This causes a case that fails in the bash tests to pass here.

Reproducer:
```py
psql_user="the@user"
psql_password="pass"
psql_database="db"
```

Output from pytests:
```
PASSED test_container_configuration.py::TestPostgreSQLConfigurationContainer::test_correct_configuration_tests[-e POSTGRESQL_USER="the@user"--e POSTGRESQL_PASSWORD="the pass"--e POSTGRESQL_DATABASE="the db"--e POSTGRESQL_ADMIN_PASSWORD=]
```

Output from bash tests:
```
Created container 99890eb43cc9bc0a105aa15ef49954f88516d4bd16da3e1058786519646207e9
  Testing PostgreSQL connection to 10.88.8.202...
    Trying to connect...
psql: error: could not translate host name "user@10.88.8.202" to address: Name or service not known
```

Obviously the bash tests don't handle it exactly correctly here, because input into the connection URI should be URI-encoded. But I would suggest adding a connection attempt here, too.

pkhartsk (changes requested):

test/test_container_configuration.py
phracek: Makes sense to me. I will do it.

pkhartsk: But that code is in `container_lib`

test/test_container_ssl.py
```diff
@@ -0,0 +1,136 @@
...
+            database="postgres",
+            expected_success=True,
+        )
+
+        output = self.dw_api.postgresql_cmd(
+            container_ip=ssl_cip,
+            container_id=VARS.IMAGE_NAME,
+            username="postgres",
+            password=admin_password,
+            database="postgres?sslmode=require",
```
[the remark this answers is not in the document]
pkhartsk: And if one were to URI encode the database name like I propose we do in the bash tests, the string would come out to `postgresql://postgres@ip:5432/postgres%3Fsslmode%3Drequire` which obviously fails.

So I suggest that postgresql_cmd be modified to include a `parameters` argument where these would be included.

test/test_container_configuration.py
```diff
@@ -0,0 +1,253 @@
...
+                "",
+                "",
+                '"the @password"',
+            ],
+            [
+                '"the user"',
+                '"the pass"',
+                '"the db"',
+                "",
+            ],
```
pkhartsk: This defines the username as `the user` when creating the container because it's passed as an environment variable, but as `"the user"` when connecting to it, which is different from the former. I suggest removing the single quotes and instead surrounding `psql_user` etc. with double quotes below (line 173-181) inside the f-strings.

test/test_container_ssl.py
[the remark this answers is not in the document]
pkhartsk: Fixed in https://github.com/pkhartsk/postgresql-container/commit/ac442759bf420a08b080de09edd292a5bbe7247c
and https://github.com/sclorg/container-ci-suite/pull/153

pkhartsk (changes requested):
Found some things and included proposed fixes for both

test/test_container_configuration.py
pkhartsk: Continued in the next review.

phracek:
[test-pytest]

phracek:
@fila43 Please review parts `test_container_upgrade.py` and `test_container_migration.py`. The pagila is now part of this repo. In case you would like to download it e.g. from https://github.com/devrimgunduz/pagila, then please let me know. Thanks.

test/conftest.py
```diff
@@ -20,12 +25,40 @@
...
+PAGILA = ["pagila-schema.sql", "pagila-data.sql", "pagila-insert-data.sql"]
+SQL_CMDS = {
+    "select count(*) from information_schema.triggers;": "16",
+    "select count(*) from staff;": "2",
+    "select * from information_schema.tables;": "28",
+}
+RHEL9 = "none 13 15 16 18 none"
+RHEL10 = "none 16 18 none"
+UPGRADE_PATH_DICT = {
+    "rhel8": "none 12 13 15 16 18 none",
```
fila43: we do not ship postgresql 18 in RHEL8

phracek: fixed by https://github.com/sclorg/postgresql-container/pull/640/commits/756999dd480d6ee47bde3ac9f5dc5eac74474a15

phracek:
Rebased against 'master' branch

[test-pytest][test]

test/conftest.py
```diff
@@ -63,11 +89,37 @@ def get_upgrade_path():
...
-        "rhel9": f"registry.redhat.io/rhel9/postgresql-{version}",
-        "rhel10": f"registry.redhat.io/rhel10/postgresql-{version}",
-        "c9s": f"quay.io/sclorg/postgresql-{version}-c9s",
-        "c10s": f"quay.io/sclorg/postgresql-{version}-c10s",
-    }
-    return ns[VARS.OS]
+    """
+    Get the image ID of the PostgreSQL container.
+    """
+    if VARS.OS in ["rhel8", "rhel9", "rhel10"]:
```
frenzymadness: `VARS.OS.startswith("rhel")` might be more future-proof.

test/test_container_password.py
```diff
@@ -0,0 +1,168 @@
...
+                    f"Login access failed for {user}:{pwd} with expected success {ret_value}"
+                )
+                login_access = False
+        assert login_access
+        self.dw_api.run_sql_command(
+            container_ip=cip_new,
+            username=username,
+            password=password,
+            container_id=VARS.IMAGE_NAME,
+        )
```
frenzymadness: This should fail as it uses the old password. If not, this part is likely worthless.

test/conftest.py
```diff
@@ -1,11 +1,16 @@
...
 import sys
+import urllib3
+
 
 from pathlib import Path
 from collections import namedtuple
 
 from container_ci_suite.utils import check_variables
 
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
frenzymadness: How many tests need this so it's worth it doing globally? Cannot we use a context manager and disable these warnings only when really needed?

test/test_container_configuration.py
```diff
@@ -0,0 +1,299 @@
...
+import pytest
+
+from container_ci_suite.engines.podman_wrapper import PodmanCLIWrapper
+from container_ci_suite.container_lib import ContainerTestLib
+from container_ci_suite.utils import ContainerTestLibUtils
+from container_ci_suite.engines.database import DatabaseWrapper
+
+from conftest import VARS
+
+volume_dir = tempfile.mkdtemp(prefix="/tmp/psql-volume-dir")
```
frenzymadness: This is not cleaned automatically.

test/test_container_extensions.py
```diff
@@ -0,0 +1,199 @@
...
+)
+from container_ci_suite.engines.database import DatabaseWrapper
+from container_ci_suite.engines.podman_wrapper import PodmanCLIWrapper
+from container_ci_suite.utils import get_file_content
+
+import pytest
+
+from conftest import VARS
+
+pg_audit_volume_dir = tempfile.mkdtemp(prefix="/tmp/psql-pgaudit-volume-dir")
```
frenzymadness: Here as well.

frenzymadness (commented):
First few ideas.

Also, there are typos in commit messages: `propoer` instead of `proper` and `test_container_generall.py` instead of `test_container_general.py`

test/conftest.py
phracek: Only two tests. Context manager looks good to me.

test/test_container_extensions.py
phracek: Fixed by https://github.com/sclorg/postgresql-container/pull/640/commits/a11650926c7df437db2ba9acdcf238a94056cd3e

test/test_container_configuration.py
phracek: Fixed by https://github.com/sclorg/postgresql-container/pull/640/commits/a11650926c7df437db2ba9acdcf238a94056cd3e

test/conftest.py
phracek: Fixed by https://github.com/sclorg/postgresql-container/pull/640/commits/a11650926c7df437db2ba9acdcf238a94056cd3e

We do not need it at all.

test/test_container_password.py
phracek: It is really not needed. Based on the run_test script.

phracek: Fixed by https://github.com/sclorg/postgresql-container/pull/640/commits/a11650926c7df437db2ba9acdcf238a94056cd3e

test/conftest.py
phracek: Fixed by https://github.com/sclorg/postgresql-container/pull/640/commits/a11650926c7df437db2ba9acdcf238a94056cd3e

test/test_container_password.py
frenzymadness: This piece of code is still there.

test/conftest.py
frenzymadness: If a list of versions does not start with "none", `prev` might be undefined here and it might be hard to debug. I'd either handle that situation here with a proper error message explaining what is wrong or add a comment above the dictionary definition explaining what is expected from each version's list.

test/test_container_extensions.py
```diff
@@ -0,0 +1,198 @@
...
+    Test PostgreSQL container configuration.
+    """
+
+    def setup_method(self):
+        """
+        Setup the test environment.
+        """
+        self.db = ContainerTestLib(image_name=VARS.IMAGE_NAME, db_type="postgresql")
+        self.db_api = DatabaseWrapper(image_name=VARS.IMAGE_NAME, db_type="postgresql")
+        self.pg_audit_volume_dir = tempfile.mkdtemp(prefix="/tmp/psql-pgaudit-volume-dir")
```
frenzymadness: This temporary directory should be deleted in the teardown method. Be aware that there are more temp dirs with no cleanup in the testsuite.

test/test_container_basics.py
```diff
@@ -0,0 +1,122 @@
...
+        self.app_image.cleanup()
+
+    def test_backup_functionality(self):
+        """
+        Test backup functionality of the PostgreSQL container.
+        Steps are:
+        1. Test if the container creation fails with invalid combinations of arguments
+        2. Test if the container creation succeeds with valid combinations of arguments
+        3. Test if the database connection works
+        4. Test if the backup functionality works
```
frenzymadness: All 4? I don't see at least the first one here.

frenzymadness (commented):
A couple more points.

test/conftest.py
phracek: Thanks for this. I will completely rewritten. We are in Python world. I should be done more effectivelly. "none" really does not make sense.

test/test_container_basics.py
phracek: Good point https://github.com/sclorg/postgresql-container/pull/640/commits/a1d537475170cb091669c9f4460b266b57d8e89c

test/test_container_extensions.py
phracek: Fixed here: https://github.com/sclorg/postgresql-container/pull/640/commits/a1d537475170cb091669c9f4460b266b57d8e89c

test/conftest.py
phracek: Rewritten by this. https://github.com/sclorg/postgresql-container/pull/640/commits/ead6067e24db32c21a47fe77d6b550b8cd88df25.

test/test_container_password.py
phracek: Hmm, I totally add to commit ;) https://github.com/sclorg/postgresql-container/pull/640/commits/a1d537475170cb091669c9f4460b266b57d8e89c

phracek:
@frenzymadness Thanks for the second round of review.

phracek:
[test][test-pytest]

test/conftest.py
```diff
@@ -34,40 +64,58 @@
...
-        prev = version
-    if prev == "none":
+    """
+    Get the upgrade path of the PostgreSQL container.
+    """
+    if VARS.VERSION not in UPGRADE_PATH_DICT[VARS.OS]:
         return None
-    return prev
+    current_index = UPGRADE_PATH_DICT[VARS.OS].index(VARS.VERSION)
+    if current_index - 1 == -1:
```
frenzymadness: What about this:

```suggestion
    if current_index == 0:
```

test/conftest.py
```diff
@@ -34,40 +64,58 @@
...
-    }
-    for version in upgrade_path.keys():
-        if version == VARS.VERSION:
-            break
-        prev = version
-    if prev == "none":
+    """
+    Get the upgrade path of the PostgreSQL container.
+    """
+    if VARS.VERSION not in UPGRADE_PATH_DICT[VARS.OS]:
```
frenzymadness: You are using `UPGRADE_PATH_DICT[VARS.OS]` lookups multiple times. What about having `upgrade_path = UPGRADE_PATH_DICT[VARS.OS]` on top of the function and then use that new variable after?

test/conftest.py
```diff
@@ -34,40 +64,58 @@
...
-        "rhel10": "none 13 15 16 none",
-        "fedora": "none 12 13 14 15 16 none",
-    }
-    for version in upgrade_path.keys():
-        if version == VARS.VERSION:
-            break
-        prev = version
-    if prev == "none":
+    """
+    Get the upgrade path of the PostgreSQL container.
```
frenzymadness, line 79: Are you sure this function works at all? `VARS.VERSION` is string (from `os.getenv`) and the lists of versions now contain integers so the very first condition is always true and this function therefore returns `None` every time you call it. Doesn't it?

test/test_container_password.py
```diff
@@ -0,0 +1,164 @@
...
+            username=username,
+            password=new_password,
+            max_attempts=10,
+        )
+        login_access = True
+        for user, pwd, ret_value in [
+            (username, new_password, True),
+            (username, password, False),
+            ("postgres", new_admin_password, True),
+            ("postgres", "admin_password", False),
```
frenzymadness: Are you sure about this? What about:

```suggestion
            ("postgres", admin_password, False),
```

test/test_container_extensions.py
```diff
@@ -0,0 +1,201 @@
...
+        assert cip and cid
+        assert self.db_api.wait_for_database(
+            container_id=cid, command="/usr/libexec/check-container"
+        )
+        output = PodmanCLIWrapper.podman_exec_shell_command(
+            cid_file_name=cid,
+            cmd='psql -tA -c "SHOW shared_preload_libraries;"',
+        )
+        assert "vector" in output, (
+            f"pgaudit should be in the shared_preload_libraries, but is {output}"
```
frenzymadness: pgaudit?

test/test_container_upgrade.py
```diff
@@ -0,0 +1,222 @@
...
+        Test upgrade functionality of the PostgreSQL container.
+        Steps are:
+        1. Create a container with the previous version and run pg_upgrade
+        2. Create a container with the new version and run pg_upgrade
+        3. Test if the database connection works
+        """
+        if VARS.OS == "fedora":
+            pytest.skip("Skip upgrade test on Fedora. Only CentOS Stream and RHELs are supported.")
+        self.datadir = datadir
+        self.upgrade_volume_dir = tempfile.mkdtemp(prefix=f"/tmp/psql-upgrade-{upgrade_type}-{datadir}")
```
frenzymadness: There is no teardown that would clean this after the test run.

test/test_container_upgrade.py
```diff
@@ -0,0 +1,222 @@
...
+            # Sometimes check-container can return before the database is fully ready to accept connections.
+            if not self.db_api.wait_for_database(
+                container_id=cid_create,
+                command='psql -h localhost -tA -c \"select 1;\"',
+            ):
+                time.sleep(5)
+                continue
+            is_db_ready = True
+            break
+        assert is_db_ready, "Database is not ready after waiting for 25 seconds"
```
frenzymadness: This is an exact piece of code as there are on lines 165-180. Can we maybe extract it somewhere and reuse?

test/test_container_upgrade.py
```diff
@@ -0,0 +1,222 @@
...
+            "1",
+            "2",
+            "3",
+        ]
+        for row in rows:
+            assert re.search(row, output), f"Row {row} not found in {output}"
+        PodmanCLIWrapper.call_podman_command(cmd=f"stop {cid}")
+        PodmanCLIWrapper.call_podman_command(cmd=f"rm -f {cid}")
+
+    def check_pagila_db(self, cid):
```
frenzymadness: This seems to be copy-pasted in two files? What about reusing it from conftest?

test/test_container_extensions.py
```diff
@@ -0,0 +1,201 @@
...
+            ]
+        )
+
+    def teardown_method(self):
+        """
+        Teardown the test environment.
+        """
+        self.db.cleanup()
+        shutil.rmtree(self.pg_audit_volume_dir)
+        shutil.rmtree(self.pg_vector_volume_dir)
```
frenzymadness: Use `ignore_errors=True` here in order to avoid exceptions if the dirs are busy or don't exist.

test/test_container_password.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        Teardown the test environment.
+        """
+        self.pwd_change.cleanup()
+
+    def test_password_change(self):
+        """
+        Test password change.
+        """
+        pwd_dir = tempfile.mkdtemp(prefix="/tmp/psql-pwd-change")
```
frenzymadness: Not removed after the tests.

test/test_container_logging.py
```diff
@@ -0,0 +1,86 @@
...
+        Teardown the test environment.
+        """
+        self.db.cleanup()
+
+    def test_logging_destination(self):
+        """
+        Test logging destination.
+        """
+        cid_file_name = "test_pg_logging"
+        logging_dir = tempfile.mkdtemp(prefix="/tmp/psql-logging-volume-dir")
```
frenzymadness: Not removed after the tests.

test/test_container_extensions.py
```diff
@@ -0,0 +1,201 @@
...
+        3. Check if the pgvector extension is loaded.
+        4. Execute the SQL commands to enable the pgvector extension.
+        5. Check if the SQL commands executed successfully.
+        """
+        if VARS.VERSION in ["11", "12", "13", "15"]:
+            pytest.skip("pgvector not expected on this version, test skipped.")
+        if VARS.OS == "rhel8":
+            pytest.skip("pgvector not expected on this OS, test skipped.")
+        cid_file_name = "test_pg_pgvector"
+        config_dir = tempfile.mkdtemp(prefix="/tmp/psql-pgvector-volume")
```
frenzymadness: Not removed after the tests run.

test/test_container_extensions.py
```diff
@@ -0,0 +1,201 @@
...
+from container_ci_suite.utils import get_file_content
+
+import pytest
+
+from conftest import VARS
+
+
+class TestPostgreSQLPluginContainer:
+    """
+    Test PostgreSQL container configuration.
```
frenzymadness: This comment does not make much sense because it's repeated in every single testing class.

test/test_container_password.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+        login_access = True
+        for user, pwd, ret_value in [
+            (username, new_password, True),
+            (username, password, False),
+            ("postgres", new_admin_password, True),
+            ("postgres", "admin_password", False),
+        ]:
+            # Let's check login access for user and pwd combinations
+            # with expected ret_value. Otherwise let's failed.
```
frenzymadness: "Otherwise let's failed."

test/test_container_upgrade.py
```diff
@@ -0,0 +1,222 @@
...
+            command="",
+        )
+        cip_create, cid_create = self.upgrade_db.get_cip_cid(
+            cid_file_name=cid_file_name
+        )
+        assert cid_create and cip_create
+        assert self.db_api.wait_for_database(
+            container_id=cid_create, command="/usr/libexec/check-container"
+        )
+        # Let's wait couple seconds
```
frenzymadness, line 108: I guess we don't need comments like that for a simple and single line of code.

test/test_container_upgrade.py
```diff
@@ -0,0 +1,222 @@
...
+    def upgrade_image(self, upgrade_type: str, bool_test_upgrade: bool = True):
+        """
+        Upgrade the image.
+        Steps are:
+        1. Create a container with the new version and run pg_upgrade
+        2. Test if the database connection works
+        3. If bool_test_upgrade is True, check the database output after upgrade.
+           If False, just check the connection as the second upgrade
+           can take more time and we have already tested the output after the first upgrade.
+        4. If datadir is empty, create a simple table and insert some data.
```
frenzymadness, line 133: I don't see any tables created in this test.

test/test_container_ssl.py
```diff
@@ -0,0 +1,137 @@
...
+        output = self.dw_api.postgresql_cmd(
+            container_ip=ssl_cip,
+            container_id=VARS.IMAGE_NAME,
+            username="postgres",
+            password=admin_password,
+            database="postgres",
+            uri_params={"sslmode": "require"},
+            sql_command="-At -c 'SELECT 1;'",
+        )
+        assert re.search(r"1", output), f"1 not found in {output}"
```
frenzymadness: Is this test satisfied just by finding number "1" whenever in the output? I mean, it can also be part of other numbers etc.

test/test_container_upgrade.py
```diff
@@ -0,0 +1,222 @@
...
+        output = PodmanCLIWrapper.podman_exec_shell_command(
+            cid_file_name=cid,
+            cmd="psql -h localhost -At -c \"SELECT * FROM blah ORDER BY id;\"",
+        )
+        rows = [
+            "1",
+            "2",
+            "3",
+        ]
+        for row in rows:
```
frenzymadness: Same here, kinda weak assertion.

test/test_container_general.py
```diff
@@ -0,0 +1,207 @@
...
+            expected_admin_success = True
+        container_all_args = [
+            psql_user_arg,
+            psql_pwd_arg,
+            db_name_arg,
+            admin_root_password_arg,
+            f"-e POSTGRESQL_MAX_CONNECTIONS={psql_max_connections}",
+            f"-e POSTGRESQL_MAX_PREPARED_TRANSACTIONS={psql_max_prepared_transactions}",
+            f"-e POSTGRESQL_SHARED_BUFFERS={psql_shared_buffers}",
+            f"{container_args}",
```
frenzymadness: container_args is a string, there is no need for f-string here.

test/conftest.py
```diff
@@ -34,40 +64,58 @@
     TAG=OS.replace("rh", "-", 1),
     TEST_APP=TEST_APP,
     VERY_LONG_IDENTIFIER=VERY_LONG_IDENTIFIER,
+    PREVIOUS_VERSION=PREVIOUS_VERSION_DICT.get(VERSION),
+    PAGILA=PAGILA,
```
frenzymadness: AFAIK, this is not used anywhere.

test/test_container_configuration.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+    Test PostgreSQL container invalid configurations tests.
+    """
+
+    def setup_method(self):
+        """
+        Setup the test environment.
+        """
+        self.db = ContainerTestLib(image_name=VARS.IMAGE_NAME, db_type="postgresql")
+        self.db_api = DatabaseWrapper(image_name=VARS.IMAGE_NAME, db_type="postgresql")
```
frenzymadness: Unused

frenzymadness (changes requested):
Another run. If you know you are copy-pasting code or using the same patterns in multiple places, it would be nice to fix the issues I point out in a single place across all instances.

phracek:
[test-pytest]

test/test_container_logging.py
```diff
@@ -0,0 +1,81 @@
+import re
+
+from pathlib import Path
+import shutil
+
+from container_ci_suite.container_lib import ContainerTestLib
+from container_ci_suite.engines.container import ContainerTestLibUtils
+from container_ci_suite.engines.podman_wrapper import PodmanCLIWrapper
+from container_ci_suite.utils import tempfile
```
frenzymadness: What is the point in importing stdlib modules from `container_ci_suite`? If you one day remove that import from `util.py`, this will break. Similar import is in `test_container_password.py`.

test/conftest.py
```diff
@@ -34,40 +66,198 @@
...
+
+def check_db_output(
+    dw_api,
+    cip,
+    username,
+    password,
+    database,
+):
+    """
+    Check the database out+put if the data is inserted correctly
```
frenzymadness: `out+put`?

test/conftest.py
```diff
@@ -34,40 +66,198 @@
...
+    Create a container and wait for it to be ready.
+    Args:
+        db: ContainerTestLib instance
+        cid_file_name: Name of the container ID file
+        container_args: List of container arguments
+        command: Command to run in the container
+    Steps:
+    1. Create the container
+    2. Get the container ID and IP
+    3. Wait for the database to be ready if required by wait_for_database flag
```
frenzymadness: Where is the `wait_for_database` flag?

test/test_container_upgrade.py
```diff
@@ -0,0 +1,161 @@
...
+        self.db.image_name = self.registry_image
+        self.create_database_in_prev_version()
+        self.upgrade_image(upgrade_type=upgrade_type)
+        self.upgrade_image(upgrade_type=upgrade_type, bool_test_upgrade=False)
+
+    def create_database_in_prev_version(self):
+        """
+        Create a database in the old version.
+        Steps are:
+        1. Create a container with the previous version and run pg_upgrade
```
frenzymadness, line 91: I don't see the upgrade happening here.

test/test_container_general.py
```diff
@@ -0,0 +1,185 @@
...
+        expected_rows = [
+            rf"max_connections\s*=\s*{psql_max_connections}",
+            rf"max_prepared_transactions\s*=\s*{psql_max_prepared_transactions}",
+            rf"shared_buffers\s*=\s*{psql_shared_buffers}",
+        ]
+        for expected_row in expected_rows:
+            assert re.search(expected_row, output), (
+                f"Row {expected_row} not found in {output}"
+            )
+        # test_postgresql
```
frenzymadness: Useless comment.

test/test_container_general.py
```diff
@@ -0,0 +1,185 @@
...
+
+from container_ci_suite.container_lib import ContainerTestLib
+from container_ci_suite.engines.podman_wrapper import PodmanCLIWrapper
+
+from conftest import VARS, check_db_output, create_and_wait_for_container
+
+
+class TestPostgreSQLGeneralContainer:
+    """
+    Test PostgreSQL container configuration.
```
frenzymadness: As mentioned earlier, the same generic docstring in multiple test classes is not useful.

test/test_container_ssl.py
```diff
@@ -0,0 +1,96 @@
...
+    def teardown_method(self):
+        """
+        Teardown the test environment.
+        """
+        self.db.cleanup()
+
+    def test_ssl(self):
+        """
+        Test SSL.
+        """
```
frenzymadness, line 66: This test seems to have nothing to do with SSL.

frenzymadness (commented):
One more round, and I hope we are done. This is mostly about comments and names, but the one about imports is important. There are also some open threads from previous rounds. I believe you can merge this once at least the imports are fixed.

test/conftest.py
phracek: Removed by commit https://github.com/sclorg/postgresql-container/pull/640/commits/e0ad923ecac41c9361428fb2db1659e4c1cad274

test/test_container_general.py
phracek: Removed by commit https://github.com/sclorg/postgresql-container/pull/640/commits/e0ad923ecac41c9361428fb2db1659e4c1cad274

test/test_container_general.py
phracek: removed by https://github.com/sclorg/postgresql-container/pull/640/commits/e0ad923ecac41c9361428fb2db1659e4c1cad274

test/test_container_logging.py
phracek: Typo. Fixed by https://github.com/sclorg/postgresql-container/pull/640/commits/e0ad923ecac41c9361428fb2db1659e4c1cad274

test/test_container_ssl.py
phracek: Renames to enable_ssl_example. https://github.com/sclorg/postgresql-container/pull/640/commits/e0ad923ecac41c9361428fb2db1659e4c1cad274

test/test_container_upgrade.py
phracek: Removed. It does not make sense at all. Right. https://github.com/sclorg/postgresql-container/pull/640/commits/e0ad923ecac41c9361428fb2db1659e4c1cad274

test/conftest.py
phracek: Fixed by https://github.com/sclorg/postgresql-container/pull/640/commits/e0ad923ecac41c9361428fb2db1659e4c1cad274

test/test_container_extensions.py
phracek: Removed, I don't see any reason, why to have it. https://github.com/sclorg/postgresql-container/pull/640/commits/e0ad923ecac41c9361428fb2db1659e4c1cad274

phracek:
[test][test-pytest]

